PR #523, opened 2014-07-30
add total time field to disk_io_counters()

joemiller:
the total-time field is necessary to calculate a `util%` metric (0 - 100%) similar to what `iostat -x` prints.

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -199,7 +199,8 @@ sdiskusage = namedtuple('sdiskusage', ['total', 'used', 'free', 'percent'])
 # psutil.disk_io_counters()
 sdiskio = namedtuple('sdiskio', ['read_count', 'write_count',
                                  'read_bytes', 'write_bytes',
-                                 'read_time', 'write_time'])
+                                 'read_time', 'write_time',
+                                 'tot_time'])
 # psutil.disk_partitions()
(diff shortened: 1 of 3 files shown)


fasionchan:
I want this feature as well. IO UTIL is an important metric I need to collection. Field name tot_time is not so specified, I suggest to rename it to busy_time or io_time.

So why read_time + write_time >= busy_time? This is because disk devices can operate several read/write operations parallelly. That is, device is busy no matter how many operations it is operating. However, write_time will be n times if n operations.

IO UTIL metric, measures how many time the device is busy. Solution on Linux is just show above. That is similar solutions on other platforms. For instance, on FreeBSD, we can use busy_time field in devstat struct(sys/devicestat.h)。

Hope busy_time to been included. I may share some of my code.

giampaolo:
Agreed this is necessary. We first need to figure out on what platforms this is retrievable though, especially OSX.

fasionchan:
All right. I will help to find out solutions for other platforms.

giampaolo:
FreeBSD implementation: https://github.com/giampaolo/psutil/pull/769
Linux: https://github.com/giampaolo/psutil/commit/edd6c223849d73f3e87154be3e4909ce258f77f1

giampaolo:
It appears this info is available only on Linux and BSD. I only added FreeBSD implementation though because on OpenBSD and NetBSD I couldn't find a way to get the right metrics. Closing this out.

PR #568, opened 2015-01-01
example/pidof.py same as pidof cli tool mentioned in TODO:

karthikrev:


examples/pidof.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
+
```
giampaolo: Add the license here:

# Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
# Use of this source code is governed by a BSD-style license that can be
# found in the LICENSE file.

examples/pidof.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+A clone of 'pidof' cmdline utility.
+$ pidof  /usr/bin/python
+1140 1138 1136 1134 1133 1129 1127 1125 1121 1120 1119
+"""
+
```
giampaolo: from __future__ import print_function

examples/pidof.py
```diff
@@ -0,0 +1,31 @@
...
+    return pids
+
+
+def main():
+    if len(sys.argv) != 2:
+        sys.exit('usage: %s pgm_name' % __file__)
+    else:
+        pgmname = sys.argv[1]
+    pids = pidsof(pgmname)
+    print " ".join(pids)
```
giampaolo: put parentheses areound print()

examples/pidof.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import psutil
+import sys
+
+
+def pidsof(pgm):
+    pids = []
+    for proc in psutil.process_iter():
+        process = proc.as_dict()
```
giampaolo: there's no need to call as_dict() (which retrieves *all* process info; instead just call proc.cmdline()

giampaolo: also, put a try/except statement around proc.cmdline() like this:

try:
     cmdline = proc.cmdline()
except psutil.Error:
     continue

examples/pidof.py
karthikrev: Hi ..Thanks for quick response ..  I am newbie .. Please advice if I should use 
    # Copyright (c) 2009, Giampaolo Rodola' 
or 
   # Copyright (c) 2009, Karthik (my name) ?

giampaolo: I guess you could put both names.

giampaolo:
OK, this is now included. Thanks.

PR #573, opened 2015-01-08
Modules aren't scripts.

tomprince:
They shouldn't have shebang lines.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
--- a/psutil/_compat.py
+++ b/psutil/_compat.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 # Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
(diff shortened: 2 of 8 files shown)


giampaolo:
AFAIK all scripts should have the executable shebang because if the script is executable he/she can run it as "./script.py" instead of "python script.py". Where the shebang is not necessary is in python (importable) modules but I tend to include it anyway, probably because I'm simply used to do that.

tomprince:
Tools like rpmlint and lintian complain about them (since it thinks they are scripts, but aren't installed as executable).

giampaolo:
They *are* scripts, only they are not *installed* at system level. I still do not understand the issue.

mrjefftang:
I think this makes sense.  There shouldn't be any reason for a user to run ./psutil/_bsd.py

These modules are made to be imported by a script (that has the #!).

giampaolo:
No, these modules are *not* meant to be imported by a script. They are stand alone scripts.

tomprince:
They aren't stand-alone --- they don't *do* anything (besides defining stuff to be imported elsewhere). Note that I only changed the modules in `psutil`, not the example directory.

giampaolo:
Ah sorry, I didn't figure out your PR touched python modules in /psutil directory. I was somehow convinced you were talking about scripts in the /examples directory.
I still think it's a good idea to leave the shebang also for Python modules. To say one, with the shebang UNIX tools like "file" can figure out that those are python files. Also, I admit it's kind of part of my mental boilerplate since I started using Python.

tomprince:
`file` is still able to figure out they are python files (the extension also helps).

I stumbled upon this because rpmlint complains (see http://pkgs.fedoraproject.org/cgit/python-psutil.git/tree/python-psutil.spec#n50) for example).

giampaolo:
Looking back at this now, I think it makes sense to be merged.

asergi:
`psutil/_common.py` is missing and the shebang is also broken.

Issue #250, opened 2014-05-23
Add network interface speed information


giampaolo:
_From [nicolash...@gmail.com](https://code.google.com/u/115069581463621203225/) on January 30, 2012 15:14:15_
    
    Proposal: add the network interface spped (100 Mbps, 1000 Mbps...) in the 
    structure returned by the psutil.network_io_counters(True) function 
    
        # python
        # import statgrab
        # statgrab.sg_get_network_iface_stats()
        [{'duplex': 2, 'interface_name': 'lo', 'speed': 0, 'up': 1}, {'duplex': 0, 
    'interface_name': 'eth0', 'speed': 100, 'up': 1}, {'duplex': 2, 
    'interface_name': 'wlan0', 'speed': 0, 'up': 1}, {'duplex': 2, 
    'interface_name': 'virbr0', 'speed': 0, 'up': 1}] 
    
    Please provide any additional information below.  
    -
    
_Original issue: http://code.google.com/p/psutil/issues/detail?id=250_

giampaolo:
Merged into master as of PR #592.

tevaum:
Thanks! :]

PR #576, opened 2015-01-23
Attempting to fix issue 250, adding NIC speed

karthikrev:
https://github.com/giampaolo/psutil/issues/250
The code just works, please advice if I need to change the format / code /anything else. 
https://github.com/giampaolo/psutil/issues/250

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1844,8 +1844,8 @@ def users():
     return _psplatform.users()
 
 
-def nics_info(nic):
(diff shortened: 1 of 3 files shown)


giampaolo:
Thanks for this contribution. Given this uses a socket and ioctl() I think this can be done in Python as well (see https://docs.python.org/2/library/fcntl.html). If that's the case, as I think, Python should be preferred over C. Can you investigate into that?

PR #578, opened 2015-02-06
move extension into the package

mindw:


setup.py
```diff
@@ -11,11 +11,7 @@
 
 import os
 import sys
-try:
-    from setuptools import setup, Extension
-except ImportError:
-    from distutils.core import setup, Extension
-
+from setuptools import setup, Extension
```
giampaolo: Why this? AFAIK setuptools is a third-party dep which is not installed by default.

giampaolo:
Thanks for writing this. The changes you did against the import system broke some imports though:
https://travis-ci.org/giampaolo/psutil/jobs/49733189
Maybe because you changed paths in setup.py perhaps?

setup.py
```diff
@@ -134,6 +130,7 @@ def get_winver():
 
 def main():
     setup_args = dict(
+        zip_safe=False,
```
giampaolo: According to this:
https://pythonhosted.org/setuptools/setuptools.html#setting-the-zip-safe-flag
zip_safe=False seems to be unnecessary because if the package contains C extension files (as psutil) zip_safe is already implicitly set to False. Given that zip_safe is an option which exists only if setuptools is installed I think it makes sense to remove it.

giampaolo:
Also, once you manage to fix the import don't forget to update README.rst and CREDITS.rst (add your name).

setup.py
mindw: Yep.

mindw:
I'm looking into the import failure on py 3.x.

giampaolo:
Yes, the import failure is because the different name in setup.py. For example, in __init__.py you'll have to do this:

-        import _psutil_linux
+        import psutil._psutil_linux

mindw:
finally fixed the tests.

setup.py
mindw: pip requires setuptools if installing from source. Wheels do not use the setup.py script. Both 'bootstrapped' in 2.7 & 3.4.

giampaolo: We still support python 2.6.

giampaolo: I'm eager to do this only if a brand new Python installation (2.6 included) has setuptools installed.

mindw:
I dropped the setuptools stuff.

psutil/__init__.py
```diff
@@ -52,42 +52,47 @@
...
+    CONN_FIN_WAIT1,
+    CONN_FIN_WAIT2,
+    CONN_TIME_WAIT,
+    CONN_CLOSE,
+    CONN_CLOSE_WAIT,
+    CONN_LAST_ACK,
+    CONN_LISTEN,
+    CONN_CLOSING,
+    CONN_NONE
+)
```
giampaolo: Please don't do this for now. I'll think about whether I'm OK with this later.

giampaolo:
I made different commits today and this PR cannot be merged cleanly anymore. 
I created a branch in here in which I merged my changes with yours:
https://github.com/giampaolo/psutil/tree/mindw-move_extension_2_pkg
Everything is fine if I run tests via "make test" but if I try to run them via "tox" I get import related errors (if you want to try just "pip install tox" then run "tox" from the main psutil directory).
Now I remember why I didn't do this before: because I tried but I always got stuck with these kind of issues.
I've been struggling with this for hours now and I really don't know how to fix this problem. Do you have any idea?

    ~/svn/psutil {mindw-move_extension_2_pkg|MERGING}$ tox
    GLOB sdist-make: /home/giampaolo/svn/psutil/setup.py
    py26 inst-nodeps: /home/giampaolo/svn/psutil/.tox/dist/psutil-3.0.0.zip
    py26 runtests: PYTHONHASHSEED='2581425950'
    py26 runtests: commands[0] | py.test
    ============================================================================================= test session starts =============================================================================================
    platform linux2 -- Python 2.6.9 -- py-1.4.26 -- pytest-2.6.4
    collected 0 items / 2 errors 

    =================================================================================================== ERRORS ====================================================================================================
    _________________________________________________________________________________ ERROR collecting test/test_memory_leaks.py __________________________________________________________________________________
    test/test_memory_leaks.py:20: in <module>
        import psutil
    psutil/__init__.py:68: in <module>
        from . import _pslinux as _psplatform
    psutil/_pslinux.py:26: in <module>
        from . import _psutil_linux as cext
    E   ImportError: cannot import name _psutil_linux
    ____________________________________________________________________________________ ERROR collecting test/test_psutil.py _____________________________________________________________________________________
    test/test_psutil.py:54: in <module>
        import psutil
    psutil/__init__.py:68: in <module>
        from . import _pslinux as _psplatform
    psutil/_pslinux.py:22: in <module>
        from . import _common
    E   ImportError: cannot import name _common
    =========================================================================================== 2 error in 0.33 seconds ===========================================================================================

giampaolo:
Note: travis tests pass so it seems to be a problem related with tox:
https://travis-ci.org/giampaolo/psutil/builds/49831656

giampaolo:
OK, I solved the issue by adding "usedevelop = True" in the tox config file. I merged your PR as of
https://github.com/giampaolo/psutil/commit/e5d09464ded836f9e57f591fc5088ad05c6a5099
Thanks a lot.

giampaolo:
Unfortunately I had to revert this change in https://github.com/giampaolo/psutil/commit/454b1ac3ae7e594f644e9da1e16ef586c4de7c70 because relative imports no longer works if I run the interactive interpreter from within psutil installation directory.

mindw:
Why on earth would you do that?

giampaolo:
Well, when I work on psutil I have a shell open in psutil root directory so it's not rare I open up a python interpreter in order to try some stuff. I also sometimes executes scripts (import psutil) from within the root directory and the same issue occurs. I also reverted the changes in setup.py because the same problem applies.

PR #580, opened 2015-02-06
Integrate history and install into main doc

mindw:
makes doc more complete

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,6 +10,8 @@
    Old 1.2.1 documentation is still available
    `here <https://code.google.com/p/psutil/wiki/Documentation>`__.
 
+.. contents::
+   
 psutil documentation
 ====================
 
@@ -20,7 +22,6 @@ Quick links
 * `Blog <http://grodola.blogspot.com/search/label/psutil>`__
 * `Download <https://pypi.python.org/pypi?:action=display&name=psutil#downloads>`__
 * `Forum <http://groups.google.com/group/psutil/topics>`__
-* `What's new <https://github.com/giampaolo/psutil/blob/master/HISTORY.rst>`__
 
 About
 -----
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -10,6 +10,8 @@
    Old 1.2.1 documentation is still available
    `here <https://code.google.com/p/psutil/wiki/Documentation>`__.
 
+.. contents::
```
giampaolo: I don't like this because the table of contents is already visible in the left column

docs/index.rst
```diff
@@ -20,7 +22,6 @@ Quick links
 * `Blog <http://grodola.blogspot.com/search/label/psutil>`__
 * `Download <https://pypi.python.org/pypi?:action=display&name=psutil#downloads>`__
 * `Forum <http://groups.google.com/group/psutil/topics>`__
-* `What's new <https://github.com/giampaolo/psutil/blob/master/HISTORY.rst>`__
```
giampaolo: I'd leave this here

docs/index.rst
```diff
@@ -43,6 +44,11 @@ From project's home page:
 
 The psutil documentation you're reading is distributed as a single HTML page.
 
+Installation
+============
+
+.. include:: ../INSTALL.rst
+
```
giampaolo: I'm -1 on this because adds too many text to the doc which I prefer it remains an "API only doc". 
Instead you can simply mention the INSTALL instruction in the "quick links" section as a link.

docs/index.rst
```diff
@@ -1253,3 +1259,8 @@ Constants
   `man prlimit <http://linux.die.net/man/2/prlimit>`__ for futher information.
 
   Availability: Linux
+
+History
+=======
+
+.. include:: ../HISTORY.rst
```
giampaolo: Same here: it adds too many cruft. I would leave this out.

mindw:
I find it best where all documentation is in one place - then it works nicely with RTD and in searches.

But it's your call.

giampaolo:
I've integrated some of your changes as of https://github.com/giampaolo/psutil/commit/c8a77563eec70404cc6991da6686bc8f3d9358af

PR #587, opened 2015-02-09
move extension into package - without relative imports.

mindw:


giampaolo:
Changing the C extension path in setup.py causes the same problem I described in #578 (importing psutil from its root source directory):

    >>> import psutil
      File "psutil/__init__.py", line 63, in <module>
        from psutil import _pslinux as _psplatform
      File "psutil/_pslinux.py", line 26, in <module>
        import psutil._psutil_linux as cext
    ImportError: No module named _psutil_linux
    >>> 

I think there's no way around that.

mindw:
Did you do setup.py build_ext -f -i before opening the prompt?

giampaolo:
``setup.py build_ext -i `` did the trick. I reapplied your original changes. Thanks a lot.

Issue #542, opened 2014-10-03
open_files  does find out doc files opened by ms office word on windows. 


tomqyp:
open_files  does find out doc files opened by ms office word on windows. Is that "Skip handles" code block missed anything?
I tried Sysinternals handle.exe, that can get it.

amolvaikar:
@giampaolo 

I investigated this a bit and it seems that this condition is responsible for the doc files getting missed:
```python
handle.GrantedAccess == 0x0012019f
```
I noticed on a different open_files related defect you had mentioned that you are unable to recollect the original source of these access permissions related code:
https://github.com/giampaolo/psutil/issues/340#issuecomment-44022221

By any chance, was it this source:
http://www.kernelmode.info/forum/viewtopic.php?f=10&t=195

Further, I think I have been able to find what the particular code 0x0012019f stands for with the help of this blog:
https://www.eldos.com/forum/read.php?FID=13&TID=2769
and the actual winnt.h
The access permissions that form this value are:
STANDARD_RIGHTS_WRITE 20000
FILE_WRITE_DATA 0002
FILE_WRITE_ATTRIBUTES 0100
FILE_WRITE_EA 0010
FILE_APPEND_DATA 0004
SYNCHRONIZE 100000
FILE_READ_EA 0008
FILE_READ_DATA 0001
FILE_READ_ATTRIBUTES 0080

so basically referring to the winnt.h, 0x0012019f  seems to stand for:
FILE_GENERIC_READ|FILE_GENERIC_WRITE
Other hex values too can be figured out similarly, I will do so when time permits.

Meanwhile, after I recompiled the latest psutil after removing the check for 0x0012019f, I was able to get the document in my listing, now need to check if that causes any hangs on some other processes though. Will try to run the test suite and let you know.

Comments? Thoughts?
And yes, please do not remove this API.
Lastly, thank you for psutil.

amolvaikar:
Ran the tests.
Here is the summary and the two failures I got:
```python

======================================================================
FAIL: test_halfway_terminated_process (__main__.TestProcess)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test\test_psutil.py", line 2025, in test_halfway_terminated_process
    self.fail("NoSuchProcess exception not raised for %r" % name)
AssertionError: NoSuchProcess exception not raised for 'num_ctx_switches'

======================================================================
FAIL: test_fetch_all (__main__.TestFetchAllProcesses)
----------------------------------------------------------------------
[...]

Issue #340, opened 2014-05-23
get_open_files might hang forever


giampaolo:
_From [amo...@google.com](https://code.google.com/u/101252647480125095856/) on November 24, 2012 15:39:51_
    
    What steps will reproduce the problem?  
    
    
    1. Create a process_iter()
    2. Call get_open_files on each of the processes 
    
    
    
    What is the expected output?  
    The list of open files for each process. 
    
    
    
    What do you see instead?  
    The program hangs forever.
    
    This has already been encountered, there are some exception in the code for 
    exactly this problem. I have found on more access mask where the problem arises 
    - please add it to the list of exceptions. Patch is attached.
    
**Attachment:** [patch](http://code.google.com/p/psutil/issues/detail?id=340)
    
_Original issue: http://code.google.com/p/psutil/issues/detail?id=340_

giampaolo:
This is now supposed to be fixed by PR #597 which spawns a thread.

daveajones:
thanks @giampaolo @mrjefftang .  much appreciated.

giampaolo:
From [original comment](https://github.com/giampaolo/psutil/issues/340#issuecomment-59365752):

> This blog post (http://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html) suggests using GetFileInformationByHandleEx (http://msdn.microsoft.com/en-us/library/aa364953(VS.85).aspx) which is introduced in WIndows Vista+

...here's an archived copy in case that blog disappears:
https://web.archive.org/web/20201023075814/https://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html

PolarGoose:
> From [original comment](https://github.com/giampaolo/psutil/issues/340#issuecomment-59365752):
> 
> > This blog post (http://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html) suggests using GetFileInformationByHandleEx (http://msdn.microsoft.com/en-us/library/aa364953(VS.85).aspx) which is introduced in WIndows Vista+
> 
> ...here's an archived copy in case that blog disappears: https://web.archive.org/web/20201023075814/https://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html

I have tried [GetFileInformationByHandleEx](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getfileinformationbyhandleex):
[...]

PR #596, opened 2015-02-23
Rewrite get_open_files

mrjefftang:
I rewrote get_open_files to clean up a lot of the code and to address the error cases where handles or memory aren't closed/free'd.

This will address #542 since the access mask check is removed.

Hopefully this will address the newest issue in #340 but I don't have a Windows machine with bluetooth to test that.  Maybe @jimbo1969 can retest this.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -515,10 +515,9 @@ class Process(object):
         if _POSIX:
             ppid = self._proc.ppid()
         else:
-            if self._ppid is None:
-                ppid = self._proc.ppid()
-            self._ppid = ppid
-        self._proc._ppid = ppid
+            # Cache the PPID if it is not set
+            self._ppid = self._ppid or self._proc.ppid()
+            ppid = self._ppid
         return ppid
 
     def name(self):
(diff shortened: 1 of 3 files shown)


mrjefftang:
I closed/re-opened the PR to trigger a new build and it looks like that test case that initially failed is nondeterministic.

giampaolo:
Have you tested this on both Python 2 and 3? Do  "make.bat test" and "make.bat test-memleaks" pass?

mrjefftang:
Note that I still think the way forward is to go the Sysinternals Handles route and spawn a thread that calls NtQueryObject and waits a for a timeout and kills it off.  Looks like we can avoid the associated memory leaks by converting the thread to a fiber.

Fixed an issue that looks like it was in the master branch relating to Process.ppid() on Windows.

Test failure on Python 3.4 is related to memory leak for net_if_addrs:
```
FAIL: test_net_if_addrs (__main__.TestModuleFunctionsLeaks)
  File "test\test_memory_leaks.py", line 413, in test_net_if_addrs
    self.execute('net_if_addrs')
  File "test\test_memory_leaks.py", line 86, in execute
    % (rss2, rss3, difference))
AssertionError: rss2=17768448, rss3=18120704, difference=352256
```

Python 2.7:
```
FAIL: test_net_if_stats (__main__.TestModuleFunctionsLeaks)
  File "test\test_memory_leaks.py", line 417, in test_net_if_stats
    self.execute('net_if_stats')
  File "test\test_memory_leaks.py", line 86, in execute
    % (rss2, rss3, difference))
AssertionError: rss2=12668928, rss3=13201408, difference=532480

```

PR #603, opened 2015-03-04
Fix ioprio range error

spacewander:
IO priority levels range from 0 (highest) to 7 (lowest), according to the manpage of `ioprio_set`


wiggin15:
This commit looks good (travis failed for other reasons). Any chance to get it merged?

giampaolo:
I merged this. Thanks for your patch.

PR #608, opened 2015-03-15
fix tests

wiggin15:
Fix multiple problems with tests.
- when run from /root, os.remove after setegid/seteuid in LimitedTestCase failed with PermissionError because the 'other' permissions are 0 in that directory. Moved os.remove to before setegid/seteuid
- fixed travis error seen here: https://travis-ci.org/giampaolo/psutil/jobs/54378596
- fixed travis race condition seen here: https://travis-ci.org/giampaolo/psutil/jobs/53887824

Still working on ~11 failing tests on Redhat 5.6

### DIFF
--- a/test/test_psutil.py
+++ b/test/test_psutil.py
@@ -264,12 +264,14 @@ def wait_for_file(fname, timeout=GLOBAL_TIMEOUT, delete_file=True):
         try:
             with open(fname, "r") as f:
                 data = f.read()
+            if not data:
+                continue
             if delete_file:
                 os.remove(fname)
             return data
         except IOError:
             time.sleep(0.001)
-    raise RuntimeError("timed out (couldn't create file)")
+    raise RuntimeError("timed out (couldn't read file)")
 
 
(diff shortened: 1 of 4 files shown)


wiggin15:
I added commits to fix a few other things. There are still more tests that fail on RHEL 5.6, but I ran out of time to fix them for now

test/test_psutil.py
```diff
@@ -710,6 +712,8 @@ def test_cpu_count(self):
         self.assertEqual(logical, len(psutil.cpu_times(percpu=True)))
         self.assertGreaterEqual(logical, 1)
         #
+        if "physical id" not in open("/proc/cpuinfo").read():
+            raise unittest.SkipTest("cpuinfo doesn't include physical id")
```
giampaolo: I think this is gonna raise a ResourceWarning on python 3 because the file isn't closed properly. I would use with statement instead.

wiggin15:
fixed comment

PR #614, opened 2015-04-01
cpu_count is inconsistent between proc count and cpu count

johnburnett:
cpu_count() seems to return inconsistent results between platforms.  I think it comes down to an assumption on whether logical=False means "physical cpus" or "physical cores".  On Windows, it looks like it returns the core count, while Linux returns the cpu count.

In my case, I was assuming the former (cores), and that's the actual useful value I'm needing (in this case, I want to run something on N-1 cores, leaving one full core for the user).  However, on Linux, with a dual-proc X5550, I'm getting 2 for logical=False, and 16 for logical=True.  Launching 15 threads swamps the machine, and 2 is obviously not great :).  It works as expected on Windows (logical=False gives 8).

I have a patch for _pslinux.py:cpu_count_physical to make it behave the way I expected, but I wanted to check if all of the above sounds right?  I also haven't looked at any other platform beyond Linux and Windows, and what is happening there.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -279,14 +279,38 @@ def cpu_count_logical():
 
 
 def cpu_count_physical():
-    """Return the number of physical CPUs in the system."""
+    """Return the number of physical cores in the system."""
     with open('/proc/cpuinfo', 'rb') as f:
-        found = set()
+        # For each 'physical id' found, find its corresponding
+        # 'cpu cores' count and then total them up.
+        proc_core_counts = {}
+        current_proc_id = None
+        current_core_count = 0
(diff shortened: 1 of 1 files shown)


johnburnett:
Ping on this? (just got around to fixing the python 3 ci break related to this...)

giovtorres:
Confirmed this PR fixes the same bug I found.  The current cpu_count_physical() returns number of sockets, not number of physical cores.

giampaolo:
I've just fixed this with a variant of your patch. Please confirm this works as expected.

giovtorres:
Works. Thanks.

Issue #562, opened 2014-12-26
OpenBSD support


landryb:
Related to #429, i'm working on a basic OpenBSD support in https://github.com/landryb/psutil/tree/openbsd - so far, builds, and passes *some* tests - dunno if you want an early pull request or when it's more complete.... Code copied from FreeBSD, then adapted for OpenBSD.

landryb:
So i've resumed working on this in https://github.com/landryb/psutil/tree/openbsd, merged last changes on master and fixing on top of this.

I used to be able to run the tests but probably after the changes from #578 it fails with version mismatch, while i have no other version of psutils..

```
creating /home/landry/.local/lib/python2.7/site-packages/psutil-3.0.0-py2.7-openbsd-5.6-amd64.egg
Extracting psutil-3.0.0-py2.7-openbsd-5.6-amd64.egg to /home/landry/.local/lib/python2.7/site-packages
psutil 3.0.0 is already the active version in easy-install.pth

Installed /home/landry/.local/lib/python2.7/site-packages/psutil-3.0.0-py2.7-openbsd-5.6-amd64.egg
Processing dependencies for psutil==3.0.0
Finished processing dependencies for psutil==3.0.0
python2.7 test/test_psutil.py
Traceback (most recent call last):
  File "test/test_psutil.py", line 49, in <module>
    import psutil
  File "/home/landry/.local/lib/python2.7/site-packages/psutil-3.0.0-py2.7-openbsd-5.6-amd64.egg/psutil/__init__.py", line 184, in <module>
    raise ImportError(msg)
ImportError: version conflict: '/home/landry/.local/lib/python2.7/site-packages/psutil-3.0.0-py2.7-openbsd-5.6-amd64.egg/psutil/_psutil_bsd.so' C extension module was built for another version of psutil (different than 3.0.0)

```

landryb:
Fixed in 3432cee, now i'm at

```
FAILED (failures=12, errors=34, skipped=67)
```

Which i think isnt so bad. That's what happens when stuff bitrots and isnt merged quickly...

landryb:
And after my last commit in 54fa691, it also works on OpenBSD/i386 (32 bits)

```
FAILED (failures=12, errors=37, skipped=67)
```

giampaolo:
[...]

PR #615, opened 2015-04-05
Openbsd support (#562)

landryb:
Isnt perfect yet, but would be nice to merge early so that the work isnt lost. Basic functionality needed by end users work...

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -123,7 +123,7 @@ elif sys.platform.startswith("freebsd"):
     from . import _psbsd as _psplatform
 
 elif sys.platform.startswith("openbsd"):
-    import psutil._psbsd as _psplatform
+    from . import _psbsd as _psplatform
 
 elif sys.platform.startswith("sunos"):
     from . import _pssunos as _psplatform
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -151,6 +151,9 @@ elif sys.platform.startswith("darwin"):
 elif sys.platform.startswith("freebsd"):
     import psutil._psbsd as _psplatform
 
+elif sys.platform.startswith("openbsd"):
+    import psutil._psbsd as _psplatform
+
 elif sys.platform.startswith("sunos"):
     import psutil._pssunos as _psplatform
     from psutil._pssunos import (CONN_IDLE,  # NOQA
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -9,6 +9,7 @@
 import errno
 import functools
 import os
+import sys
 import xml.etree.ElementTree as ET
 from collections import namedtuple
 
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -134,6 +134,8 @@ def cpu_count_physical():
     # We may get None in case "sysctl kern.sched.topology_spec"
     # is not supported on this BSD version, in which case we'll mimic
     # os.cpu_count() and return None.
+    if sys.platform.startswith("openbsd"):
+        return cext.cpu_count_logical()
     s = cext.cpu_count_phys()
     if s is not None:
         # get rid of padding chars appended at the end of the string
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -81,6 +81,8 @@ def virtual_memory():
 
 def swap_memory():
     """System swap memory as (total, used, free, sin, sout) namedtuple."""
+    if sys.platform.startswith("openbsd"):
+        PAGESIZE = 1
     total, used, free, sin, sout = [x * PAGESIZE for x in cext.swap_mem()]
     percent = usage_percent(used, total, _round=1)
     return _common.sswap(total, used, free, percent, sin, sout)
--- a/psutil/_psutil_openbsd.c
+++ b/psutil/_psutil_openbsd.c
@@ -1,0 +1,2212 @@
+/*
+ * Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ *
+ * FreeBSD platform-specific module methods for _psutil_bsd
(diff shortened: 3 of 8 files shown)


benoitc:
+1

giampaolo:
I took a look back at this and updated code:
https://github.com/giampaolo/psutil/tree/landryb-openbsd
What's missing:
- `Process.memory_info().vms` (aka `kp.p_vm_map_size`) is always 0. "ps" manages to show VMS somehow; as a last resort we might parse ps output.
- `Process.open_files()`'s path is an empty string; `kif->f_mntonname` refers to the mountpoint (useless), `kif->unp_path` is always empty. `lsof` somehow manages to do it.
- `Process.cwd()` is not implemented; couldn't find any way to determine this. `lsof` somehow manages to get this.
- `Process.memory_maps()` is not implemented.
- `Process.threads()` returns EINVAL.
- `Process.connetions()` and `net_connections()`'s `status` is always `None`
- `Process.connetions()` and `net_connections()`'s `addr` is always `None` for IPv6 sockets.
- `Process.cmdline()` does not return the actual cmdline; we get things such as `['pflogd: [running] -s 160 -i pflog0 -f /var/log/pflog']`, `['sshd: user@notty']` and `['smtpd: scheduler']`.
- `Process.num_ctx_switches()`'s `involuntary` field is always `0`; I suppose this means it is just not supported on OpenBSD so it's OK to skip this test.

giampaolo:
`Process.cwd()` implemented in https://github.com/giampaolo/psutil/commit/04102058124bc41a0630a5d6e5d8f22eb035e26d

giampaolo:
Process vms fixed in https://github.com/giampaolo/psutil/commit/eeac17f3b01bd82c61a3067558a2e559ba358ba6.

giampaolo:
Process threads() implemented in https://github.com/giampaolo/psutil/commit/67da39e5e7b6e6ecd8ceef00db6f028ffd3ce6e1.

giampaolo:
Process `num_threads()` implemented in https://github.com/giampaolo/psutil/commit/2ce156bc2540598a91e9c311eb1d553143d7b3aa.

landryb:
Thanks for picking up this, i admit it feel out of the radar since i had no time for it...

giampaolo:
OK, I merged everything in here: https://github.com/giampaolo/psutil/pull/709
Last remaining issues are:
- `Process.memory_maps()` is not supported
- `Process.open_files()` return empty paths (also FreeBSD has this problem)
- `Process.exe()` is not fully implemented and currently relies on `cmdline()`

landryb:
as i already said several times, open_files() will never get implemented, the kernel won't provide such an API.

landryb:
But yay for seeing this merged \o/

giampaolo:
> as i already said several times, open_files() will never get implemented, the kernel won't provide such an API.

Do you know if this is a kernel bug? We have the same problem on FreeBSD which *does* provide the necessary API but the path sometimes is empty. Also OpenBSD should provide it (`kif->unp_path`) and it's empty. Any internet reference which explains any of this is welcome.

giampaolo:
@landryb  can you please comment about this? https://github.com/giampaolo/psutil/blob/fc1e59d08c968898c2ede425a621b62ccf44681c/psutil/arch/bsd/openbsd.c#L743

landryb:
> Do you know if this is a kernel bug? We have the same problem on FreeBSD which does provide the necessary API but the path sometimes is empty. Also OpenBSD should provide it (kif->unp_path) and it's empty. Any internet reference which explains any of this is welcome.

This is not a bug, it's on purpose for 'privacy reasons'. As for ```kif->unp_path```, it's not filled much, see http://bxr.su/search?q=unp_path&project=OpenBSD

The only information you can get is the inode (see http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man1/fstat.1 from which you can list open inodes for a process), from which you *could* theoretically do lookups in the filesystem but this is cumbersome.

```
$fstat -p 21801   
USER     CMD          PID   FD MOUNT        INUM MODE       R/W    SZ|DV
landry   xfce4-session 21801 text /         1147672 -r-xr-xr-x   r   189368
landry   xfce4-session 21801   wd /home    29491840 drwxr-xr-x   r     7680
landry   xfce4-session 21801    0 /         1716753 crw-rw-rw-  rw     null
landry   xfce4-session 21801    1 /home    29496358 -rw-r--r--   w   407233
landry   xfce4-session 21801    2 /home    29496358 -rw-r--r--   w   407233
landry   xfce4-session 21801    3 /          572089 -rw-r--r--   w     2643
landry   xfce4-session 21801    4* unix stream 0x0
landry   xfce4-session 21801    5 pipe 0x0 state: 
landry   xfce4-session 21801    6 pipe 0x0 state: 
landry   xfce4-session 21801    7* unix stream 0x0
landry   xfce4-session 21801    8* unix stream 0x0
landry   xfce4-session 21801    9 /home    29492111 -rw-r--r--   w    11834
landry   xfce4-session 21801   10 pipe 0x0 state: 
landry   xfce4-session 21801   11 pipe 0x0 state: 
landry   xfce4-session 21801   12* unix stream 0x0
landry   xfce4-session 21801   13* unix stream 0x0
landry   xfce4-session 21801   14* unix stream 0x0
landry   xfce4-session 21801   15* unix stream 0x0
landry   xfce4-session 21801   16* unix stream 0x0
landry   xfce4-session 21801   17* unix stream 0x0

```

> @landryb can you please comment about this? https://github.com/giampaolo/psutil/blob/fc1e59d08c968898c2ede425a621b62ccf44681c/psutil/arch/bsd/openbsd.c#L743

struct diskstats doesnt provide a breakdown between read and write time i think, see
http://bxr.su/OpenBSD/sys/sys/disk.h#62

landryb:
As for the open files, the OS providing this information have to keep a cache of fd/inode/path around, and update it if the file is moved/hardlinked/etc - OpenBSD just doesnt do this.

Issue #611, opened 2015-03-28
psutil_net_io_counters on Solaris has send and received swapped


grobian:
Python code:
```
def net_io_counters(pernic=False):
    """Return network I/O statistics as a namedtuple including
    the following fields:

     - bytes_sent:   number of bytes sent
     - bytes_recv:   number of bytes received
     - packets_sent: number of packets sent
     - packets_recv: number of packets received
```

C-code however, has them swapped (psutil/_psutil_sunos.c):
```
        py_ifc_info = Py_BuildValue("(KKKKkkii)",
                                    rbytes->value.ui64,
                                    wbytes->value.ui64,
                                    rpkts->value.ui64,
                                    wpkts->value.ui64,
```

I can come up with a simple patch, but I figure it's obvious what needs to be changed :)

giampaolo:
Please submit a pull request (and also update HISTORY.rst and CREDITS.rst).

PR #617, opened 2015-04-16
psutil_net_io_counters: fix swapped send and receive

grobian:
On SunOS/Solaris the bytes and packets send/received were swapped.
Fixed it.

Issue #611


giampaolo:
Thanks a lot!

Issue #618, opened 2015-04-17
swap tests fail on solaris when run as normal user


grobian:
The code relies on the tool "swap", which lives in /usr/sbin/.

PR #619, opened 2015-04-17
swap: fix finding the 'swap' utility on SunOS

grobian:
This is a bit of a kludge, but allows to find the swap utility when
invoced by normal users which shouldn't have the sbin paths in PATH by
default.  One can argue the paths should be appended instead of
prepended, but I doubt whether one wants to use another swap utility,
since the code relies on its output format.

Fixes issue #618.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -292,4 +292,4 @@ E: spacewanderlzx@gmail.com
 I: 555
 
 N: Fabian Groffen
-I: 611
+I: 611, 618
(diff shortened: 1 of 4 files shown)


grobian:
this will conflict with https://github.com/Infinidat/psutil/commit/03c271aaf49e84dbd75e69ebe3c4238674476aec but that change (-k removal) will still be necessary.

giampaolo:
Thanks for this.

Issue #620, opened 2015-04-17
test_cpu_count fails on SunOS: no such file /proc/cpuinfo


grobian:
The test obviously fails, could try to use nproc from gnu-coreutils, but more likely on Solaris /usr/sbin/psrinfo (-pv) should be used to get the number of physical and virtual processors.

grobian:
https://github.com/Infinidat/psutil/commit/5c28b2b16f3a1fcc4683d21d762504d85dfa79e7
I thought about the alternative if LINUX: but I wonder what the intention of the test was, it looks like it tries to avoid a false positive, question is if that can happen on non-Linux systems (and how to detect it).

giampaolo:
Yeah, the real intent of that test should be to run on Linux only so the `if` condition should be:

    if LINUX and os.path.exists("/proc/cpuinfo"):
         ...

giampaolo:
This is supposed to have been fixed by #610.

PR #610, opened 2015-03-22
fix build and tests on Solaris 10

wiggin15:

PR #621, opened 2015-04-17
test_cpu_count: fix dependency on /proc/cpuinfo

grobian:
Fixes issue #620

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -292,4 +292,4 @@ E: spacewanderlzx@gmail.com
 I: 555
 
 N: Fabian Groffen
-I: 611, 618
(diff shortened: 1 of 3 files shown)


giampaolo:
There's no need to update HISTORY.rst for this.

wiggin15:
Hi, I just wanted to mention that there's a fix for this (and other tests on Solaris) in #610.

grobian:
that would be 5c28b2b16f3a1fcc4683d21d762504d85dfa79e7

giampaolo:
This is supposed to have been fixed by #610.

PR #633, opened 2015-06-16
Fixing cmdline() parsing on linux

arnimarj:
I had problems getting the correct Process.cmdline() when some of the parameters were the empty string. E.g. when running this script:

```python
import psutil, os, sys

pid = os.getpid()

for p in psutil.process_iter():
	if p.pid == os.getpid():
		break

	p = None

A = sys.argv[:]
B = p.cmdline()[1:]

print A
print B
print A == B```

A and B would not be the same when running ```python ../proc.py '' ''```

It seems that /proc/{pid}/cmdline contains strings separated by the zero byte, along with a final zero byte: http://linux.die.net/man/5/proc

This patch seems to fix the issue.


giampaolo:
duplicate of #634

PR #634, opened 2015-06-16
fixing cmdline() for empty-string argumemts on posix systems

arnimarj:


giampaolo:
Mmmm for extra safety I would do this instead:

```
    @wrap_exceptions
    def cmdline(self):
        fname = "/proc/%s/cmdline" % self.pid
        kw = dict(encoding=DEFAULT_ENCODING) if PY3 else dict()
        with open(fname, "rt", **kw) as f:
            data = f.read()
        if data.endswith('\x00'):  # this should always be true
            data = data[:-1]
        return [x for x in data.split('\x00') if x]
``` 

Please change this, and also:
- make test_psutil.py PRP8 compliant (new line are longer than 80 chars)
- update HISTORY.rst
- update CREDITS
...then it's ready to be merged.

giampaolo:
@arnimarj : it turns out this caused https://github.com/giampaolo/psutil/issues/639. I should have fixed it in https://github.com/giampaolo/psutil/commit/3a620d94b347a7044e1bd6c6aabcf5528c3089c3. Can you please confirm this works as expected?

Issue #550, opened 2014-11-19
Enhancement: get threads per core


linlinlinlin:
Is it possible add a feature: get thread(s) count per core

linlinlinlin:
Thanks for your reply.
In my device, it have one logical CPU and have two cores.Each cores have **2** threads.
So, the result is
```
>>> psutil.cpu_count(logical=False)
1
>>> psutil.cpu_count(logical=True)
4
```
So, now I am parsing `/proc/cpuinfo` by myself.

giampaolo:
Ah I see what you mean. Then no, right now you can't determine that kind of info with psutil. But why do you need to know that?

mrvn:
It would be nice to get the counts for sockets (cpus), cores and threads. But I don't think a simple count() will do. Potentially you can have multiple cpus with different core and thread counts. And some cores or threads can be powered down.

So maybe it would make sense (under linux) to parse /proc/cpuinfo and/or /sys/bus/cpu/devices/cpu*/topology/* and generate a tree-ish structure. Top level would be the sockets (physical_package_id), next level the cores and the threads as leaves. The top level would have cpu, core and thread counts. The lower levels would have more specific counts. E.g. cpu_tree()[0][2].thread_count() would be the number of threads on core 2 in physical package 0.

At that point it would be nice to include the cpu frequency as well. But maybe that's better left to another issue.

giampaolo:
I think there's already a lot of confusion around the concept of CPU cores, sockets, non/logical cores etc. and how they differ. 
I think it's safe to assume that what users need most of the times is to figure out the number of all `logical` cores in order to spawn exactly that many subprocesses. 
psutil provides that as `psutil.cpu_count(logical=True)`, which is the same provided by stdlib `multiprocessing.cpu_count()` and `os.cpu_count()`.
psutil pushes that a little further and also provides `psutil.cpu_count(logical=False)` which returns the number of physical cores only (hyper thread CPUs are excluded).
I'm not sure how many users need that but I think they are not many as it's a very specific use case already.

With that said, to me returning per CPU sockets, threads or even frequency is even more specific and potentially useless as I expect very few people would be using that and the people NOT using those things will be even more confused. 
Last but not least this may be easy to retrieve on Linux but I have no idea about the other platforms.

My main concern remains the use case though: except for some very specific circumstances my feeling is that nobody would be needing this.

mrvn:
[...]

PR #643, opened 2015-07-09
proof-of-concept for cpu topology tree

mrvn:
This is just a quick proof-of-concept for getting the topology of the cpus in a system in tree form.
Its only to illustrate my comment in issue #550 "Enhancement: get threads per core" and only for Linux, not to be actually merged yet.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -151,6 +151,7 @@ __all__ = [
     "pid_exists", "pids", "process_iter", "wait_procs",             # proc
     "virtual_memory", "swap_memory",                                # memory
     "cpu_times", "cpu_percent", "cpu_times_percent", "cpu_count",   # cpu
+    "cpu_tree",
     "net_io_counters", "net_connections", "net_if_addrs",           # network
     "net_if_stats",
     "disk_io_counters", "disk_partitions", "disk_usage",            # disk
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1363,6 +1363,20 @@ def cpu_count(logical=True):
         return _psplatform.cpu_count_physical()
 
 
+@memoize
+def cpu_tree():
+    """Return the topology of CPUs in the system in tree form.
+
+    Return None if undetermined.
+
+    The return value is cached after first call.
+    If desired cache can be cleared like this:
+
+    >>> psutil.cpu_tree.cache_clear()
+    """
+    return _psplatform.cpu_tree()
+
+
 def cpu_times(percpu=False):
     """Return system-wide CPU times as a namedtuple.
     Every CPU time represents the seconds the CPU has spent in the given mode.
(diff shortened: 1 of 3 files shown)


SimonHeimberg:
On my linux machine, this would work perfect.

Not sure if it would work on every system. I found some formats which would not:
https://unix.stackexchange.com/questions/255558/understanding-the-output-of-proc-cpuinfo (AArch64, in 1 section, 7 years old)
https://github.com/pytorch/cpuinfo/blob/main/test/cpuinfo/xiaomi-redmi-note-3.log (Aarch64, in 1 section)
https://github.com/pytorch/cpuinfo/blob/main/test/cpuinfo/arndaleboard.log has additional section
https://github.com/pytorch/cpuinfo/blob/main/test/cpuinfo/raspberrypi3.log has additional section
https://github.com/pytorch/cpuinfo/blob/main/test/cpuinfo/atm7029b-tablet.log (ARMv7) has additional section

All examples would work when a new section would be started on `^processor:[ \t]*[0-9]`, but some values would have to be adapted.

SimonHeimberg:
More concerning **linux**:

Looks like /proc/cpuinfo is meant for human consumption and thus lacks a stable format (from https://stackoverflow.com/a/23378780). Makes it hard to use.
`lscpu -p` gives the topology information (and allows to select columns with -p=col1,col2), but is not available on every system. (from same so question)

``lscpu -p also shows the tree structure for linux: CPU,Core,Socket,Node (Node is missing in your code. No idea yet what a node is.)

Some inspiration might come from [lscpu from linux-utils](https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/sys-utils/lscpu-topology.c?h=stable/v2.39#n92). It uses information from `/sys/devices/system/cpu/cpu[0-9]*/topology/*_siblings` and from `/proc/cpu_info`. Their advantage is that the development is together with the kernel version, so they know what virtual files are around.

*Other* os:
seems to be difficult on FreeBsd: https://stackoverflow.com/questions/4083848/what-is-the-equivalent-of-proc-cpuinfo-on-freebsd-v8-1
windows maybe https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformation

SimonHeimberg:
Wondering if another structure would be more useful for many cases. Functions like `os.sched_getaffinity()` or `psutil.Process().cpu_affinity()` work with the id of the logical cpu. Taking the (logical) cpu id as base, we get something like

```python
(LogicalCpu(cpu_id=0, core_id=0, socket_id=0, node_id=0, l1d_id=0, l1i_id=0, l2_id=0, l3_id=0, online='Y'),
 LogicalCpu(cpu_id=1, core_id=0, socket_id=0, node_id=0, l1d_id=0, l1i_id=0, l2_id=0, l3_id=0, online='Y'),
 LogicalCpu(cpu_id=2, core_id=1, socket_id=0, node_id=0, l1d_id=1, l1i_id=1, l2_id=1, l3_id=0, online='Y'),
...
```
with `LogicalCpu = collections.namedtuple("LogicalCpu", ("cpu_id", "core_id", ...)` (or a class)

The tree structure makes sense for other information (vendor, model name, flags), and for understanding how the system looks like (it took me a long time).

Issue #641, opened 2015-07-09
(Windows) fix compiler warnings


giampaolo:
```
psutil/_psutil_windows.c(1329) : warning C4995: '_stprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(1329) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2250) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2304) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2315) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2456) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2461) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2464) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2467) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2472) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2474) : warning C4995: 'strcat': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2585) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2940) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2948) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(2952) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(3118) : warning C4995: 'sprintf': name was marked as #pragma deprecated
psutil/_psutil_windows.c(3120) : warning C4995: 'sprintf': name was marked as #pragma deprecated
```

giampaolo:
@mrjefftang maybe you want to take a look at this?

mrjefftang:
I'm not sure when I'll get around to working on this but it's just a matter of using `sprintf_s` and passing in the buffer size.

PR #654, opened 2015-07-14
Initial pass through with safe strcat/sprintf/stprintf

mrjefftang:
Addresses #641 

Currently work in progress

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -255,7 +255,7 @@ I: 492
 
 N: Jeff Tang
 W: https://github.com/mrjefftang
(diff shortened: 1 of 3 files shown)


mrjefftang:
I think this solves the current issue at hand.

There's a larger overarching issue which is inconsistent use of TCHAR, char[], WCHAR, etc.

giampaolo:
Cool. Thanks. Please update HISTORY/CREDITS and this is good to go.
For the other issue, can you please file a new ticket?

mrjefftang:
Done and done

PR #660, opened 2015-08-02
Improve make.bat's resiliency to Visual Studio install locations

mpderbec:
This is a small change that allows make.bat to respond to non-standard Visual Studio install locations.

The VS90COMNTOOLS variable is always defined globally if Visual Studio 9.0 is installed.

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -26,6 +26,8 @@ if "%TSCRIPT%" == "" (
     set TSCRIPT=test\test_psutil.py
 )
 
+set VSINSTALLDIR=%VS90COMNTOOLS%..\..
+
 set PYTHON26=C:\Python26\python.exe
 set PYTHON27=C:\Python27\python.exe
 set PYTHON33=C:\Python33\python.exe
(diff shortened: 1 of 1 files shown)


mrjefftang:
This should be dependent on the version of Python.

Visual studio 2008 for Python 2.6, 2.7, 3.2
Visual studio 2010 for Python 3.1 - 3.4
Visual studio 2015 for Python 3.5+

mpderbec:
I'm new to this codebase, so I might not understand exactly how it works. My changes are meant to preserve the status quo -- as you can see in the old version of the batch file, it is always using VS2008 (9.0). So the only thing this PR does is accommodate users that may have installed VS2008 to somewhere other than C:\Program Files (x86)\Microsoft Visual Studio 9.0.

I wouldn't feel comfortable making the changes you propose -- at least not in this particular PR -- because it would significantly change current behavior and I would fear breaking something in the process.

Again-- I may not understand the situation well enough so please let me know if I'm off base.

giampaolo:
I tested this works as expected. Thanks (I also added you to CREDITS).

Issue #648, opened 2015-07-11
Add continuous test integration for OSX


giampaolo:
Similarly to #646 but for OSX. Still have to investigate whether there are public free services providing this.

giampaolo:
http://www.yegor256.com/2014/10/05/ten-hosted-continuous-integration-services.html
Apparently the only solution is https://hosted-ci.com/ but it's not free. https://ship.io/ does not seem to support OSX, only IOS.

mrjefftang:
Travis CI supports it but they are full currently.  You may want to get in contact with them for since psutil is an open source project.

giampaolo:
from http://docs.travis-ci.com/user/multi-os/

> We are at capacity with Mac workers in both .org and .com environments. Consequently, we are not 
> accepting requests for the multi-OS feature at this time. As more capacity is added, we will resume 
> accepting requests.

I'm gonna try to send them an email anyway.

giampaolo:
See comments in: https://github.com/travis-ci/travis-ci/issues/2312#issuecomment-122850231

giampaolo:
It seems this project is using Travis + OSX by using `language: generic`. AFAIU the trick should be not using `language: python`, then use `os: osx` and for that install python separately.

mrjefftang:
Figuring out the correct Travis settings will be a matter of trial and error as long as we can get it access to Travis's OSX builders.

giampaolo:
Fixed by @mrjefftang. Closing.

PR #662, opened 2015-08-03
Test OSX Travis config

mrjefftang:


mrjefftang:
This should take care of #648.  There's probably some cleanup that can be done to make it look nicer and flow with the Linux requirements.

giampaolo:
This is awesome! Thanks a lot Jeff.

PR #672, opened 2015-08-21
Prevent enum clash when using Windows SDK v8.0

stevenwinfield:
ProcessDebugPort and ProcessImageFileName are defined in PROCESSINFOCLASS in winternl.h - use the same workaround as ProcessBasicInformation, with underscores and #defines.


giampaolo:
Merged, thanks.

PR #680, opened 2015-09-06
Fix #517 (net_io_counters on Solaris 10)

### LINKED ISSUE #517: Problems with network statistics on Solaris 10
We’ve been evaluating psutil for suitability in a cross-platform machine statistic gathering application that we maintain for internal use. We’ve had several issues on Solaris and the attached patch file contains changes that made to get it working on Solaris 10.
 
* We didn’t have any kstat network results where ks_module == “link”, so we hacked it to look for ks_name == “mac”. There is probably a better solution. The “kstat -c net”  output from several of our servers is linked below for reference. Our servers have multi-port NIC cards, so we needed to mangle ks_name + ks_instance as the key of the returned dictionary.
* The code was compiled as 32-bit code to match our Python installation. The #if code assumes that the kstat statistics will match the number of bits used to compile. That is not a valid assumption. A hack was made to check at runtime.

This patch is not ready to merge, but I would be happy to get feedback to make it acceptable.
 
Patch - https://gist.github.com/lymanepp/c5e403df4cbd73561d9a
Kstat output (server 1) - https://gist.github.com/lymanepp/2d105d01da798760f2c5
Kstat output (server 2) - https://gist.github.com/lymanepp/2b6c103ff97fdf5995f2

wiggin15:
During kstat enumeration in net_io_counters on Solaris, we check for class=net and module=link to find kstats for network interfaces. As reported in #517 (example kstat output is submitted there), module is not always 'link' so on Solaris 10 we don't find any interfaces.
net_if_stats does the same enumeration but there is no problem there because it doesn't check the module, and instead sends a ioctl and decides if the current stat name is a network interface name according to the result. I copied the same logic from there to net_io_counters.
Also fixed 2 minor bugs along the way that would only happen in error cases.

The second commit is modelled after parts of the patch in #517 but also fixes the data types passed to Py_BuildValue.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,7 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 
 - #677: [Linux] can't install psutil due to bug in setup.py.
 - #610: [SunOS] fix build and tests on Solaris 10
+- #517: [SunOS] fix net_io_counters on Solaris 10
 
 
 3.2.0 - 2015-09-02
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,7 +7,8 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 
 - #677: [Linux] can't install psutil due to bug in setup.py.
 - #610: [SunOS] fix build and tests on Solaris 10
-- #517: [SunOS] fix net_io_counters on Solaris 10
+- #517: [SunOS] net_io_counters failed to detect network interfaces
+  correctly on Solaris 10
 
 
 3.2.0 - 2015-09-02
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -728,26 +728,32 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
             goto error;
         }
 
-#if defined(_INT64_TYPE)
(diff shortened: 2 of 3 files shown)


HISTORY.rst
```diff
@@ -7,6 +7,7 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 
 - #677: [Linux] can't install psutil due to bug in setup.py.
 - #610: [SunOS] fix build and tests on Solaris 10
+- #517: [SunOS] fix net_io_counters on Solaris 10
```
giampaolo: please be more specific (describe what issue is being addressed)

giampaolo:
Does this fix #517?

psutil/_psutil_sunos.c
```diff
@@ -685,25 +688,30 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
     if (kc == NULL)
         goto error;
 
+    sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1)
+        goto error;
```
giampaolo: You need to `PyErr_SetFromErrno(PyExc_OSError)` before `goto error;`.

wiggin15:
Yes, it does.
I fixed what you wrote in the comments. Thanks.

giampaolo:
Thank you!

PR #689, opened 2015-09-26
specify encoding for py3k stdout (utf-8)

msarahan:
Fixes errors raised during testing (sys.stdout.encoding is None for some reason)

### DIFF
--- a/test/test_psutil.py
+++ b/test/test_psutil.py
@@ -204,7 +204,7 @@ def sh(cmdline, stdout=subprocess.PIPE, stderr=subprocess.PIPE):
     if stderr:
         warn(stderr)
     if PY3:
-        stdout = str(stdout, sys.stdout.encoding)
+        stdout = str(stdout, sys.stdout.encoding or sys.getfilesystemencoding() or 'utf-8')
     return stdout.strip()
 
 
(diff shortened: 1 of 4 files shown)


giampaolo:
I think it makes more sense to use sys.getfilesystemencoding() instead of "utf8". Can you please update your PR including changes to HISTORY and CREDITS files?

msarahan:
Thanks for reviewing this.

A contributor to Conda, @jakirkham, came up with a better fix: https://github.com/conda/conda-build/pull/614

That style of fix is applied here, as well.  I have updated HISTORY and CREDITS for this PR.

test/test_psutil.py
```diff
@@ -204,7 +204,7 @@ def sh(cmdline, stdout=subprocess.PIPE, stderr=subprocess.PIPE):
     if stderr:
         warn(stderr)
     if PY3:
-        stdout = str(stdout, sys.stdout.encoding)
+        stdout = str(out, sys.stdout.encoding or sys.getfilesystemencoding() or 'utf-8')
```
jakirkham: Sorry, I took a look as I got pinged and was curious about where the problem arose here.

It looks like this change from `stdout` to `out` was unintentional.

msarahan: Thanks!  Good catch, clumsy cut/paste.

jakirkham: Glad to help. No worries. Happens to all of us.

PR #696, opened 2015-10-05
Use pargs on Solaris to get full command line

### LINKED ISSUE #694: Process cmdline is truncated on Solaris
`p.cmdline()` is truncated to 80 characters on Solaris 11. `/proc/PID/psinfo` that psutil uses to read the command line contains only the beginning - See e.g. https://unix.stackexchange.com/questions/3934/how-can-i-get-a-full-process-list-in-solaris-without-truncated-lines.
The solution would be to use the same method as `pargs` tool - https://github.com/illumos/illumos-gate/blob/09f79f7c66b85f056db11f58210dc6182c9b1aef/usr/src/cmd/ptools/pargs/pargs.c. It's not however clear to me how can it be easily ported as it's using `libproc`, which is documented in https://github.com/illumos/illumos-gate/blob/09f79f7c66b85f056db11f58210dc6182c9b1aef/usr/src/lib/libproc/common/libproc.h as:

```
 * In the meantime, be aware that any program linked with libproc in this
 * release of Solaris is almost guaranteed to break in the next release.
 *
 * In short, do not use this header file or libproc for any purpose.
```

As a workaround `cmdline()` could try to launch `pargs` if available but it bears a `fork` cost.

Didn't test any other Solaris version.

gozdal:
Fixes https://github.com/giampaolo/psutil/issues/694

### DIFF
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -270,6 +270,8 @@ def wrap_exceptions(fun):
 class Process(object):
     """Wrapper class around underlying C implementation."""
 
+    PARGS_BINARY = '/usr/bin/pargs'
+
     __slots__ = ["pid", "_name", "_ppid"]
 
     def __init__(self, pid):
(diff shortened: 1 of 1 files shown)


giampaolo:
I am sorry but I am going to reject this. Parsing another process output is a last resort otherwise it should be done in C as this appears to be the case.

gozdal:
No problem, though I lack the expertise to do it properly C. Maybe an update to documentation to warn potential users would be helpful?

PR #715, opened 2015-12-03
import psutil raises IOError on Linux when /proc/stat is inaccessible

sethp-jive:
Hello! I found a few issues trying to take advantage of the new relocatable procfs, predominately that trying to import the psutils module on a system with no /proc/stat file fails. Additionally, I found that attempting to reload the psutils module (which is very useful for testing import-time issues) on linux caused some trouble. 

Please see the individual commit messages for more details. Thanks!

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -107,7 +107,6 @@ if sys.platform.startswith("linux"):
             RLIMIT_SIGPENDING = _psutil_linux.RLIMIT_SIGPENDING
         except AttributeError:
             pass
-        del _psutil_linux
 
 elif sys.platform.startswith("win32"):
     from . import _pswindows as _psplatform
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1411,9 +1411,13 @@ def cpu_times(percpu=False):
     else:
         return _psplatform.per_cpu_times()
 
-
-_last_cpu_times = cpu_times()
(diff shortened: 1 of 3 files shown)


psutil/__init__.py
```diff
@@ -1522,8 +1525,8 @@ def cpu_times_percent(interval=None, percpu=False):
     def calculate(t1, t2):
         nums = []
         all_delta = sum(t2) - sum(t1)
-        for field in t1._fields:
-            field_delta = getattr(t2, field) - getattr(t1, field)
+        for field in t2._fields:
```
giampaolo: why did you change from `t1` to `t2`?

giampaolo:
I reorganized the import machinery, please see: https://github.com/giampaolo/psutil/commit/d8327888261aa8d43d3df470fb5eae58ab12bad3.
With that this PR is no longer valid but I think we can include those 2 tests you added. Can you open another PR for that?

psutil/__init__.py
sethp-jive: `t1` is the least-recent set of data points, so if the underlying /proc/stat file changes to provide a different number of fields (e.g. if we go from not having fields to having some fields) then `t2` will have the latest and most correct set of fields.

The danger is that `t1` will be missing some values, hence the default attr provided on the next line.

sethp-jive:
Indeed I will – give me a few minutes and I should have it ready.

PR #719, opened 2015-12-15
Add import-time tests for linux psutil

sethp-jive:
Two new tests that validate:

1) that psutil can be sucessfuly reload()'d on linux
2) that a missing procfs doesn't prevent importing, or
   correct functionality.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -108,7 +108,6 @@ if sys.platform.startswith("linux"):
             RLIMIT_SIGPENDING = _psutil_linux.RLIMIT_SIGPENDING
         except AttributeError:
             pass
-        del _psutil_linux
 
 elif sys.platform.startswith("win32"):
     from . import _pswindows as _psplatform
--- a/test/_linux.py
+++ b/test/_linux.py
@@ -15,6 +15,7 @@ import io
 import os
 import pprint
 import re
+import shutil
 import socket
 import struct
 import sys
@@ -34,6 +35,7 @@ from psutil._compat import u
 from test_psutil import call_until
 from test_psutil import get_kernel_version
 from test_psutil import get_test_subprocess
+from test_psutil import importlib
 from test_psutil import LINUX
 from test_psutil import MEMORY_TOLERANCE
 from test_psutil import POSIX
@@ -43,7 +45,6 @@ from test_psutil import skip_on_not_implemented
 from test_psutil import TRAVIS
 from test_psutil import unittest
 from test_psutil import which
-
 
 HERE = os.path.abspath(os.path.dirname(__file__))
 # procps-ng 3.3.10 changed the output format of free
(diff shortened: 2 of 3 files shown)


psutil/__init__.py
```diff
@@ -1545,8 +1544,8 @@ def cpu_times_percent(interval=None, percpu=False):
     def calculate(t1, t2):
         nums = []
         all_delta = sum(t2) - sum(t1)
-        for field in t1._fields:
-            field_delta = getattr(t2, field) - getattr(t1, field)
+        for field in t2._fields:
+            field_delta = getattr(t2, field) - getattr(t1, field, 0)
```
giampaolo: I am afraid of this change. I think it makes more sense to set `_last_cpu_times` and `_last_per_cpu_times` to None in case they cannot be determined at import time, and take that into account in this function. I just did that in e40efe13102390aa769b0de65877da57c0b3429d. Please update your branch.

test/_linux.py
```diff
@@ -472,6 +480,73 @@ def test_procfs_path(self):
             psutil.PROCFS_PATH = "/proc"
             os.rmdir(tdir)
 
+    def test_psutil_is_reloadable(self):
+        imp.reload(psutil)
```
giampaolo: Please move this test in `test_psutil.py` (also imports).

test/_linux.py
```diff
@@ -472,6 +480,73 @@ def test_procfs_path(self):
...
+    def test_no_procfs_for_import(self):
+        my_procfs = tempfile.mkdtemp()
+
+        with open(os.path.join(my_procfs, 'stat'), 'w') as f:
+            f.write('cpu   0 0 0 0 0 0 0 0 0 0\n')
+            f.write('cpu0  0 0 0 0 0 0 0 0 0 0\n')
+            f.write('cpu1  0 0 0 0 0 0 0 0 0 0\n')
+
+        self.assertNotAlmostEqual(psutil.cpu_percent(), 0)
+        self.assertNotAlmostEqual(sum(psutil.cpu_times_percent()), 0)
```
giampaolo: What are you trying to test here? Why did you use `assertNotAlmostEqual` and not `assertNotEqual`?

test/_linux.py
```diff
@@ -472,6 +480,73 @@ def test_procfs_path(self):
...
+                self.assertRaises(IOError, psutil.cpu_percent)
+                self.assertRaises(IOError, psutil.cpu_percent, percpu=True)
+                self.assertRaises(IOError, psutil.cpu_times_percent)
+                self.assertRaises(
+                    IOError, psutil.cpu_times_percent, percpu=True)
+
+                psutil.PROCFS_PATH = my_procfs
+
+                self.assertAlmostEqual(psutil.cpu_percent(), 0)
+                self.assertAlmostEqual(sum(psutil.cpu_times_percent()), 0)
```
giampaolo: same here

test/_linux.py
```diff
@@ -472,6 +480,73 @@ def test_procfs_path(self):
...
+                self.assertNotAlmostEqual(
+                    sum(psutil.cpu_percent(percpu=True)), 0)
+                self.assertNotAlmostEqual(sum(psutil.cpu_times_percent()), 0)
+                self.assertNotAlmostEqual(
+                    sum(map(sum, psutil.cpu_times_percent(percpu=True))), 0)
+        finally:
+            shutil.rmtree(my_procfs)
+            imp.reload(psutil)
+
+        assert psutil.PROCFS_PATH == '/proc'
```
giampaolo: Please use `self.assertEqual` instead.

test/_linux.py
```diff
@@ -472,6 +480,73 @@ def test_procfs_path(self):
...
+
+        self.assertNotAlmostEqual(psutil.cpu_percent(), 0)
+        self.assertNotAlmostEqual(sum(psutil.cpu_times_percent()), 0)
+        try:
+            orig_open = open
+
+            def open_mock(name, *args):
+                if name.startswith('/proc'):
+                    # simulate an ENOENT
+                    raise IOError('rejecting access to /proc')
```
giampaolo: You can `raise IOError(errno.ENOENT, "")` instead.

psutil/__init__.py
sethp-jive: Perhaps it's better to think of this as an `izip_longest` filling the shorter CPU values tuple with 0s in missing slots? 

In a any event, I've updated the PR to revert this change – I'll hopefully be able to take a look at your other comments within a few hours.

Thanks for the help & feedback! :)

test/_linux.py
sethp-jive: There I was checking that my test further down would be valid – if the system as a whole reported no CPU usage (which seemed unlikely), then the `assertAlmostEqual` against `0` could spuriously pass even if it weren't using `my_procfs`. 

It's using `assertNotAlmostEqual` because the cpu times are reported as floats, and in the spurious pass case it might be that the system reported e.g. `0.00000001` CPU usage twice – which could be both `almostEqual` and `notEqual` to 0, providing a narrow path along which the test might pass even if the underlying functionality was broken.

It's a holdover from an older version of the test, though, since I now change `my_procfs` and ensure I can read that back, so I can remove it if you would like.

test/_linux.py
sethp-jive: Here I'm testing that the cpu_percent is returning a reasonable default when it was unable to load /proc/stat at import-time.

I suppose it should be safe to check that it's specifically the floating point number 0, but as there is arithmetic going on under the hood I'm not sure we want the test to fail if the immediate post-import call happens to return a very small non-zero number such as `0.00000001`.

If you want to assert that the test always returns precisely 0.0f instead, I'm happy to make that change.

giampaolo: If it doesn't fail with 0 I would use `assertEqual`, yes, same for the comment above. Thanks.

sethp-jive:
@giampaolo I think I've addressed all of your feedback, and the appveyor failure seems unrelated to this commit. Would you agree?

giampaolo:
Looks good. Merged. Thanks.

PR #732, opened 2016-01-19
Revive Process.environ

fbenkstein:
Hi,

here's another shot at implementing Process.environ.  This has been running productively at my employer for a couple of months which shook out a few bugs.  We're using this to track subprocess trees and terminate a whole hierarchy by marking them with an environment variable.  This is similar to Jenkins' process tree killer.  Please have a look.

Best regards,
Frank.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -744,7 +744,7 @@ class Process(object):
             else:
                 self._proc.cpu_affinity_set(list(set(cpus)))
 
-    # Linux only
+    # Linux and OSX only
     if hasattr(_psplatform.Process, "environ"):
         def environ(self):
             """The environment variables of the process as a dict.  Note: this
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -744,6 +744,13 @@ class Process(object):
             else:
                 self._proc.cpu_affinity_set(list(set(cpus)))
 
+    # Linux only
+    if hasattr(_psplatform.Process, "environ"):
+        def environ(self):
+            """The environment variables of the process as a dict.  Note: this
+            might not reflect changes made after the process started.  """
+            return self._proc.environ()
+
     if _WINDOWS:
 
         def num_handles(self):
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -169,6 +169,32 @@ def supports_ipv6():
         return False
 
 
+def parse_environ_block(data):
+    """Parse a C environ block of environment variables into a dictionary."""
+    # The block is usually raw data from the target process.  It might contain
+    # trailing garbage and lines that do not look like assignments.
+    ret = {}
+    pos = 0
+
+    while True:
+        next_pos = data.find("\0", pos)
+        # nul byte at the beginning or double nul byte means finish
+        if next_pos <= pos:
+            break
+        # there might not be an equals sign
+        equal_pos = data.find("=", pos, next_pos)
+        if equal_pos > pos:
+            key = data[pos:equal_pos]
+            value = data[equal_pos+1:next_pos]
+            # Windows expects environment variables to be uppercase only
+            if os.name == "nt":
+                key = key.upper()
+            ret[key] = value
+        pos = next_pos + 1
+
+    return ret
+
+
 def sockfam_to_enum(num):
     """Convert a numeric socket family value to an IntEnum member.
     If it's not a known member, return the numeric value itself.
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -25,6 +25,7 @@ from . import _psutil_linux as cext
 from . import _psutil_posix as cext_posix
 from ._common import isfile_strict
 from ._common import memoize
+from ._common import parse_environ_block
 from ._common import NIC_DUPLEX_FULL
 from ._common import NIC_DUPLEX_HALF
 from ._common import NIC_DUPLEX_UNKNOWN
@@ -885,6 +886,12 @@ class Process(object):
         if data.endswith('\x00'):
             data = data[:-1]
         return [x for x in data.split('\x00')]
+
+    @wrap_exceptions
+    def environ(self):
+        with open_text("%s/%s/environ" % (self._procfs_path, self.pid)) as f:
+            data = f.read()
+        return parse_environ_block(data)
 
     @wrap_exceptions
     def terminal(self):
--- a/psutil/arch/osx/process_info.c
+++ b/psutil/arch/osx/process_info.c
@@ -238,6 +238,124 @@ error:
 }
 
 
+// return process environment as a python string
+PyObject *
+psutil_get_environ(long pid) {
+    int mib[3];
+    int nargs;
+    char *procargs = NULL;
+    char *procenv = NULL;
+    char *arg_ptr;
+    char *arg_end;
+    char *env_start;
+    size_t argmax;
+    PyObject *py_ret = NULL;
+
+    // special case for PID 0 (kernel_task) where cmdline cannot be fetched
+    if (pid == 0)
+        goto empty;
+
+    // read argmax and allocate memory for argument space.
+    argmax = psutil_get_argmax();
+    if (! argmax) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    procargs = (char *)malloc(argmax);
+    if (NULL == procargs) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    // read argument space
+    mib[0] = CTL_KERN;
+    mib[1] = KERN_PROCARGS2;
+    mib[2] = pid;
+    if (sysctl(mib, 3, procargs, &argmax, NULL, 0) < 0) {
+        if (EINVAL == errno) {
+            // EINVAL == access denied OR nonexistent PID
+            if (psutil_pid_exists(pid))
+                AccessDenied();
+            else
+                NoSuchProcess();
+        }
+        goto error;
+    }
+
+    arg_end = &procargs[argmax];
+    // copy the number of arguments to nargs
+    memcpy(&nargs, procargs, sizeof(nargs));
+
+    // skip executable path
+    arg_ptr = procargs + sizeof(nargs);
+    arg_ptr = memchr(arg_ptr, '\0', arg_end - arg_ptr);
+
+    if (arg_ptr == NULL || arg_ptr == arg_end)
+        goto empty;
+
+    // skip ahead to the first argument
+    for (; arg_ptr < arg_end; arg_ptr++) {
+        if (*arg_ptr != '\0')
+            break;
+    }
+
+    // iterate through arguments
+    while (arg_ptr < arg_end && nargs > 0) {
+        if (*arg_ptr++ == '\0')
+            nargs--;
+    }
+
+    // build an environment variable block
+    env_start = arg_ptr;
+
+    procenv = calloc(1, arg_end - arg_ptr);
+
+    if (procenv == NULL) {
+        PyErr_NoMemory();
+        goto error;
+    }
+
+    while (*arg_ptr != '\0' && arg_ptr < arg_end) {
+        char *s = memchr(arg_ptr + 1, '\0', arg_end - arg_ptr);
+
+        if (s == NULL)
+            break;
+
+        memcpy(procenv + (arg_ptr - env_start), arg_ptr, s - arg_ptr);
+
+        arg_ptr = s + 1;
+    }
+
+#if PY_MAJOR_VERSION >= 3
+    py_ret = PyUnicode_FromStringAndSize(procenv, arg_ptr - env_start + 1);
+#else
+    py_ret = PyString_FromStringAndSize(procenv, arg_ptr - env_start + 1);
+#endif
+
+    if (!py_ret)
+        goto error;
+
+    free(procargs);
+    free(procenv);
+
+    return py_ret;
+
+empty:
+    if (procargs != NULL)
+        free(procargs);
+    return Py_BuildValue("s", "");
+
+error:
+    Py_XDECREF(py_ret);
+    if (procargs != NULL)
+        free(procargs);
+    if (procenv != NULL)
+        free(procargs);
+    return NULL;
+}
+
+
 int
 psutil_get_kinfo_proc(pid_t pid, struct kinfo_proc *kp) {
     int mib[4];
--- a/test/test_psutil.py
+++ b/test/test_psutil.py
@@ -507,7 +507,7 @@ def skip_on_not_implemented(only_if=None):
     return decorator
 
 
-def create_temp_executable_file(suffix):
+def create_temp_executable_file(suffix, code="void main() { pause(); }"):
     tmpdir = None
     if TRAVIS and OSX:
         tmpdir = "/private/tmp"
@@ -520,7 +520,7 @@ def create_temp_executable_file(suffix):
             prefix='psu', suffix='.c', dir=tmpdir)
         os.close(fd)
         with open(c_file, "w") as f:
-            f.write("void main() { pause(); }")
+            f.write(code)
         subprocess.check_call(["gcc", c_file, "-o", path])
         safe_remove(c_file)
     else:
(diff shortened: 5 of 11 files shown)


fbenkstein:
This looks to be ready for review.

psutil/_common.py
```diff
@@ -169,6 +171,39 @@ def supports_ipv6():
...
+            break
+        equal_pos = data.find(equals, pos, next_pos)
+        if equal_pos > pos:
+            key = convert_string(data[pos:equal_pos])
+            value = convert_string(data[equal_pos+1:next_pos])
+            if uppercase_key:
+                key = key.upper()
+            ret[key] = value
+        pos = next_pos + 1
+    return ret
```
giampaolo: I don't understand why this function is so complicated. You just need to split for '\0' and then for '='. This is an example: https://code.google.com/p/psutil/source/detail?r=3e6905e0c623#

psutil/_pslinux.py
```diff
@@ -883,6 +884,15 @@ def cmdline(self):
         return [x for x in data.split('\x00')]
 
     @wrap_exceptions
+    def environ(self):
+        f = open_binary("/proc/%s/environ" % self.pid)
+        try:
+            data = f.read()
+        finally:
+            f.close()
```
giampaolo: You can use the `with` statement.

psutil/_pslinux.py
```diff
@@ -883,6 +884,15 @@ def cmdline(self):
         return [x for x in data.split('\x00')]
 
     @wrap_exceptions
+    def environ(self):
+        f = open_binary("/proc/%s/environ" % self.pid)
```
giampaolo: Mmm... I think you should use `open_text` instead.

psutil/_psutil_windows.c
```diff
@@ -2863,7 +2727,7 @@ psutil_net_if_addrs(PyObject *self, PyObject *args) {
 #if PY_MAJOR_VERSION >= 3
             py_mac_address = PyUnicode_FromString(buff);
 #else
-            py_mac_address = PyString_FromString(buff);
+            py_mac_address = PyBytes_FromString(buff);
```
giampaolo: why?

psutil/_psutil_windows.c
```diff
@@ -2917,7 +2781,7 @@ psutil_net_if_addrs(PyObject *self, PyObject *args) {
 #if PY_MAJOR_VERSION >= 3
                 py_address = PyUnicode_FromString(buff);
 #else
-                py_address = PyString_FromString(buff);
+                py_address = PyBytes_FromString(buff);
```
giampaolo: why?

psutil/arch/windows/process_info.c
```diff
@@ -204,118 +190,346 @@ handlep_is_running(HANDLE hProcess) {
...
+            goto error;
+
+        free(envData);
+        envData = NULL;
+    }
+
+    if (hProcess != NULL)
+        CloseHandle(hProcess);
+
+    return ret;
```
giampaolo: This is a huge refactoring involving 2 other "big" and complex functions (cmdline and cwd). This is too hard to review. Please provide a *new* isolated `psutil_proc_environ` function instead. The refactoring for merging the 3 functions (cwd, cmdline, environ) into a single one can be evaluated and made in a separate PR.

test/test_psutil.py
```diff
@@ -2475,6 +2475,25 @@ def test_Popen(self):
             proc.wait()
             self.assertIsNotNone(proc.returncode)
 
+    @unittest.skipUnless(hasattr(psutil.Process, "environ"),
+                         "environ not available")
+    def test_environ(self):
+        self.maxDiff = None
```
giampaolo: ?

psutil/arch/windows/process_info.c
giampaolo: ...also, as a side note, the C implementation should not return a string which is then passed to `parse_environ_block(data)`. Instead it should simply return a dict. Same goes for OSX implementation.

psutil/arch/osx/process_info.c
```diff
@@ -222,12 +234,43 @@ psutil_get_cmdline(long pid) {
...
+        goto error;
+
+    py_ret = Py_BuildValue("(N, N)", py_retlist, py_retenv);
+
+    if (!py_ret)
+        goto error;
+
+   free(procargs);
+
+   return py_ret;
```
giampaolo: Same as for Windows, I would like this to be defined as a separate `psutil_proc_environ` function instead of merging it with cmdline (the merge / refactoring can be done in a separate PR). Also calculating cmdline + environ together is slower thant cmdline alone or environ alone.

psutil/_pslinux.py
fbenkstein: But there could very well be non-utf-8 data in the environment.

psutil/_psutil_windows.c
fbenkstein: This was originally a separate change.  PyString_AS_STRING vanished from Python 3.4(?) and I thought it would make sense to use PyBytes everywhere instead.  I can revert this.

psutil/arch/windows/process_info.c
fbenkstein: Okay, but that would mean doing the parsing in C.  It's doable but much more error-prone IMHO.

giampaolo:
This is cool, thanks for working on this. FYI in case you don't know there's some previous work that had been done here: https://code.google.com/p/psutil/issues/detail?id=52
Also, some confusing code regarding Solaris:
https://code.google.com/p/psutil/issues/detail?id=247

One question for you: in here (https://code.google.com/p/psutil/issues/detail?id=52#c18) I stated that on Windows you can get environ only for the current process (or else AccessDenied exception). Does this patch fix that issue or not?

test/test_psutil.py
fbenkstein: OSX has weird variables that might be missing in either Process.environ or os.environ.  I could not reproduce this on my local system but the context was truncated.  maxDiff = None makes the errors much easier to spot.

psutil/arch/osx/process_info.c
fbenkstein: Okay.

psutil/arch/windows/process_info.c
giampaolo: Ah I see. Then feel free to do the parsing in Python instead.

test/test_psutil.py
giampaolo: https://docs.python.org/2/library/unittest.html#unittest.TestCase.maxDiff 
interesting; didn't know about this

fbenkstein:
I saw the earlier tries and used it early in my implementation.  Because of the problems on Windows I took the code from the blog post mentioned in the commit message instead.  I can confirm that the Windows code works not just for the current process.  I think there was a wrong offset or address in the earlier implementation when calling ReadProcessMemory.

Should I put this into separate PRs so it can be reviewed separately?

psutil/_pslinux.py
giampaolo: Mmmm, I suppose that is true, but in that case also your `convert_string()` function should fail.
I think what we need is this:

```
        kw = dict(errors='replace') if PY3 else dict()
        with open_text("%s/%s/environ" % (get_procfs_path(), self.pid), **kw) as f:
             ...
```

psutil/_psutil_windows.c
giampaolo: Yes revert please. Thanks.

giampaolo:
> I can confirm that the Windows code works not just for the current process. I think there was 
> a wrong offset or address in the earlier implementation when calling ReadProcessMemory.

Great! This was one of the main reasons which led me to give up on this.

> Should I put this into separate PRs so it can be reviewed separately?

I don't think it's necessary.

psutil/_common.py
fbenkstein: I'll try to remove the difference between strings and bytes.  However, only splits do not work because there are two cases that will be handled improperly.  Firstly, when reading from a process' memory we might read more data than necessary.  In that case the double nul byte signals a stop.  Secondly, it's possible that an item does not contain an equals sign.  I've handled both these cases in the tests but I'll add a comment in the code  as well.

giampaolo: On what platforms you need to handle those cases? Windows only? It seems to me that on Linux, for instance, the logic is pretty straightforward as in https://code.google.com/p/psutil/source/detail?r=3e6905e0c623#. Perhaps you should have each platform define its own parser?

fbenkstein: I do not think that's restricted to any particular platform.  When reading from the process memory on OSX and Windows you might get more than the environment block.  And on POSIX you can call putenv with an argument that does not contain an equals sighn.  I'm actually preparing a new version right now that will test these cases with a subprocess.

fbenkstein:
I'm still not sure how to handle non-utf8 data consistently between Python 2 and 3.  Should it be byte strings in both cases?  I don't have much experience with Python 3 but it seems that users there seem to expect str (i.e. unicode) and the rest of the psutil seems to be modelled that way.  Unfortunately that means that process names, command lines and now environments that contain non-utf8 data will break applications.

Example:
    cp /bin/cat $'\xff\xfe'
    pid=$!
    python2.7 -c "import psutil; print(psutil.Process($pid).name())"
    ▒▒
    python3.5 -c "import psutil; print(psutil.Process($pid).name())"
  File "<string>", line 1, in <module>
  File "/home/xxx/Projects/psutil/psutil/__init__.py", line 543, in name
    name = self._proc.name()
  File "/home/xxx/Projects/psutil/psutil/_pslinux.py", line 811, in wrapper
  File "/home/xxx/Projects/psutil/psutil/_pslinux.py", line 855, in name
    data = f.read()
  File "/usr/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 7: invalid start byte

fbenkstein:
I've side stepped the encoding issue for now.  Also, I think it's better if I move the Windows code into a separate PR if that's okay with you.

giampaolo:
Sure, make another PR if you want.

psutil/_pslinux.py
```diff
@@ -883,6 +884,12 @@ def cmdline(self):
         return [x for x in data.split('\x00')]
 
     @wrap_exceptions
+    def environ(self):
+        with open_text("/proc/%s/environ" % self.pid) as f:
```
giampaolo: This has to be `with open_text("%s/%s/environ" % (get_profs_path(), self.pid)) as f:`

test/test_psutil.py
```diff
@@ -501,7 +501,7 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
-def create_temp_executable_file(suffix):
+def create_temp_executable_file(suffix, code="void main() { pause(); }"):
```
giampaolo: nice

test/test_psutil.py
```diff
@@ -2475,6 +2475,45 @@ def test_Popen(self):
...
+    @unittest.skipUnless(POSIX, "posix only")
+    def test_weird_environ(self):
+        path = create_temp_executable_file("x", code="""
+#include <unistd.h>
+char * const argv[] = {"cat", 0};
+char * const envp[] = {"A=1", "X", "C=3", 0};
+int main(void) {
+    return execve("/bin/cat", argv, envp);
+}
+""")
```
giampaolo: consider using `textwrap.dedent` (test_psutil.py has examples)

test/test_psutil.py
```diff
@@ -2475,6 +2475,55 @@ def test_Popen(self):
...
+            ):
+                d.pop(key, None)
+                d2.pop(key, None)
+
+        self.assertEqual(d, d2)
+
+    @unittest.skipUnless(hasattr(psutil.Process, "environ"),
+                         "environ not available")
+    @unittest.skipUnless(POSIX, "posix only")
+    def test_weird_environ(self):
```
giampaolo: perhaps add a comment explaining what you're testing exactly (aka why it is "weird"?).

fbenkstein:
Okay.  I'm done for today.  Will try to prepare the Windows change tomorrow.

test/test_psutil.py
```diff
@@ -2475,6 +2475,60 @@ def test_Popen(self):
...
+        # environment variables can contain values without an equals sign
+        # TODO: arbitrary binary data should also be tested but currently
+        # breaks on Python 3
```
giampaolo: What do you mean by that? Does it break because of [en/de]coding errors? Only on Linux? On Linux have you tried to do this?

```
        kw = dict(errors='replace') if PY3 else dict()
        with open_text("/path...", **kw) as f:
              ...
```

test/test_psutil.py
```diff
@@ -2475,6 +2475,60 @@ def test_Popen(self):
...
+                                    stderr=subprocess.PIPE)
+        p = psutil.Process(sproc.pid)
+        wait_for_pid(p.pid)
+        self.assertTrue(p.is_running())
+        # Wait for process to exec or exit.
+        self.assertEquals(sproc.stderr.read(), "")
+        self.assertEquals(p.name(), "cat")
+        self.assertEquals(p.environ(), {"A": "1", "C": "3"})
+        sproc.communicate()
+        self.assertEquals(sproc.returncode, 0)
```
giampaolo: Replace all `assertEquals` with `assertEqual` (the former is deprecated)

test/test_psutil.py
```diff
@@ -2475,6 +2475,60 @@ def test_Popen(self):
...
+        self.addCleanup(safe_remove, path)
+        sproc = get_test_subprocess([path],
+                                    stdin=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+        p = psutil.Process(sproc.pid)
+        wait_for_pid(p.pid)
+        self.assertTrue(p.is_running())
+        # Wait for process to exec or exit.
+        self.assertEquals(sproc.stderr.read(), "")
+        self.assertEquals(p.name(), "cat")
```
giampaolo: I think this is not reliable across all POSIX platforms (name can be different) - I would remove it.

test/test_psutil.py
fbenkstein: Please see my other comment here:  https://github.com/giampaolo/psutil/pull/732#issuecomment-173223649

The replace-argument is possible but I'd rather do this in a way works across different platforms.  That was the original reason for the ugliness in parse_environment_block.

giampaolo: Ah I missed that sorry. That looks like a (Linux) bug. Can you please file a new ticket for https://github.com/giampaolo/psutil/pull/732#issuecomment-173223649?

fbenkstein:
The AppVeyor test failures seem unrelated.  Is there anything else missing here?

psutil/__init__.py
```diff
@@ -744,6 +744,15 @@ def cpu_affinity(self, cpus=None):
             else:
                 self._proc.cpu_affinity_set(list(set(cpus)))
 
+    # Linux and OSX only
+    if hasattr(_psplatform.Process, "environ"):
+        def environ(self):
+            """The process environment variables as a dict.
+            Note: this usually reflects the state when the process was started.
+            Later changes done by the process might not show up here.
+            """
```
giampaolo: If it is true that on all platforms environ updates are not reflected then I would just make that clear.

psutil/arch/osx/process_info.c
```diff
@@ -234,6 +234,122 @@ psutil_get_cmdline(long pid) {
...
+            nargs--;
+    }
+
+    // build an environment variable block
+    env_start = arg_ptr;
+
+    procenv = calloc(1, arg_end - arg_ptr);
+
+    if (procenv == NULL)
+        goto error;
```
giampaolo: I think you are supposed to `PyErr_NoMemory()` before `goto error`

test/test_psutil.py
```diff
@@ -2974,6 +3028,27 @@ def foo(*args, **kwargs):
...
+        self.assertEqual(parse_environ_block("a=1\0b=\0\0"),
+                         {k("a"): "1", k("b"): ""})
+        # ignore everything after \0\0
+        self.assertEqual(parse_environ_block("a=1\0b=2\0\0c=3\0"),
+                         {k("a"): "1", k("b"): "2"})
+        # ignore everything that is not an assignment
+        self.assertEqual(parse_environ_block("xxx\0a=1\0"), {k("a"): "1"})
+        self.assertEqual(parse_environ_block("a=1\0=b=2\0"), {k("a"): "1"})
+        # do not fail if the block is incomplete
+        self.assertEqual(parse_environ_block("a=1\0b=2"), {k("a"): "1"})
```
giampaolo: nice set of test cases!

giampaolo:
I added a couple more comments, the rest is good.

psutil/arch/osx/process_info.c
fbenkstein: Nice.  Thanks.

psutil/__init__.py
fbenkstein: I don't think it's that easy.  The memory for the environment variables is owned by the process which can do whatever it wants with it as far as I know.  I looked at a couple of setenv implementations and Linux's libcs seem to leave the original memory in place.  It seems that BSD might reuse the memory if you're overwriting a value with a smaller one.  Finally, you can also change the memory returned by getenv.  I only added this note because one of your comment here https://code.google.com/p/psutil/issues/detail?id=52#c18

giampaolo:
OSX failure https://travis-ci.org/giampaolo/psutil/jobs/104056877:

```
'/private/var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/tmpnGpanB/\xc0\x80'
```
vs:
```
             '/var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/tmpnGpanB/\xc0\x80'
```
I currently don't have access to OSX but I seem to remember this happens because /private is a symlink so perhaps you should be able to fix this by using `os.path.realpath` which resolves symlinks.

giampaolo:
Resources for getting environ on Solaris:
- https://github.com/hyperic/sigar/blob/07b221a25345367f55fbfba6d3113c0e04068846/src/os/solaris/solaris_sigar.c#L1013
- http://code.metager.de/source/xref/netbeans/dlight.nativeexecution/tools/killall/src/solaris/pfind.c#25
- `pargs -e {pid}`
- pargs source code: https://github.com/illumos/illumos-gate/blob/139510fb6efa97dbe5f5479594b308d940cab8d1/usr/src/cmd/ptools/pargs/pargs.c

giampaolo:
@fbenkstein do yu have solaris and want to give that a try?

fbenkstein:
I have a Solaris VM which calls itself Oracle Solaris 11.3 (SunOS 5.11 11.3).  I'll see what I can do but Windows support is a higher priority for me at the moment.  After that is done I might have to switch to doing something else for a week or two.

giampaolo:
Clear, thanks for the update.

PR #733, opened 2016-01-21
add non-unicode testcase

fbenkstein:
This adds a test case for non-unicode filenames.  It passes with Python 2.7 on the platforms that I have access to (Windows, Linux, OSX).  Only on Windows does the Python 3 test pass as well and I see no easy way to make them pass on Linux and OSX.

### DIFF
--- a/test/test_psutil.py
+++ b/test/test_psutil.py
@@ -20,6 +20,7 @@ import atexit
 import collections
 import contextlib
 import datetime
+import distutils.spawn
 import errno
 import functools
 import json
@@ -153,8 +154,7 @@ atexit.register(lambda: DEVNULL.close())
 _subprocesses_started = set()
 
 
-def get_test_subprocess(cmd=None, stdout=DEVNULL, stderr=DEVNULL,
-                        stdin=DEVNULL, wait=False):
+def get_test_subprocess(cmd=None, wait=False, **kwds):
     """Return a subprocess.Popen object to use in tests.
     By default stdout and stderr are redirected to /dev/null and the
     python interpreter is used as test process.
@@ -169,7 +169,10 @@ def get_test_subprocess(cmd=None, stdout=DEVNULL, stderr=DEVNULL,
         cmd_ = [PYTHON, "-c", pyline]
     else:
         cmd_ = cmd
-    sproc = subprocess.Popen(cmd_, stdout=stdout, stderr=stderr, stdin=stdin)
+    kwds.setdefault("stdin", DEVNULL)
+    kwds.setdefault("stdout", DEVNULL)
+    kwds.setdefault("stderr", DEVNULL)
+    sproc = subprocess.Popen(cmd_, **kwds)
     if wait:
         if cmd is None:
             stop_at = time.time() + 3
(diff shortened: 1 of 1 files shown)


test/test_psutil.py
```diff
@@ -3231,6 +3252,116 @@ def test_proc_open_files(self):
...
+class TestNonUnicode(unittest.TestCase):
+    "Test handling of non-utf8 data."
+
+    @classmethod
+    def setUpClass(cls):
+        cls.temp_directory = tempfile.mkdtemp(suffix=b"")
+
+        # Return an executable that runs until we close its stdin
+        if WINDOWS:
+            cls.test_executable = distutils.spawn.find_executable("cmd.exe")
```
giampaolo: I'd rather use `psutil._compat.which` instead.

test/test_psutil.py
```diff
@@ -3231,6 +3252,116 @@ def test_proc_open_files(self):
...
+        else:
+            assert POSIX
+            cls.test_executable = "/bin/cat"
```
giampaolo: Same here; `which('cat')` is more portable/safe.

test/test_psutil.py
```diff
@@ -3231,6 +3252,116 @@ def test_proc_open_files(self):
         self.assertEqual(os.path.normcase(path), os.path.normcase(self.uexe))
 
 
+class TestNonUnicode(unittest.TestCase):
+    "Test handling of non-utf8 data."
```
giampaolo: triple quotes

test/test_psutil.py
```diff
@@ -3231,6 +3252,116 @@ def test_proc_open_files(self):
...
+        self.addCleanup(safe_remove, test_script)
+        subp = get_test_subprocess(cmd=[PYTHON, decode_path(test_script)],
+                                   stdin=subprocess.PIPE,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.STDOUT)
+        self.addCleanup(safe_remove, funny_file)
+        p = psutil.Process(subp.pid)
+        # wait for the file to appear
+        while len(os.listdir(self.temp_directory)) == 1:
+            time.sleep(0.01)
```
giampaolo: You can use `call_until(os.listdir(self.temp_directory), "len(ret) == 1")` instead. The advantage is that it will timeout after a while in case the directory doesn't change.

test/test_psutil.py
fbenkstein: Actually I think cat is always at /bin/cat on POSIX.

test/test_psutil.py
fbenkstein: Okay.

fbenkstein:
This is just meant as a starting point as the test cases are failing on Python 3 at the moment.  I'll post an update but probably won't fix the underlying issue.

giampaolo:
I fixed the underlying problem causing the encode/decode exceptions (on Linux at least) and exposed a new ENCODING_ERRORS_HANDLER constants which can be used to control this behavior:  https://github.com/giampaolo/psutil/commit/c62b41f3df3e3ee98cc605a64b90152d3a9c631e
I'm happy with your test cases: they test an important use case and also gave me the idea for the new constant.

PR #738, opened 2016-01-25
Refactor win cwd cmdline

fbenkstein:
As discussed this creates a common function for Process.cwd and Process.cwdline on Windows.

PR #739, opened 2016-01-25
Windows bitness confusion

fbenkstein:
This is a greater refactoring of the calls to ReadProcessMemory.  Currently all the catch-all error handlers mapping ERROR_PARTIAL_READ to AccessDenied are gone.

This builds on top of https://github.com/giampaolo/psutil/pull/738.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,9 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
   UnicodeDecodeError exceptions. 'surrogateescape' error handler is now
   used as a workaround for replacing the corrupted data.
 - #741: [OpenBSD] fix compilation on mips64.
+- #737: [Windows] when the bitness of psutil and the target process was
+  different cmdline() and cwd() could return a wrong result or incorrectly
+  report an AccessDenied error.
 
 
 3.4.2 - 2016-01-20
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -575,7 +575,6 @@ static PyObject *
 psutil_proc_cmdline(PyObject *self, PyObject *args) {
     long pid;
     int pid_return;
-    PyObject *py_retlist;
 
     if (! PyArg_ParseTuple(args, "l", &pid))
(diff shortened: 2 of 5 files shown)


giampaolo:
What happens now instead of `AccessDenied`? Can you describe what changed in HISTORY.rst?

fbenkstein:
> What happens now instead of AccessDenied?

I haven't seen an ERROR_PARTIAL_READ so far.  My hunch is that most of them were caused by mixing bitness.  Maybe if they get rarer it's worth letting them trickle up the stack.  They might happen again with my environ change though because that reads another even bigger chunk.  If that turns out to be an issue I'll handle them as before (either ignore as in cmdline() or convert to AccessDenied).

> Can you describe what changed in HISTORY.rst?

Will do.

fbenkstein:
I'm still not sure about throwing NotImplementedError.  It breaks the 32 bit test cases and it's probably unexpected.  I'll see if I can actually fix the case where python is 32 bit and the target process is 64 bit.

fbenkstein:
Okay, all the cases are implemented and seem to be working now.  Unfortunately the change is quite large now.  The different paths are similar but not quite the same I do not have a good idea on how to unify them yet.

giampaolo:
Can you fix conflicts (`git merge master` etc.)?

fbenkstein:
Rebased.

giampaolo:
Please fix lines lenght so that they are no longer than 80 chars then it's good to go.

fbenkstein:
Okay.  I changed the long lines.  What do you think about adding an auto-formatting target (clang-format or uncrustify or some such) for the C code?

giampaolo:
I am not familiar with that. How would it work?

fbenkstein:
> I am not familiar with that. How would it work?

There could be a target that automatically rewrites the C files according to some configuration file.  I'll upload a proposal.

giampaolo:
Thanks. I ll get back home and merge the pr (or you do if you have
permissions)

giampaolo:
Thanks

PR #740, opened 2016-01-26
fix net_if_stats on Windows wip

fbenkstein:
Saw this compiler warning popping up and thought I could take a look:
```
psutil/_psutil_windows.c(2081) : warning C4133: 'function' : incompatible types - from 'PyObject *' to 'const char *'
```


giampaolo:
Nice one! (feel free to update CREDITS if you want)

PR #741, opened 2016-01-26
Get system page size via sysconf()

landryb:
Get system page size at runtime via sysconf(). This allows use of the
same binary on systems with different page sizes.


giampaolo:
Please also update HISTORY.rst and CREDITS.rst

giampaolo:
Thanks

PR #743, opened 2016-01-27
add code formatter for c code

fbenkstein:
This adds an uncrustify target which will rewrite all C files.  I tried apply PEP 7 rules and keep the rest mostly as it is.  Might need some tweaking.

### DIFF
--- a/.uncrustify.cfg
+++ b/.uncrustify.cfg
@@ -1,0 +1,1434 @@
+# Uncrustify 0.59
+
+#
+# General options
+#
+
+# The type of line endings
+newlines                                 = auto     # auto/lf/crlf/cr
+
+# The original size of tabs in the input
+input_tab_size                           = 8        # number
+
+# The size of tabs in the output (only used if align_with_tabs=true)
+output_tab_size                          = 8        # number
+
+# The ASCII value of the string escape char, usually 92 (\) or 94 (^). (Pawn)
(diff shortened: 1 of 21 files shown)


giampaolo:
Thanks for looking into this. There are some style changes I don't agree with but I like this idea. Let me try to experiment with this a little (probably tomorrow).
Also, it would be good to integrate this with the already existent GIT commit hook which now checks edited *.py files with flake8. The same thing should be done for the edited C file and reject the commit if they are not compliant. I wonder whether it exists an `.uncrustify.cfg` for PEP7. I not finding any... which is kinda weird.

giampaolo:
I'm trying to play a bit with uncrustify. What I would like to do is to apply one fix at a time, for instance this:

```
# Try to limit code width to N number of columns
code_width                               = 79        # number
```

...but apparently there's a set of fixes which are applied implicitly by default and I can't disable them.

fbenkstein:
Hm.  I think it should be possible to set almost all fixers to "ignore".

giampaolo:
Kept experimenting. It just seems to hard to configure this the way I want to. I think it's better if we post-pone this for a later time; perhaps by finding a linter which is easier to configure and maintain.

fbenkstein:
Seems reasonable.  I also looked at clang-format but it doesn't look stable
enough to be usable.

Issue #289, opened 2014-05-23
Enhance mapped memory regions used by process to get dylib path out of dyld shared cache file


giampaolo:
_From [jcscoob...@gmail.com](https://code.google.com/u/103960149060463949301/) on June 26, 2012 01:04:00_

```
OS X support for mapped memory regions used by process added in Issue 260 is in 
need of a final enhancement that will parse the actual dylib filename/path out 
of the dyld shared cache map file.  Right now, any dylib that is loaded from 
the OS X dyld shared cache reports the path to the dyld shared cache file 
instead of the actual dylib that was requested to be loaded.  To do this, based 
on conversations with experienced mach kernel developers, we will need to parse 
the dyld_shared_cache_*.map file to find the real dylib path.
```

_Original issue: http://code.google.com/p/psutil/issues/detail?id=289_

PR #745, opened 2016-01-27
Measure USS on OSX

EricRahm:
This adds a USS (unique set size) measurement to `memory_info_ex` on OSX. It is based on the implementation in Firefox's [memory measurement subsystem](https://dxr.mozilla.org/mozilla-central/rev/aa90f482e16db77cdb7dea84564ea1cbd8f7f6b3/xpcom/base/nsMemoryReporterManager.cpp#472-562).


giampaolo:
I would be interested in seeing the impact on performances. Could you please measure it before and after the patch and paste the result?

```
python -m timeit -s "import psutil; p = psutil.Process()" "p.memory_info_ex()"
```

...also, while we're at it, is it possible to provide also PSS stats?

EricRahm:
before (master):
```
python -m timeit -s "import psutil; p = psutil.Process()" "p.memory_info_ex()"
100000 loops, best of 3: 2.17 usec per loop
```
after:
```
python -m timeit -s "import psutil; p = psutil.Process()" "p.memory_info_ex()"
10000 loops, best of 3: 113 usec per loop
```

So slower, but not as slow as on Linux.

giampaolo:
Any idea how to determine PSS?

EricRahm:
I do not know of a method for calculating PSS on OSX.

giampaolo:
Sorry for replying late. I was looking back at this and I wonder how this relates to `memory_maps` implementation. Currently `memory_maps` implementation on OSX provides a "private" field. As such we should already be able to determine USS by doing:

```
>>> sum([x.private for x in psutil.Process().memory_maps()])
```

...but on the other hand they way you are determining USS (I'm talking about the implementation) is different. What's the difference between `sum([x.private for x in psutil.Process().memory_maps()])` and `psutil.Process().memory_info_ex().uss`? Are they equal?

psutil/_psutil_osx_uss.c
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  vm_size_t pageSize;
+  if (host_page_size(mach_host_self(), &pageSize) != KERN_SUCCESS) {
+    pageSize = PAGE_SIZE;
+  }
+
+  *aN = privatePages * pageSize;
+  return true;
+}
```
giampaolo: - indentation should be 4 spaces
- I don't like this being in a separate file, you can use `psutil/arch/osx/process_info.c
- use `//` for all comments (instead of `/*`)

psutil/_psutil_osx_uss.c
```diff
@@ -0,0 +1,105 @@
...
+#include <mach/mach_init.h>
+#include <mach/mach_vm.h>
+#include <mach/shared_region.h>
+#include <mach/task.h>
+#include <sys/sysctl.h>
+
+#include "_psutil_osx_uss.h"
+
+bool
+InSharedRegion(mach_vm_address_t aAddr, cpu_type_t aType)
```
giampaolo: please rename this to `psutil_in_shared_region`; also add a comment "docstring" explaining what this function does and do not use `camelCase` for variable name but `var_name` instead.

psutil/_psutil_osx_uss.c
```diff
@@ -0,0 +1,105 @@
...
+      break;
+    default:
+      return false;
+  }
+
+  return base <= aAddr && aAddr < (base + size);
+}
+
+bool
+calc_uss(mach_port_t target, int64_t* aN)
```
giampaolo: rename this to `psutil_proc_uss`; also a better (more meaningful) name instead of `aN` would be nice

psutil/_psutil_osx_uss.c
```diff
@@ -0,0 +1,105 @@
...
+{
+  if (!aN) {
+    return false;
+  }
+
+  cpu_type_t cpu_type;
+  size_t len = sizeof(cpu_type);
+  if (sysctlbyname("sysctl.proc_cputype", &cpu_type, &len, NULL, 0) != 0) {
+    return false;
```
giampaolo: Looks like you want to `PyErr_SetFromErrno(PyExc_OSError);` here. The calling function should then return NULL so that you'll get an exception at Python level.

psutil/_psutil_osx_uss.c
```diff
@@ -0,0 +1,105 @@
...
+    mach_port_t objectName;
+
+    kern_return_t kr =
+        mach_vm_region(target, &addr, &size, VM_REGION_TOP_INFO,
+                       (vm_region_info_t)&info,
+                       &infoCount, &objectName);
+    if (kr == KERN_INVALID_ADDRESS) {
+      /* Done iterating VM regions. */
+      break;
+    } else if (kr != KERN_SUCCESS) {
```
giampaolo: this should be indented as:

```
} 
else if (kr != KERN_SUCCESS) {
```

psutil/_psutil_osx_uss.c
giampaolo: On a second thought I would prefer you define this in psutil/_psutil_osx.c as proc_memory_uss.
From the python file (psutil/_psosx.py) after you call cext.psutil_proc_memory_info you will call `cext.proc_memory_uss` to get USS stats separately.

giampaolo:
I found this:
https://gitlab.phusion.nl/passenger/temp-passenger/blob/99bf9b590e2a5c3620d5fc34f540b749452448ca/ext/common/Utils/ProcessMetricsCollector.h#L381
...and here the same operation is referenced as "PSS" not "USS". Weird...

EricRahm:
> What's the difference between sum([x.private for x in psutil.Process().memory_maps()]) and psutil.Process().memory_info_ex().uss results? Are they equal?

They are not equal, `memory_maps` seems very low:
```
>>>print p.memory_info_ex(); print sum([x.private for x in psutil.Process().memory_maps()])
pextmem(rss=7880704L, vms=2534256640L, pfaults=9531392, pageins=1564672, uss=5849088L)
622592
```

The primary difference I can tell is `memory_maps` uses `vm_region_recurse_64` and skips all submaps.


> I found this:
> https://gitlab.phusion.nl/passenger/temp-passenger/blob/99bf9b590e2a5c3620d5fc34f540b749452448ca/ext/common/Utils/ProcessMetricsCollector.h#L381
> ...and here the same operation is referenced as "PSS" not "USS". Weird...

They seem to be calculating some sort of USS/PSS hybrid.

EricRahm:
@giampaolo Thanks for the review, I'll update per your style requests and merge into the main file.

EricRahm:
@giampaolo Updated per requests, I think I covered everything:
- 4 spaces, foo_bar naming, // for comments
- Moved impl into _psutil_osx.c, renamed `psutil_proc_memory_uss`, exported in cext
- Added function level docs
- Use `cext.proc_memory_uss` to add uss to `memory_info_ex` result

If all is good I'd recommend squashing before merging if that's not in your standard workflow.

giampaolo:
What do you mean by that?

EricRahm:
@giampaolo I pushed an update that covered your code review changes, just wanted to list out what I changed and make sure you noticed.

giampaolo:
Merged. Thanks a lot for this (I'll also update CREDITS).

giampaolo:
> The primary difference I can tell is memory_maps uses vm_region_recurse_64 and skips all submaps.

This made me think of https://github.com/giampaolo/psutil/issues/289.
If you understand the issue maybe you could fix `memory_maps` so that it takes submaps into account?

PR #746, opened 2016-01-28
Measure USS on Windows

EricRahm:
This adds a USS (unique set size, aka private working set) measurement to `memory_info_ex` on Windows.

Performance before patch:

```
$ python -m timeit -s "import psutil; p = psutil.Process()" "p.memory_info_ex()"
100000 loops, best of 3: 3.44 usec per loop
```

Performance after patch:

```
$ python -m timeit -s "import psutil; p = psutil.Process()" "p.memory_info_ex()"
10000 loops, best of 3: 44 usec per loop
```

So slower, but not as bad as Linux and OSX.


EricRahm:
This is based off the [memory reporting implementation in Firefox](https://dxr.mozilla.org/mozilla-central/rev/aa90f482e16db77cdb7dea84564ea1cbd8f7f6b3/xpcom/base/nsMemoryReporterManager.cpp#607-652).

EricRahm:
It would appear AppVeyor is unhappy about the usage of stdint.h on Windows, I'm guessing Visual C++ for Python 9 does not include it, but it was in my path due to having a full-fledged modern msvc install.

fbenkstein:
Python 2.7 still has to be compiled with Visual Studio 2008 which does not support stdint.h.

giampaolo:
Related: https://github.com/giampaolo/psutil/issues/752

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+  if (proc) {
+    CloseHandle(proc);
+  }
+
+  if (infoArray) {
+    free(infoArray);
+  }
+
+  return result;
+}
```
giampaolo: Indentation should be 4 spaces. Also, there's no need to put this in a separate file. You can use `psutil/arch/windows/process_info.c` instead.

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+/* vim: set ts=8 sts=2 et sw=2 tw=80: */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "_psutil_windows_uss.h"
+#include <psapi.h>
+
+BOOL
+calc_uss(DWORD target, unsigned long long* aN)
```
giampaolo: rename this to "psutil_proc_uss"; also use a better variable name (instead of `aN`)

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+  HANDLE proc;
+  PSAPI_WORKING_SET_INFORMATION tmp;
+  DWORD tmpSize = sizeof(tmp);
+  size_t entries, privatePages, i;
+  DWORD infoArraySize;
+  PSAPI_WORKING_SET_INFORMATION* infoArray;
+  SYSTEM_INFO si;
+
+  proc = OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ,
+                     FALSE, target);
```
giampaolo: You should use `psutil_handle_from_pid` defined in `psutil/arch/windows/process_info.c`; that will take care of raising `NoSuchProcess` and `AccessDenied` exception if needed.

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+
+  if (!proc) {
+    goto done;
+  }
+
+  /* Determine how many entries we need. */
+  memset(&tmp, 0, tmpSize);
+  QueryWorkingSet(proc, &tmp, tmpSize);
```
giampaolo: You should check the return value; if `0` use `PyErr_SetFromWindowsErr(0);`

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+
+  /* Fudge the size in case new entries are added between calls. */
```
giampaolo: Please use `//` comments instead

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+  entries = tmp.NumberOfEntries * 2;
+
+  if (!entries) {
+    goto done;
+  }
+
+  infoArraySize = tmpSize + (entries * sizeof(PSAPI_WORKING_SET_BLOCK));
+  infoArray = (PSAPI_WORKING_SET_INFORMATION*)malloc(infoArraySize);
```
giampaolo: You should check malloc return value; if `NULL` then do `PyErr_NoMemory();` and return `NULL`

psutil/_psutil_windows_uss.c
```diff
@@ -0,0 +1,73 @@
...
+  privatePages = 0;
+  for (i = 0; i < entries; i++) {
+    /* Count shared pages that only one process is using as private. */
+    if (!infoArray->WorkingSetInfo[i].Shared ||
+        infoArray->WorkingSetInfo[i].ShareCount <= 1) {
+      privatePages++;
+    }
+  }
+
+  GetSystemInfo(&si);
```
giampaolo: please add a comment saying that this function does not return any value; also I would rename `si` to `system_info` for clarity.

psutil/_psutil_windows_uss.c
giampaolo: On a second thought I would prefer you define this in `psutil/_psutil_windows.c` as `proc_memory_uss`.
From the python file (`psutil/_pswindows.py`) after you call `cext.psutil_proc_memory_info` (or `cext.psutil_proc_memory_info_2` you will call `cext.proc_memory_uss` to get USS stats separately.

EricRahm:
@giampaolo The latest pushes should cover your code review comments.

psutil/_psutil_windows.c
```diff
@@ -776,6 +776,87 @@ psutil_proc_memory_info_2(PyObject *self, PyObject *args) {
...
+{
+    DWORD pid;
+    HANDLE proc;
+    PSAPI_WORKING_SET_INFORMATION tmp;
+    DWORD tmp_size = sizeof(tmp);
+    size_t entries, private_pages, i;
+    DWORD info_array_size;
+    PSAPI_WORKING_SET_INFORMATION* info_array;
+    SYSTEM_INFO system_info;
+    PyObject* result = NULL;
```
giampaolo: please rename this to `py_result`

psutil/_psutil_windows.c
```diff
@@ -776,6 +776,87 @@ psutil_proc_memory_info_2(PyObject *self, PyObject *args) {
...
+ * Returns the USS of the process.
+ */
+static PyObject *
+psutil_proc_memory_uss(PyObject *self, PyObject *args)
+{
+    DWORD pid;
+    HANDLE proc;
+    PSAPI_WORKING_SET_INFORMATION tmp;
+    DWORD tmp_size = sizeof(tmp);
+    size_t entries, private_pages, i;
```
giampaolo: please only one variable declaration per line

psutil/_psutil_windows.c
```diff
@@ -776,6 +776,87 @@ psutil_proc_memory_info_2(PyObject *self, PyObject *args) {
...
+
+    proc = psutil_handle_from_pid(pid);
+    if (proc == NULL)
+        return NULL;
+
+    // Determine how many entries we need.
+    memset(&tmp, 0, tmp_size);
+    if (!QueryWorkingSet(proc, &tmp, tmp_size)) {
+        // NB: QueryWorkingSet is expected to fail here due to the
+        //     buffer being too small.
```
giampaolo: unnecessary spaces?

EricRahm:
@giampaolo ee6024f should address the final comments.

PR #747, opened 2016-01-28
Revive environ windows

fbenkstein:
I believe this could be ready now.  Getting the 32bit -> 64bit accesses was a pain but it appears to be working now.  Unfortunately, NtWow64QueryVirtualMemory64 appears to be missing from Windows 10.  So Process.environ() will throw NotImplementedError there when accessing a 64bit process from 32bit Python.

### DIFF
--- a/psutil/arch/windows/process_info.c
+++ b/psutil/arch/windows/process_info.c
@@ -308,15 +308,20 @@ typedef struct {
 } PEB64;
 #endif
 
-/* Get one or more parameters of the process with the given pid:
+enum psutil_process_data_kind {
+    KIND_CMDLINE,
+    KIND_CWD,
+};
 
-   pcmdline: the command line as a Python list
-   pcwd: the current working directory as a Python string
+/* Get data from the process with the given pid.  The data is returned in the
+   pdata output member as a nul terminated string which must be freed on
+   success.
 
-   On success 0 is returned.  On error the given output parameters are not
-   touched, -1 is returned, and an appropriate Python exception is set. */
-static int psutil_get_parameters(long pid, PyObject **pcmdline,
-                                 PyObject **pcwd) {
+   On success 0 is returned.  On error the output parameter is not touched, -1
+   is returned, and an appropriate Python exception is set. */
+static int psutil_get_process_data(long pid,
+                                   enum psutil_process_data_kind kind,
+                                   WCHAR **pdata) {
     /* This function is quite complex because there are several cases to be
        considered:
 
@@ -346,19 +351,17 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
     static _NtQueryInformationProcess NtWow64QueryInformationProcess64 = NULL;
     static _NtWow64ReadVirtualMemory64 NtWow64ReadVirtualMemory64 = NULL;
 #endif
-    int nArgs, i;
-    LPWSTR *szArglist = NULL;
     HANDLE hProcess = NULL;
-    WCHAR *commandLineContents = NULL;
-    WCHAR *currentDirectoryContent = NULL;
+    LPCVOID src;
+    SIZE_T size;
+    WCHAR *buffer = NULL;
 #ifdef _WIN64
     LPVOID ppeb32 = NULL;
 #else
+    PVOID64 src64;
     BOOL weAreWow64;
     BOOL theyAreWow64;
 #endif
-    PyObject *py_unicode = NULL;
-    PyObject *py_retlist = NULL;
 
     hProcess = psutil_handle_from_pid(pid);
     if (hProcess == NULL)
@@ -380,29 +383,20 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
         PyErr_SetFromWindowsErr(0);
         goto error;
     }
-#else
-    /* 32 bit case.  Check if the target is also 32 bit. */
-    if (!IsWow64Process(GetCurrentProcess(), &weAreWow64) ||
-        !IsWow64Process(hProcess, &theyAreWow64)) {
-        PyErr_SetFromWindowsErr(0);
-        goto error;
-    }
-#endif
 
-#ifdef _WIN64
     if (ppeb32 != NULL) {
         /* We are 64 bit.  Target process is 32 bit running in WoW64 mode. */
         PEB32 peb32;
         RTL_USER_PROCESS_PARAMETERS32 procParameters32;
 
         // read PEB
-        if(!ReadProcessMemory(hProcess, ppeb32, &peb32, sizeof(peb32), NULL)) {
+        if (!ReadProcessMemory(hProcess, ppeb32, &peb32, sizeof(peb32), NULL)) {
             PyErr_SetFromWindowsErr(0);
             goto error;
         }
 
         // read process parameters
-        if(!ReadProcessMemory(hProcess,
+        if (!ReadProcessMemory(hProcess,
                               UlongToPtr(peb32.ProcessParameters),
                               &procParameters32,
                               sizeof(procParameters32),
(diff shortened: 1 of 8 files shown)


psutil/arch/windows/process_info.c
```diff
@@ -273,6 +273,30 @@ typedef NTSTATUS (NTAPI *_NtWow64ReadVirtualMemory64)(
...
+*/
+
+typedef NTSTATUS (NTAPI *_NtWow64QueryVirtualMemory64)(
+        IN HANDLE ProcessHandle,
+        IN PVOID64 BaseAddress,
+        IN MEMORY_INFORMATION_CLASS MemoryInformationClass,
+        OUT PMEMORY_BASIC_INFORMATION64 MemoryInformation,
+        IN ULONG64 Size,
+        OUT PULONG64 ReturnLength OPTIONAL);
+
```
giampaolo: 4 space indentation

psutil/arch/windows/process_info.c
```diff
@@ -411,47 +432,28 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+            src = UlongToPtr(procParameters32.CommandLine.Buffer),
+            size = procParameters32.CommandLine.Length;
+            break;
+        case KIND_CWD:
+            src = UlongToPtr(procParameters32.CurrentDirectoryPath.Buffer);
+            size = procParameters32.CurrentDirectoryPath.Length;
+            break;
+        case KIND_ENVIRON:
+            src = UlongToPtr(procParameters32.env);
+            break;
```
giampaolo: please indent all the `case` clauses

psutil/arch/windows/process_info.c
```diff
@@ -411,47 +432,28 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
         }
     } else
 #else
+    /* 32 bit case.  Check if the target is also 32 bit. */
+    if (!IsWow64Process(GetCurrentProcess(), &weAreWow64) ||
+        !IsWow64Process(hProcess, &theyAreWow64)) {
+        PyErr_SetFromWindowsErr(0);
```
giampaolo: indent by 1

psutil/arch/windows/process_info.c
```diff
@@ -502,44 +516,18 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+            src64 = procParameters64.CommandLine.Buffer;
+            size = procParameters64.CommandLine.Length;
+            break;
+        case KIND_CWD:
+            src64 = procParameters64.CurrentDirectoryPath.Buffer,
+            size = procParameters64.CurrentDirectoryPath.Length;
+            break;
+        case KIND_ENVIRON:
+            src64 = procParameters64.env;
+            break;
```
giampaolo: indentation

psutil/arch/windows/process_info.c
```diff
@@ -560,7 +548,7 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
         }
 
         // read peb
-        if(!ReadProcessMemory(hProcess,
+        if (!ReadProcessMemory(hProcess,
                               pbi.PebBaseAddress,
                               &peb,
                               sizeof(peb),
```
giampaolo: indent by 1

psutil/arch/windows/process_info.c
```diff
@@ -570,7 +558,7 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
         }
 
         // read process parameters
-        if(!ReadProcessMemory(hProcess,
+        if (!ReadProcessMemory(hProcess,
                               peb.ProcessParameters,
                               &procParameters,
                               sizeof(procParameters),
```
giampaolo: indent by 1

psutil/arch/windows/process_info.c
```diff
@@ -579,104 +567,100 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+            src = procParameters.CommandLine.Buffer;
+            size = procParameters.CommandLine.Length;
+            break;
+        case KIND_CWD: 
+            src = procParameters.CurrentDirectoryPath.Buffer;
+            size = procParameters.CurrentDirectoryPath.Length;
+            break;
+        case KIND_ENVIRON:
+            src = procParameters.env;
+            break;
```
giampaolo: indent by 4

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+    }
+
+    // arglist parsed as array of UNICODE_STRING, so convert each to
+    // Python string object and add to arg list
+    py_retlist = PyList_New(nArgs);
+    if (py_retlist == NULL)
+        goto out;
+    for (i = 0; i < nArgs; i++) {
+        py_unicode = PyUnicode_FromWideChar(
+                szArglist[i], wcslen(szArglist[i]));
```
giampaolo: indentation should be multiple of 4

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+        if (py_unicode == NULL)
+            goto out;
+        PyList_SET_ITEM(py_retlist, i, py_unicode);
```
giampaolo: Mmmm I see this is a macro which doesn't check return value. For consistency with the rest of the code base I think it's better to use `PyList_Append` and check return value. Also, but I'm not entirely sure, I think you'll have to `Py_DECREF(py_unicode);` after that.

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+        py_unicode = NULL;
+    }
+
+    ret = py_retlist;
+    py_retlist = NULL;
+
+out:
```
giampaolo: for consistency, can you please rename this to "error"?

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+    LocalFree(szArglist);
+    free(data);
```
giampaolo: I think you should check `if (data != NULL)` first

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+    WCHAR *data = NULL;
+    SIZE_T size;
+
+    if (psutil_get_process_data(pid, KIND_ENVIRON, &data, &size) != 0)
+        goto out;
+
+    // convert wchar array to a Python unicode string
+    ret = PyUnicode_FromWideChar(data, size / 2);
+
+out:
```
giampaolo: Can you please rename this to "error"?

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+    free(data);
```
giampaolo: check `if (data != NULL)` first

psutil/arch/windows/process_info.c
```diff
@@ -687,16 +671,86 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+PyObject *psutil_get_environ(long pid) {
+    PyObject *ret = NULL;
+    WCHAR *data = NULL;
+    SIZE_T size;
+
+    if (psutil_get_process_data(pid, KIND_ENVIRON, &data, &size) != 0)
+        goto out;
+
+    // convert wchar array to a Python unicode string
+    ret = PyUnicode_FromWideChar(data, size / 2);
```
giampaolo: this is weird... you never "return".

test/_windows.py
```diff
@@ -556,6 +562,19 @@ def test_cwd_64(self):
...
 
+    def test_environ_32(self):
+        p = psutil.Process(self.proc32.pid)
+        e = p.environ()
+        self.assertIn("THINK_OF_A_NUMBER", e)
+        self.assertEquals(e["THINK_OF_A_NUMBER"], str(os.getpid()))
+
+    def test_environ_64(self):
+        p = psutil.Process(self.proc64.pid)
+        time.sleep(0.1)
```
giampaolo: you can use `wait_for_pid()` utility test function instead

psutil/arch/windows/process_info.c
```diff
@@ -579,104 +567,100 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
+#ifndef _WIN64
+    if (weAreWow64 && !theyAreWow64) {
+        // figure out the size of the env var block
+        MEMORY_BASIC_INFORMATION64 info64;
+
+        if (NtWow64QueryVirtualMemory64 == NULL) {
+            NtWow64QueryVirtualMemory64 =
+                (_NtWow64QueryVirtualMemory64)GetProcAddress(
+                        GetModuleHandleA("ntdll.dll"),
+                        "NtWow64QueryVirtualMemory64");
```
giampaolo: indentation should be a multiple of 4

psutil/arch/windows/process_info.c
```diff
@@ -579,104 +567,100 @@ static int psutil_get_parameters(long pid, PyObject **pcmdline,
...
-                    szArglist[i], wcslen(szArglist[i]));
-                if (py_unicode == NULL)
-                    goto error;
-                if (PyList_Append(py_retlist, py_unicode))
-                    goto error;
-                Py_CLEAR(py_unicode);
-            }
+
+        size = info64.RegionSize - ((char*)src64 - (char*)info64.BaseAddress);
+    } else
```
giampaolo: `else` should be on the next line

giampaolo:
This is cool! I like the "enum" refactoring you applied. I just added a bunch of stylistic comments but LGTM. Feel free to merge it after you address them.

psutil/arch/windows/process_info.c
fbenkstein: Hm.  `PyList_Append` seems pointless since we already know the correct size.  `PyList_SetItem`/`PyList_SET_ITEM` steal so there's no need to `Py_DECREF`.  Anyway, I'll revert it if you want since I changed for no real reason.

psutil/arch/windows/process_info.c
fbenkstein: Should this be "error" even though it's the success case also?

psutil/arch/windows/process_info.c
fbenkstein: I've seen these checks in the rest of the code.  Is there a reason for that?  free(NULL) is documented as a no-op.

psutil/arch/windows/process_info.c
giampaolo: Ahhh I understand the logic now. Then can you add a `goto out;` earlier (it's more explicit)?

psutil/arch/windows/process_info.c
giampaolo: In that case it's fine.

psutil/arch/windows/process_info.c
fbenkstein: There's no need to return since the success and error paths are the same.

giampaolo: then just add `goto out;` to be more explicit please

psutil/arch/windows/process_info.c
giampaolo: No it's fine; leave it as is.

test/_windows.py
fbenkstein: The sleep shouldn't have made it in.  There was actually a bug where the process would die before the ReadProcessMemory would finish triggering ERROR_PARTIAL_READ sporadically...

giampaolo: Hmm... this is interesting. If that's the case then maybe we should check that condition (ERROR_PARTIAL_READ) and raise NoSuchProcess if pid does not exist anymore. Makes sense?

giampaolo: ...you can use `psutil_pid_exists` for that BTW.

fbenkstein:
Thanks!  I'll post an update later :zzz:

fbenkstein:
Please have another look.

giampaolo:
All good. Merged. Thanks for the great work.

giampaolo:
@fbenkstein psutil 4.0.0 is out and I gave you credits for this great piece of code http://grodola.blogspot.com/2016/02/psutil-4-real-process-memory-and-environ.html 
Thanks again.

PR #764, opened 2016-02-10
NetBSD-6.x fixes for psutil

0-wiz-0:
I've recently compiled the latest psutil release on NetBSD-6.x (the previous, but still supported NetBSD release) and had to fix two issues, see patches. Please include them.


giampaolo:
Thanks for this. Please update also HISTORY.txt describing the problem which has being fixed (I guess it's gonna be "can't compile on NetBSD 6.x") and HISTORY file.

0-wiz-0:
Like this?

giampaolo:
Yes, thanks.

Issue #490, opened 2014-05-23
Expose cgroup as process attr


giampaolo:
_From [kap...@gmail.com](https://code.google.com/u/108276830347560657704/) on March 30, 2014 05:12:36_
_Original issue: http://code.google.com/p/psutil/issues/detail?id=490_

On linux, cgroup usage is increasingly common (systemd, lxc, docker), it would 
be nice to have that exposed on a process as an attribute.

ushachar:
@giampaolo - Are you open to PRs for resolving this?
(Asking since you rejected the PR for #765)
It will have to be linux specific (since cgroups are a linux construct), and I guess it will use the libcgroups python bindings internally.

One possible use case (out of many) would be to discover how much RAM is REALLY available for my process.
When running in a containerized/cgroup environment, the general memory usage/limits (which are discovered internally using sysinfo) are often incorrect.

giampaolo:
The main problem I have with cgroups is that I don't use/know them and it looks like a pretty extended subject. Before a PR I would like to see a discussion about what capabilities should be implemented and what the API should look like.

The main question is: should psutil be a fully featured cgroups library? My hunch is "no", because that would be too much and outside of the scope of psutil as a project. 

For instance, the API of libs like this one: 
https://github.com/cloudsigma/cgroupspy
...gives me the sensation that this cgroup thing is something big which deserves a separate, dedicated project. I don't want something like that in psutil.

On the other hand, something like this:
https://github.com/francisbouvier/cgroups
...intrigues me because it seems it could be relatively easy to associate a cgroup to a PID, and that looks like something which could be integrated in `psutil.Process` class somehow.

As to how to do that in terms of API, I'm not sure, and that's because I'm ignorant about how cgroups work exactly. Judging from https://github.com/francisbouvier/cgroups it seems they are bound to a name: you first create a cgroup, then you associate a PID to it. That's kind of a bummer because it requires a separate `Cgroup(name)` class of some sort.
That's why I'm calling for others who are more familiar than me with cgroups to think about a simple API and start a discussion.

giampaolo:
If this were to land in psutil it's likely gonna be a read-only method returning a dict. Something like:

```python
@wrap_exceptions
def cgroups(self):
    with open_binary("%s/%s/cgroup" % (self._procfs_path, self.pid)) as f:
        cgroups = {}
        for line in f:
            line = line.strip()
            cid, subsys, name = line.split(':', 2)
            if name.startswith('/'):
[...]

PR #765, opened 2016-02-10
Implement cgroups() method on linux

polachok:


giampaolo:
I'm always skeptical about adding features which are not cross platform unless they are very useful. I don't know anything about cgroups. Why are they important and why do you think psutil should support them  (e.g. provide a use case), also considering the info you provide in this PR is "read-only".

socketpair:
#490

PR #796, opened 2016-03-17
Add disk queue metrics to psutil for linux

awreece:
This enables more metrics like what would be seen in iostat -x

I investigating adding these metrics to OSX too: I _suspect_ that [`kIOBlockStorageDriverStatisticsLatentReadTimeKey`](https://developer.apple.com/library/mac/documentation/Kernel/Reference/IOBlockStorageDriver_header_reference/index.html) is what we want, but I couldn't find good documentation on it.

I ran make test on linux:

```
...<snip> ...
test_special_pid (test_windows.WindowsSpecificTestCase) ... skipped 'not a Windows system'
test_total_phymem (test_windows.WindowsSpecificTestCase) ... skipped 'not a Windows system'

Ran 358 tests in 6.033s

OK (skipped=144)
```

And also manually verified that the new functionality is working:

```
>>> psutil.disk_io_counters()
sdiskio(read_count=50430, write_count=9749, read_bytes=658885632, write_bytes=98039808, read_time=117556, write_time=4380, read_merged_count=10700, write_merged_count=4712, busy_time=14836, in_flight=0, time_in_queue=71580)
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -306,6 +306,9 @@ Disks
     (in milliseconds)
   - **busy_time**: (*Linux*, *FreeBSD*) time spent doing actual I/Os (in
     milliseconds)
+  - **time_in_queue**: (*Linux*) total wait time for all requests (in
+    milliseconds)
+  - **in_flight**: (*Linux*) number of I/Os currently in flight
   - **read_merged_count** (*Linux*): number of merged reads
     (see `iostat doc <https://www.kernel.org/doc/Documentation/iostats.txt>`__)
   - **write_merged_count** (*Linux*): number of merged writes
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -245,7 +245,7 @@ sdiskio = namedtuple('sdiskio', ['read_count', 'write_count',
                                  'read_bytes', 'write_bytes',
                                  'read_time', 'write_time',
                                  'read_merged_count', 'write_merged_count',
-                                 'busy_time'])
+                                 'busy_time', 'in_flight', 'time_in_queue'])
 popenfile = namedtuple('popenfile',
                        ['path', 'fd', 'position', 'mode', 'flags'])
 pmem = namedtuple('pmem', 'rss vms shared text lib data dirty')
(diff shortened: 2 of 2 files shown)


awreece:
bump?

psutil/_pslinux.py
```diff
@@ -785,6 +785,20 @@ def net_if_stats():
...
+    """Returns a tuple of (ios_in_flight, time_in_queue_ms) for a disk."""
+    try:
+        with open_text("/sys/block/%s/stat" % device) as f:
+            line = f.readlines()[0]
+        fields = line.split()
+        (reads, read_merges, read_sectors, read_ticks,
+         writes, write_merges, write_sectors, write_ticks,
+         in_flight, io_ticks, time_in_queue) = map(int, fields[:11])
+        return (in_flight, time_in_queue)
+    except IOError:
```
giampaolo: please move this up so that only the "open_text()" call is wrapped

psutil/_pslinux.py
```diff
@@ -785,6 +785,20 @@ def net_if_stats():
 
 # --- disks
 
+def disk_block_stats(device):
+    """Returns a tuple of (ios_in_flight, time_in_queue_ms) for a disk."""
```
giampaolo: Please add a reference to https://www.kernel.org/doc/Documentation/block/stat.txt in the docstring.

psutil/_pslinux.py
```diff
@@ -785,6 +785,20 @@ def net_if_stats():
...
+    try:
+        with open_text("/sys/block/%s/stat" % device) as f:
+            line = f.readlines()[0]
```
giampaolo: You can simply use `f.readline()` instead

docs/index.rst
```diff
@@ -306,6 +306,9 @@ Disks
     (in milliseconds)
   - **busy_time**: (*Linux*, *FreeBSD*) time spent doing actual I/Os (in
     milliseconds)
+  - **time_in_queue**: (*Linux*) total wait time for all requests (in
+    milliseconds)
+  - **in_flight**: (*Linux*) number of I/Os currently in flight
```
giampaolo: In order to be consistent with the established naming convention this field should be called `in_flight_count`.
`in_flight_count` looks like a good name (well, I don't like it, but I can't think of anything better).
Also, change the explanation to something more meaningful such as:
"The number of I/O requests that have been issued to the device driver but have not yet completed."

docs/index.rst
```diff
@@ -306,6 +306,9 @@ Disks
     (in milliseconds)
   - **busy_time**: (*Linux*, *FreeBSD*) time spent doing actual I/Os (in
     milliseconds)
+  - **time_in_queue**: (*Linux*) total wait time for all requests (in
+    milliseconds)
```
giampaolo: In order to be consistent with the established naming convention this field should be called `queue_time`.
Also change the explanation to "number of milliseconds that I/O requests have waited on this block device".

psutil/_pslinux.py
```diff
@@ -785,6 +785,20 @@ def net_if_stats():
...
+        fields = line.split()
+        (reads, read_merges, read_sectors, read_ticks,
+         writes, write_merges, write_sectors, write_ticks,
+         in_flight, io_ticks, time_in_queue) = map(int, fields[:11])
```
giampaolo: I'm wondering whether it makes sense to return also `io_ticks`. I would be OK with that if `iostat -x` returns it but I'm not sure what's the column of reference (`svctm` perhaps?).

giampaolo:
Sorry for taking a look at this so late. I added some comments to your PR.

PR #801, opened 2016-03-31
Add development library for pcap to INSTALL instructions

psivesely:
The `pcap.h` library is needed to build `psutil`.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -8,12 +8,12 @@ Linux
 
 Ubuntu / Debian::
 
-    $ sudo apt-get install gcc python-dev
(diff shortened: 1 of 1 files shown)


fowlslegs:
I built `psutil` in a headless Ubuntu 14.04 container with `pip` and found it required `libpcap-dev`. I never tested on RHEL (I just looked up the equivalent package name).

giampaolo:
Libpcap, the library for sniffing networks traffic? Why do you think psutil requires that?

fowlslegs:
Woops, dealing with a large amount of output. That was needed for `pcapy`.

PR #815, opened 2016-05-06
add STAT for ps.py

chenhanxiao:
Add STAT to display  the  state of a process.

Signed-off-by: Chen Hanxiao chenhanxiao@gmail.com

### DIFF
--- a/scripts/ps.py
+++ b/scripts/ps.py
@@ -17,17 +17,37 @@ import time
 
 import psutil
 
+PROC_STATUSES_RAW = {
+    psutil.STATUS_RUNNING: "R",
+    psutil.STATUS_SLEEPING: "S",
+    psutil.STATUS_DISK_SLEEP: "D",
+    psutil.STATUS_STOPPED: "T",
+    psutil.STATUS_TRACING_STOP: "t",
+    psutil.STATUS_ZOMBIE: "Z",
+    psutil.STATUS_DEAD: "X",
+    psutil.STATUS_WAKING: "WA",
+    psutil.STATUS_IDLE: "I",
+    psutil.STATUS_LOCKED: "L",
+    psutil.STATUS_WAITING: "W",
+}
+
+if hasattr(psutil, 'STATUS_WAKE_KILL'):
+    PROC_STATUSES_RAW[psutil.STATUS_WAKE_KILL] = "WK"
+
+if hasattr(psutil, 'STATUS_SUSPENDED'):
+    PROC_STATUSES_RAW[psutil.STATUS_SUSPENDED] = "V"
+
 
 def main():
     today_day = datetime.date.today()
-    templ = "%-10s %5s %4s %4s %7s %7s %-13s %5s %7s  %s"
+    templ = "%-10s %5s %4s %4s %7s %7s %-13s %-5s %5s %7s  %s"
     attrs = ['pid', 'cpu_percent', 'memory_percent', 'name', 'cpu_times',
-             'create_time', 'memory_info']
+             'create_time', 'memory_info', 'status']
     if os.name == 'posix':
         attrs.append('uids')
         attrs.append('terminal')
     print(templ % ("USER", "PID", "%CPU", "%MEM", "VSZ", "RSS", "TTY",
-                   "START", "TIME", "COMMAND"))
+                   "STAT", "START", "TIME", "COMMAND"))
     for p in psutil.process_iter():
         try:
             pinfo = p.as_dict(attrs, ad_value='')
@@ -64,6 +84,7 @@ def main():
                 int(pinfo['memory_info'].rss / 1024) or '?'
             memp = pinfo['memory_percent'] and \
                 round(pinfo['memory_percent'], 1) or '?'
+            status = PROC_STATUSES_RAW.get(pinfo['status'], pinfo['status'])
             print(templ % (
                 user[:10],
                 pinfo['pid'],
@@ -72,6 +93,7 @@ def main():
                 vms,
                 rss,
                 pinfo.get('terminal', '') or '?',
+                status,
                 ctime,
                 cputime,
                 pinfo['name'].strip() or '?'))


giampaolo:
There's no need to modify _pslinux.py. The change should happen in ps.py only.

chenhanxiao:
Ah, thanks.
I'll post a v2.

chenhanxiao:
@giampaolo 
Any comments？
btw: this PR did not bring new ci failures.

giampaolo:
I am flying to US. I will try to take a look at this sometime during next week.

chenhanxiao:
@giampaolo 
ping

chenhanxiao:
@giampaolo 
hi, any comments?

scripts/ps.py
```diff
@@ -17,17 +17,43 @@
...
+STATUS_STOPPED = "stopped"
+STATUS_TRACING_STOP = "tracing-stop"
+STATUS_ZOMBIE = "zombie"
+STATUS_DEAD = "dead"
+STATUS_WAKE_KILL = "wake-kill"
+STATUS_WAKING = "waking"
+STATUS_IDLE = "idle"  # FreeBSD, OSX
+STATUS_LOCKED = "locked"  # FreeBSD
+STATUS_WAITING = "waiting"  # FreeBSD
+STATUS_SUSPENDED = "suspended"  # NetBSD
```
giampaolo: there's no need to redefine all these; they are defined in the psutil namespaces aready (use `psutil.STATUS_SUSPENDED`, etc.)

scripts/ps.py
```diff
@@ -17,17 +17,43 @@
...
+PROC_STATUSES_RAW = {
+    STATUS_RUNNING:             "R",
+    STATUS_SLEEPING:            "S",
+    STATUS_DISK_SLEEP:          "D",
+    STATUS_STOPPED:             "T",
+    STATUS_TRACING_STOP:        "t",
+    STATUS_ZOMBIE:              "Z",
+    STATUS_DEAD:                "X",
+    STATUS_WAKE_KILL:           "K",
+    STATUS_WAKING:              "W"
```
giampaolo: these are not PEP-8 compliant. Redefine the above by removing unnecessary spaces (`STATUS_WAKING: "W"` etc.)

scripts/ps.py
```diff
@@ -17,17 +17,28 @@
...
+PROC_STATUSES_RAW = {
+    psutil.STATUS_RUNNING: "R",
+    psutil.STATUS_SLEEPING: "S",
+    psutil.STATUS_DISK_SLEEP: "D",
+    psutil.STATUS_STOPPED: "T",
+    psutil.STATUS_TRACING_STOP: "t",
+    psutil.STATUS_ZOMBIE: "Z",
+    psutil.STATUS_DEAD: "X",
+    psutil.STATUS_WAKING: "W"
+}
```
giampaolo: This list is incomplete. It's missing:

```
psutil.STATUS_WAKE_KILL
psutil.STATUS_WAKING
psutil.STATUS_IDLE(OSX, FreeBSD)
psutil.STATUS_LOCKED(FreeBSD)
psutil.STATUS_WAITING(FreeBSD)
psutil.STATUS_SUSPENDED(NetBSD)
```

You should also take that into account. You can probably come up with new letters.

scripts/ps.py
```diff
@@ -64,6 +75,7 @@ def main():
                 int(pinfo['memory_info'].rss / 1024) or '?'
             memp = pinfo['memory_percent'] and \
                 round(pinfo['memory_percent'], 1) or '?'
+            stat = PROC_STATUSES_RAW.get(pinfo['status'], '')
```
giampaolo: please rename the variable to "status". "stat" is a stdlib module and it's bad practice to obfuscate "reserved" names.

giampaolo: Also if the letter is not available simply return the original `p['status']` as in:

```
PROC_STATUSES_RAW.get(pinfo['status'], pinfo['status'])
```

scripts/ps.py
chenhanxiao: I checked take man page of ps in linux for a reference.
Found nothing about the status above.
Could you give me some hints?

giampaolo: These are from other UNIX platforms (FreeBSD, OSX). psutil exposes them also if you're on Linux. You should take all posible statuses into account.

chenhanxiao: OK, I'll download them and have a look.
Thanks.

chenhanxiao: I updated these:
psutil.STATUS_IDLE(OSX, FreeBSD):		I
psutil.STATUS_LOCKED(FreeBSD):		L
psutil.STATUS_WAITING(FreeBSD):			W
psutil.STATUS_SUSPENDED(NetBSD):			V

But:
psutil.STATUS_WAKE_KILL
psutil.STATUS_WAKING
the ps cmd of Linux did not take into account.

Maybe we should do the same thing.

giampaolo: No we shouldn't. Come up with some letter. `psutil.STATUS_WAKE_KILL` can be mapped to `"WK" and `psutil.STATUS_WAKING` could be `"WA"`.

scripts/ps.py
```diff
@@ -17,17 +17,33 @@
...
+    psutil.STATUS_STOPPED: "T",
+    psutil.STATUS_TRACING_STOP: "t",
+    psutil.STATUS_ZOMBIE: "Z",
+    psutil.STATUS_DEAD: "X",
+    psutil.STATUS_WAKING: "WA",
+    psutil.STATUS_IDLE: "I",
+    psutil._common.STATUS_WAKE_KILL: "WK",
+    psutil.STATUS_LOCKED: "L",
+    psutil.STATUS_WAITING: "W",
+    psutil._common.STATUS_SUSPENDED: "V"
```
giampaolo: Don't do this. Instead do:

```
if hasattr(psutil, 'STATUS_SUSPENDED'):
    PROC_STATUSES_RAW[psutil.STATUS_SUSPENDED] = "V"
```

...same for `STATUS_WAKE_KILL`.

giampaolo:
OK, looks good. Merging. Thanks.

PR #825, opened 2016-05-24
Fix possible double close and use of unopened socket

hardikar:
- On line 407, I assume the code should jump to error if either value is -1 _or_ python error-ed, and not when both.
- According to http://linux.die.net/man/3/socket, `Upon successful completion, socket() shall return a non-negative integer, the socket file descriptor. Otherwise, a value of -1 shall be returned and errno set to indicate the error.`.  Thus on line 540, 0 is a valid socket fd - and the check should be against -1
- If py_retlist > 1, i.e Py_BuildValue returned an error, the code will jump to error and double close the socket.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -19,6 +19,7 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 - #797: [Linux] net_if_stats() may raise OSError for certain NIC cards.
 - #813: Process.as_dict() should ignore extraneous attribute names which gets
   attached to the Process instance.
+- #825: Fix possible double close and use of unopened socket 
 
 
 4.1.0 - 2016-03-12
--- a/psutil/_psutil_linux.c
+++ b/psutil/_psutil_linux.c
@@ -538,17 +538,16 @@ psutil_net_if_stats(PyObject* self, PyObject* args) {
         }
     }
 
-    close(sock);
     py_retlist = Py_BuildValue("[Oiii]", py_is_up, duplex, speed, mtu);
     if (!py_retlist)
-        goto error_after_close;
+        goto error;
+    close(sock);
     Py_DECREF(py_is_up);
     return py_retlist;
 
 error:
     if (sock != -1)
         close(sock);
-error_after_close:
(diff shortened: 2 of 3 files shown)


psutil/_psutil_linux.c
```diff
@@ -541,14 +541,15 @@ psutil_net_if_stats(PyObject* self, PyObject* args) {
...
     Py_DECREF(py_is_up);
     return py_retlist;
 
 error:
-    Py_XDECREF(py_is_up);
-    if (sock != 0)
+    if (sock != -1)
         close(sock);
+error_after_close:
+    Py_XDECREF(py_is_up);
```
giampaolo: I kinda don't like having two separate goto handlers. I think `close()`sing the socket twice will not be a problem,

hardikar: Well, once a socket is closed - the file descriptor is available for use elsewhere. And thus a double close could end up closing that - which may cause a problem elsewhere. If you don't like multiple goto handlers, you might prefer moving the close(sock) on line 541 down to 545 before the return. That way there is a close in every path in the function - and no double close.

giampaolo: I like. Can you update the PR also including changes for the HISTORY.rst file?

hardikar:
@giampaolo Sorry for getting back to this after so long - have look at the updated diff.

HISTORY.rst
```diff
@@ -19,6 +19,7 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 - #797: [Linux] net_if_stats() may raise OSError for certain NIC cards.
 - #813: Process.as_dict() should ignore extraneous attribute names which gets
   attached to the Process instance.
+- #825: Fix possible double close and use of unopened socket
```
giampaolo: Please specify the platform (`[Linux]`) and the function which was affected (the high level, end-user psutil function, not the C function)

giampaolo:
Just fix this https://github.com/giampaolo/psutil/pull/825/files#r66893740 and update CREDITS file then it's good to go.

PR #822, opened 2016-05-22
Fixed bug in NetBSD's execution of virtual_memory()

khanzf:
virtual_memory() opened /proc/meminfo as a binary. As such, each line was a binary, but lines 135 and 137 were checking against a string type. This caused str.startswith() to raise a TypeError and fail.
The solution was merely to change the str.startswith() comparison to a binary.

PR #833, opened 2016-06-03
Fix psutil.virtual_memory() type mismatch for NetBSD.

khanzf:
psutil.virtual_memory() on NetBSD opened /proc/meminfo as a binary file, but was using startswith() against a non-binary string. This resulted in a TypeError. This match changes the startswith() string to a bytes type.


FarhansCode:
This is a revision of pull request #822 which I closed due to multiple errors in my git history.

Issue #816, opened 2016-05-08
[Windows] psutil.net_io_counters I/O is limited to 32bit unsigned integer because of MIB_IFROW


jhomann:
Interesting problem I have seen for a while with using psutil. I'm by no means an expert with this but it is strange that when I approach 4 billion bytes it seems to rollover... I thought Python accounted for 32-bit integers and whatnot. Can someone explain if this can be accounted for?

What I'm using:
Python 2.7.11 (tried both 32 and 64 bit)
Windows 10 64 bit
psutil 4.1.0

```
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:40:30) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.

>>> import psutil
>>> psutil.net_io_counters(pernic=True)
>>> {'Ethernet 4': snetio(bytes_sent=378813079, bytes_recv=838898995, packets_sent=4236049, packets_recv=20927779, errin=0, errout=0, dropin=212, dropout=0)}
```

Is this a product of Windows? Are the counters only a certain size? Because in my windows ethernet status screen under 'Activity' it shows:

Bytes sent: 337,887,481
Byes recieved: 30,797,026,512

While the bytes_sent is off it doesn't surprise me that there may be small differences but the bytes_recv about a hour ago showed a value of around 3,400,000,000 bytes and now it is only showing in the 800 millions.

jhomann:
Okay I did a little digging and I am wondering if we can work around this; for windows the net IO counters are coming from a MIB, MIB_IFROW which uses dword ([apparently 32-bit unsigned integer with a max of 4.3bil decimal](pxhttps://msdn.microsoft.com/en-us/library/cc230318.aspx)) and since I haven't written anything in C/C++ I am not as familiar with all this. However, I do believe that there are other MIBs available with the values in larger types.

The structure of [MIB_IF_ROW2](https://msdn.microsoft.com/en-us/library/windows/desktop/aa814491(v=vs.85).aspx) seems to have much larger numbers and in Microsofts documentation it does say to use this one if you have a NIC > x amount.

Would it be possible to check for the existence of the MIB_IF_ROW2 and use that, and if it doesn't exist fall back on the MIB_IFROW since it looks like the updated one supports Windows Vista + but I do believe this would give the proper value for the output of octets/bytes for interfaces with > 4.3GB throughput.

jhomann:
[...]

PR #835, opened 2016-06-08
[Windows] Fix for net_io_counters wrapping after 4.3GB due to MIB_IFROW

jhomann:
I am updating this with a new pull request. The other one was very old and didn't quite work with the current version of psutil. See more #816

### DIFF
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -18,12 +18,12 @@
 #include <tchar.h>
 #include <tlhelp32.h>
 #include <winsock2.h>
-#if (_WIN32_WINNT >= 0x0600)
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
 #include <ws2tcpip.h>
 #endif
 #include <iphlpapi.h>
 #include <wtsapi32.h>
-#if (_WIN32_WINNT < 0x0600)
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
 #include <ws2tcpip.h>
 #endif
 #include <Winsvc.h>
@@ -90,7 +90,7 @@ typedef struct _DISK_PERFORMANCE_WIN_2008 {
 
 // --- network connections mingw32 support
 #ifndef _IPRTRMIB_H
-#if (_WIN32_WINNT < 0x0600)
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
 typedef struct _MIB_TCP6ROW_OWNER_PID {
     UCHAR ucLocalAddr[16];
     DWORD dwLocalScopeId;
@@ -135,7 +135,7 @@ typedef struct _MIB_UDPTABLE_OWNER_PID {
 } MIB_UDPTABLE_OWNER_PID, *PMIB_UDPTABLE_OWNER_PID;
 #endif
 
-#if (_WIN32_WINNT < 0x0600)
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
 typedef struct _MIB_UDP6ROW_OWNER_PID {
     UCHAR ucLocalAddr[16];
     DWORD dwLocalScopeId;
@@ -2176,10 +2176,10 @@ return_:
 static PyObject *
 psutil_net_io_counters(PyObject *self, PyObject *args) {
     DWORD dwRetVal = 0;
-#if (_WIN32_WINNT >= 0x0600)
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
     MIB_IF_ROW2 *pIfRow = NULL;
 #endif
-#if (_WIN32_WINNT < 0x0600)
+#if (_WIN32_WINNT < 0x0600) // Windows 2000 / XP
     MIB_IFROW *pIfRow = NULL;
 #endif
     PIP_ADAPTER_ADDRESSES pAddresses = NULL;
@@ -2198,10 +2198,10 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
     while (pCurrAddresses) {
         py_nic_name = NULL;
         py_nic_info = NULL;
-#if (_WIN32_WINNT >= 0x0600)
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
         pIfRow = (MIB_IF_ROW2 *) malloc(sizeof(MIB_IF_ROW2));
 #endif
-#if (_WIN32_WINNT < 0x0600)
+#if (_WIN32_WINNT < 0x0600) // Windows 2000 / XP
 		pIfRow = (MIB_IFROW *) malloc(sizeof(MIB_IFROW));
 #endif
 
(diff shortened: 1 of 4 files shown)


giampaolo:
Mmm have you tested this on Win XP? I don't see the conditionals around MIB_IF_ROW2 I would expect. Also please update HISTORY file describing what have been fixed here.

jhomann:
Oh sorry I thought you read my post in #816 , _psutils 4.0.0+ already does not work with Windows XP_
 psutils 4.0.0+ already does not work with Windows XP so I put this PR in in. like me to try making 4.3.0 compatible with XP I can.

jhomann:
Interesting, it seems like although the pip installs didn't work ... manually compiling it on Win XP does still work. I will need to find out a good way to do this...

jhomann:
I am so sorry for all those pushes in a row! It was the easiest way to test it across multiple systems that I had set up to test compiling with. Compiling is working in both Win XP and Win 7 (my normal workstation) though so hopefully the final tests will show it building OK.

jhomann:
Well that's crap. I fixed the tests but it looks like the final test in AppVeyor gave an error (which I think is completely false?)

```
FAIL: test_parent_ppid (test_process.TestProcess)
  File "C:\projects\psutil\psutil\tests\test_process.py", line 1145, in test_parent_ppid
    self.assertNotEqual(p.ppid(), this_parent)
AssertionError: 1996 == 1996
 
Ran 354 tests in 8.188s
 
FAILED (failures=1, skipped=198)
```

Which is interesting since the last few commits worked just fine and that was never edited by me.

jhomann:
For the record I am done with my commits, I believe everything is in order. The tests that are failing now seem like random false positives. please take a look at let me know if you would like anything else done for it.

psutil/_psutil_windows.c
```diff
@@ -18,9 +18,14 @@
...
 #include <tlhelp32.h>
 #include <winsock2.h>
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
+#include <ws2tcpip.h>
+#endif
 #include <iphlpapi.h>
 #include <wtsapi32.h>
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
 #include <ws2tcpip.h>
+#endif
```
giampaolo: I don't understand: you include the same module no matter what the windows version is so why using ifdef in the first place?

psutil/_psutil_windows.c
```diff
@@ -85,6 +90,7 @@ typedef struct _DISK_PERFORMANCE_WIN_2008 {
 
 // --- network connections mingw32 support
 #ifndef _IPRTRMIB_H
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
```
giampaolo: Just specify "Windows XP". Win 2000 support is long gone.

psutil/_psutil_windows.c
```diff
@@ -128,6 +135,7 @@ typedef struct _MIB_UDPTABLE_OWNER_PID {
 } MIB_UDPTABLE_OWNER_PID, *PMIB_UDPTABLE_OWNER_PID;
 #endif
 
+#if (_WIN32_WINNT < 0x0600) // Windows XP / 2000
```
giampaolo: Just specify "Windows XP". Win 2000 support is long gone.

psutil/_psutil_windows.c
```diff
@@ -2168,7 +2176,12 @@ psutil_disk_usage(PyObject *self, PyObject *args) {
 static PyObject *
 psutil_net_io_counters(PyObject *self, PyObject *args) {
     DWORD dwRetVal = 0;
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
+    MIB_IF_ROW2 *pIfRow = NULL;
+#endif
+#if (_WIN32_WINNT < 0x0600) // Windows 2000 / XP
     MIB_IFROW *pIfRow = NULL;
+#endif
```
giampaolo: You can use a single `ifdef/else/endif` block here.

psutil/_psutil_windows.c
```diff
@@ -2185,20 +2198,43 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
     while (pCurrAddresses) {
         py_nic_name = NULL;
         py_nic_info = NULL;
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
+        pIfRow = (MIB_IF_ROW2 *) malloc(sizeof(MIB_IF_ROW2));
+#endif
+#if (_WIN32_WINNT < 0x0600) // Windows 2000 / XP
         pIfRow = (MIB_IFROW *) malloc(sizeof(MIB_IFROW));
+#endif
```
giampaolo: same here

psutil/_psutil_windows.c
```diff
@@ -2185,20 +2198,43 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
...
-
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista, 7, 8, 8.1, 10
+        SecureZeroMemory((PVOID)pIfRow, sizeof(MIB_IF_ROW2));
+        pIfRow->InterfaceIndex = pCurrAddresses->IfIndex;
+        dwRetVal = GetIfEntry2(pIfRow);
+#endif
+#if (_WIN32_WINNT < 0x0600) // Windows 2000 / XP
         pIfRow->dwIndex = pCurrAddresses->IfIndex;
         dwRetVal = GetIfEntry(pIfRow);
+#endif
```
giampaolo: ...and here

psutil/_psutil_windows.c
```diff
@@ -2185,20 +2198,43 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
...
         if (dwRetVal != NO_ERROR) {
-            PyErr_SetString(PyExc_RuntimeError, "GetIfEntry() failed.");
+            PyErr_SetString(PyExc_RuntimeError, "GetIfEntry() or GetIfEntry2() failed.");
```
giampaolo: line > 80 chars; please indent this

psutil/_psutil_windows.c
```diff
@@ -2208,6 +2244,8 @@ psutil_net_io_counters(PyObject *self, PyObject *args) {
                                     pIfRow->dwOutErrors,
                                     pIfRow->dwInDiscards,
                                     pIfRow->dwOutDiscards);
+#endif
```
giampaolo: use a single ifdef/else/endif block

psutil/tests/test_posix.py
```diff
@@ -314,19 +314,25 @@ def test_os_waitpid_bad_ret_status(self):
...
         tolerance = 4 * 1024 * 1024  # 4MB
         for part in psutil.disk_partitions(all=False):
             usage = psutil.disk_usage(part.mountpoint)
-            total, used, free, percent = df(part.device)
+            try:
+                total, used, free, percent = df(part.device)
+            except RuntimeError:
+                # Issue with OS X not being able to read certain partitions
+                # Issue with Linux systems not able to read Docker mappings
+                continue
```
giampaolo: I don't like this as it can potentially `continue` all iterations practially skipping the whole test. You should use some other logic rather than the mere `RuntimeError` catch. Specifically you should inspect the cmd stderr which you can pass along the RuntimeError exception. Since this is probably too much troubles I would revert this change as it's unrelated with the Windows issue (feel free to open a new one if you want).

giampaolo:
I added some inline comments. If you reintroduced Win XP support which was broken please specify this in HISTORY. ALso please update CREDITS including your name and contribution. Thanks.

psutil/_psutil_windows.c
jhomann: This actually doesn't need to be included. The issue was that it matters where it's located in the list of includes, but after further testing, this header is not required for Win XP.

jhomann:
I've updated with the proper changes! I definitely should have looked up the else, that was stupid of me. Again, sorry for making so many commits for this change.

As for re-introducing Win XP support, it seems that building 4.0.0-4.2.0 works however installing it from pip does not. That being said, it still had support for XP if you built it manually, which I tried and it worked, so we are definitely good there.

giampaolo:
You can do as many pushes as you want: AppVeyor tests are for free and are designed exactly for that (make our life easier). =) Patch looks great, I'm merging it. Thanks a lot for working on this. It was a long standing issue and the solution you came up with is perfect. If you will ever feel about taking some other Windows issue from the bug tracker you're more than welcome.

PR #836, opened 2016-06-12
Potential fix for test_disk_usage on Linux and Mac OS X

jhomann:
From your comments on the other pull, this version checks the output of the df error for specific values that would be in the error if it was caused by docker builds and lets any other error pass through.

### DIFF
--- a/psutil/tests/test_posix.py
+++ b/psutil/tests/test_posix.py
@@ -315,7 +315,7 @@ class TestSystemAPIs(unittest.TestCase):
     def test_disk_usage(self):
         def df(device):
             # Use 1 kB block sizes since OS X doesn't have -B flag
-            out = sh("df -k 1 %s" % device).strip()
+            out = sh("df -k %s" % device).strip()
             line = out.split('\n')[1]
             fields = line.split()
             total = int(fields[1]) * 1024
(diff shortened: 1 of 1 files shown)


psutil/tests/test_posix.py
```diff
@@ -314,19 +314,27 @@ def test_os_waitpid_bad_ret_status(self):
...
             usage = psutil.disk_usage(part.mountpoint)
-            total, used, free, percent = df(part.device)
+            try:
+                total, used, free, percent = df(part.device)
+            except RuntimeError as err:
+                # Issue on Mac OS X and Linux not being able to read certain
+                # locations such as /dev/mapper/docker* so skip them
+                if ('/dev/mapper/docker' in str(err) and
+                    'No such file or directory' in str(err)):
+                    continue
```
giampaolo: Add an `else:` here and indent the block below so that you'll fix `UnboundLocalError: local variable 'total' referenced before assignment` error.

jhomann:
I didn't like doing it via the RuntimeError so I tried just skipping the docker mappings but it looks like there are other devices that df cannot read. I'm just going to close this for now since I am unsure of how the travis ci dockers work.

`RuntimeError: df: /dev/disk0s2: Raw devices not supported`
Also
`RuntimeError: df: `/dev/xvdb': No such file or directory`

PR #838, opened 2016-06-14
overrides check_output returning bytes by default

tijko:
On python 3.x > when running the `.git-pre-commit` the `subprocess.check_output` returns bytes by default.  This overrides the default behavior by setting the `universal_newlines` to `True`.

### DIFF
--- a/.git-pre-commit
+++ b/.git-pre-commit
@@ -10,9 +10,8 @@ import sys
 
 
 def main():
-    out = subprocess.check_output("git diff --cached --name-only", shell=True,
-                                                      universal_newlines=True)
-    files = [x for x in out.split('\n') if x.endswith('.py') and
+    out = subprocess.check_output("git diff --cached --name-only", shell=True)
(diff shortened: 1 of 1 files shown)


.git-pre-commit
```diff
@@ -10,7 +10,8 @@ import sys
 
 
 def main():
-    out = subprocess.check_output("git diff --cached --name-only", shell=True)
+    out = subprocess.check_output("git diff --cached --name-only", shell=True,
+                                                      universal_newlines=True)
```
giampaolo: what happens if you do this instead?
```
- files = [x for x in out.split('\n') if x.endswith('.py') and
+ files = [x for x in out.split(b'\n') if x.endswith('.py') and
```

tijko: Converting the string arguments to bytes would work correct the error too.

`files = [x for x in out.split(b'\n') if x.endswith(b'.py') and`

Issue #783, opened 2016-02-28
Process.status() returns 'running' for a ZombieProcess


the-gigi:
This is on a Mac OSX El Capitan Here is a quick session to demonstrate. Process 70 on my machine right now is indeed a zombie process.

```
Python 2.7.10 (default, Jun 10 2015, 19:43:32)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> psutil.__version__
'4.0.0'
>>> p = psutil.Process(pid=70)
>>> p.status()
'running'
>>> p.name()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/gigi/.virtualenvs/work/lib/python2.7/site-packages/psutil/__init__.py", line 564, in name
    cmdline = self.cmdline()
  File "/Users/gigi/.virtualenvs/work/lib/python2.7/site-packages/psutil/__init__.py", line 617, in cmdline
    return self._proc.cmdline()
  File "/Users/gigi/.virtualenvs/work/lib/python2.7/site-packages/psutil/_psosx.py", line 207, in wrapper
    raise ZombieProcess(self.pid, self._name, self._ppid)
psutil.ZombieProcess: psutil.ZombieProcess process still exists but it's a zombie (pid=70)
```

I followed the code to the status() method in _psxosx.py to see why it returns 'running'

```
    @wrap_exceptions
    def status(self):
        code = cext.proc_status(self.pid)
        # XXX is '?' legit? (we're not supposed to return it anyway)
        return PROC_STATUSES.get(code, '?')
```

The cext.proc_status(self.pid) returns 2, which is the code for 'running'.

giampaolo:
Mmmm if that is true it's the kernel which is returning the bad info.
There's a test specific for this: https://github.com/giampaolo/psutil/blob/5c772a131269285b57f81f774e4aaf24fb30ba0b/psutil/tests/test_process.py#L1287
....which explicitly test this case (`assert status() == ZOMBIE`):
https://github.com/giampaolo/psutil/blob/5c772a131269285b57f81f774e4aaf24fb30ba0b/psutil/tests/test_process.py#L1325
Can you run tests (`make test`) and see if that fails?

giampaolo:
Bump?

zachriggle:
Many tests currently fail immediately OS X 10.11 El Capitain.

[...]

PR #855, opened 2016-07-06
Handle no-permission when checking pid on osx

heewa:
## Why

Before this change, on OS X, dealing with a proc the user doesn't have permission to send signals to incorrectly throws `ZombieProcess` errors. After this change, we can now do stuff like get the name of those processes.

This should fix issue #783.
## What

When checking whether a process with a pid exists on OS X using `kill(pid, 0)`, access & permission errors only indicate that we don't have permission to send signals to that process, not that it doesn't exist. So, the `psutil_pid_exists()` fn will now return true (`1`) in that case as well.

The comment in the function probably didn't work because we need to check `errno`, not the returned value from the `kill` fn. A lot of tests are broken on OSX, but not the test that comment mentions (`test_halfway_terminated`), which still passes with these changes.

### DIFF
--- a/psutil/arch/osx/process_info.c
+++ b/psutil/arch/osx/process_info.c
@@ -34,24 +34,14 @@ psutil_pid_exists(long pid) {
     int ret;
     if (pid < 0)
         return 0;
+
+    // Permission/access errors only indicate that we cannot send signals
+    // to that  process, but it does exist.
     ret = kill(pid , 0);
-    if (ret == 0)
+    if (ret == 0 || errno == EPERM || errno == EACCES)
         return 1;
-    else {
-        return 0;
-        /*
-        // This is how it is handled on other POSIX systems but it causes
-        // test_halfway_terminated test to fail with AccessDenied.
-        if (ret == ESRCH)
-            return 0;
-        else if (ret == EPERM)
-            return 1;
-        else {
-            PyErr_SetFromErrno(PyExc_OSError);
-            return -1;
-        }
-        */
-    }
+
+    return 0;
 }
 
 


giampaolo:
I will not have time to look into this too soon as I will have to reinstall OSX on VirtualBox (and it's always a pain) but it's good to know the nature of this issue has been revealed. Speaking of which, this same function is defined multiple times for basically all UNIXes. It would be nice to refactor it and define it in one place only (`_psutil_posix.c`). Also, I believe it should look more or less like this:

```
/*
 * Return 1 if PID exists in the current process list, else 0, -1
 * on error.
 * TODO: this should live in _psutil_posix.c but for some reason if I
 * move it there I get a "include undefined symbol" error.
 */
int
psutil_pid_exists(long pid) {
    int ret;
    if (pid < 0)
        return 0;
    ret = kill(pid , 0);
    if (ret == 0)
        return 1;
    else {
        if (ret == ESRCH) {
            // ESRCH == No such process
            return 0;
       }
        else if ((ret == EPERM) || (ret == EACCES)) {
            // EPERM / EACCES clearly means there's a process to deny access to
            return 1;
        }
        else {
            return -1;
        }
    }
}
```

Can you please update the PR?

heewa:
Sure, I can do those things. A couple questions, though:

1. Considering the style of the rest of the code, it seems more consistent for this util fn to not set the `PyErr`, and let the caller choose how to handle what it wants to do between itself and the python layer. So, basically, maybe it should return -1, and allow the caller to set `PyErr` or not.
2. Because there are now 3 possible return values, all the call locations have to change. I can either use an `else if` chain, which requires adding a stack var to hold the return code, or a `switch` statement, which ends up taking a few extra lines. Which would you prefer?

Going from:

```c
if (psutil_pid_exists(pid)) {
  // handle pid existing
} else {
  // handle pid not existing
}
```

to either:

```c
int exists;

...

exists = psutil_pid_exists(pid);
if (exists == 0) {
  // handle pid not existing
} else if (exists == 1) {
  // handle pid existing
} else {
  // handle error
}
```

or:

```c
switch (psutil_pid_exists(pid)) {
case 0:
  // handle pid not existing
  break;
case 1:
  // handle pid existing
  break;
default:
  // handle error
}
```

giampaolo:
I fixed this in https://github.com/giampaolo/psutil/pull/908

PR #859, opened 2016-07-11
Extend open_files

tijko:
This allows for tracking of anonymous inodes and also adds a field for `mnt_id` inside of `popenfile`.  

> mnt_id This field, present since Linux 3.15, is the ID of the
> mount point containing this file.  See the description
> of /proc/[pid]/mountinfo

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -159,9 +159,7 @@ sdiskio = namedtuple(
                 'read_time', 'write_time',
                 'read_merged_count', 'write_merged_count',
                 'busy_time'])
-# mnt_id? check version or just add it for 'popenfile'?
-# use 'popenfile' and combine with additional attributes for other anon_inodes?
-# instead of creating numerous namedtuples with shared fields.
+
 popenfile = namedtuple('popenfile', ['path', 'fd', 'position', 'mode', 
     'flags', 'mnt_id', 'anon_fields'])
 
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -159,8 +159,16 @@ sdiskio = namedtuple(
                 'read_time', 'write_time',
                 'read_merged_count', 'write_merged_count',
                 'busy_time'])
-popenfile = namedtuple(
-    'popenfile', ['path', 'fd', 'position', 'mode', 'flags'])
+# mnt_id? check version or just add it for 'popenfile'?
+# use 'popenfile' and combine with additional attributes for other anon_inodes?
+# instead of creating numerous namedtuples with shared fields.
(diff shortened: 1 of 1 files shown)


psutil/_pslinux.py
```diff
@@ -159,8 +159,14 @@ class IOPriority(enum.IntEnum):
                 'read_time', 'write_time',
                 'read_merged_count', 'write_merged_count',
                 'busy_time'])
-popenfile = namedtuple(
-    'popenfile', ['path', 'fd', 'position', 'mode', 'flags'])
+
+popenfile = namedtuple('popenfile', ['path', 'fd', 'position', 'mode', 
+    'flags', 'mnt_id', 'anon_fields'])
+
```
tijko: Added the `mnt_id` field and a `anon_fields` which defaults to an empty `dict` if a regular file.

psutil/_pslinux.py
```diff
@@ -159,8 +159,14 @@ class IOPriority(enum.IntEnum):
...
+ppoll_tfd = namedtuple('ppoll_tfd', ['events', 'data'])
+pinotify = namedtuple('pinotify_attrs', ['ino', 'sdev', 'mask', 'ignored_mask', 
+     'fhandle_bytes', 'fhandle_type', 'f_handle'])
+
```
tijko: Two new `namedtuple`s for `ppoll` and `inotify` descriptors.  Using the fd's as the keys and these `namedtuple`s as the values.  The other anon inodes `signalfd` and `eventfd`, don't contain multiple fields and can use just their event masks as the values.

giampaolo:
What's the use case for this? What's an anonymous file? What's a `mnt_id`? You're introducing a difference in term of API for one platform, so it must be worth it.

tijko:
The anonymous files are open files that don't have a corresponding inodes (i.e. they do not have a path on the filesystem).  Examples of these would be created using `epoll_create`, `inotify_init`, `signalfd`, `eventfd`, and others.  Nevertheless, there is still very useful data that is contained in their respective `/proc/$pid/fdinfo` entries.  For instance, if a process had numerous fd's being monitored by an `epoll` fd, you can then see which events were masked into said poll.  Another use (and the one that I had been needing) would be if you have an `inotify` fd in a process and needed to see what the inode of contained watch is, you would be able to find this under the `inotify`s `ino` field.

The `mnt_id` has been exposed since [Linux 3.15](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=49d063cb353265c3af701bab215ac438ca7df36d).  Which needed when a there are two file descriptors with the same path but, are on different mount points with separate namespaces.

a-sk:
I would like to see `mountinfo` information for a given process.

giampaolo:
Closing this out as outdated.

PR #861, opened 2016-07-16
rm dupl newlines

vkukshiev:


giampaolo:
Have you checked whether "make pep8" and "make flake8" do not complain about this?

Issue #811, opened 2016-04-28
Packaged wheels/exes are not compatible with Win XP


bitranox:
can not load psutil 4.1.0 on Windows XP, SP3, Python 3.4.3 :

Microsoft Windows XP [Version 5.1.2600]
Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:43:06) [MSC v.1600 32 bit (Intel)] on win32

import psutil
Traceback (most recent call last):
File "", line 1, in 
File "C:\Python34\lib\site-packages\psutil__init__.py", line 124, in
from . import _pswindows as _psplatform
File "C:\Python34\lib\site-packages\psutil_pswindows.py", line 14, in 
from . import _psutil_windows as cext
ImportError: DLL load failed: The specified procedure could not be found.

And I can not update that particular machine, because it is a gateway for an ancient database.
Please Help !

giampaolo:
This is likely due to the old pip version you're using. `pip install -U pip` should fix this issue.

giampaolo:
Also, closing this as a duplicate of #810.

robodude666:
@giampaolo This is actually due to the fact that the psutil whls deployed to pypi are built on Windows Vista or newer (or it seems that way).

The `setup.py` script sets `_WIN32_WINNT` based on the result of `sys.getwindowsversion()` which returns the host OS version, not the target OS. Windows XP does not have the `GetTickCount64` API, which was added in Vista. This causes the above crash on `import psutil`.

Currently, the only way to get psutil to work on Windows XP is to install from source. This can be done with `pip install psutil --no-binary :all:` which will force pip to download the source and run setup.py manually.

However... that requires you to run on Windows XP. If you want to compile on Windows Vista or newer, and be able to run your python application on Windows XP you can't unless you manually modify the setup.py script before installing on your Vista or newer box.

I recommend not having the setup.py file compile based on the host OS, but instead on an extra parameter, like `pip install psutil[winxp]` or `psutil[win32]` and have a lookup table of `_WIN32_WINNT` [values](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx#macros_for_conditional_declarations) values.

-robodude666

JokerQyou:
I think @robodude666 is right. `pip install psutil` is broken on Windows XP, even with the latest pip (which is `8.1.2` currently).

JokerQyou:
**Update**

[...]

PR #874, opened 2016-08-24
Added getting IPv4 netmask in Windows 7 and above

ewedlund:
This adds support for getting the IPv4 netmask on Windows (7 and above). On older versions of Windows, it returns a netmask of value None as before. Tested on Windows 7 and 10 with Python 2.7. I have not managed to build  it on Windows XP so not tested there.

### DIFF
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -2918,7 +2918,7 @@ psutil_net_if_addrs(PyObject *self, PyObject *args) {
     char netmask_buff[100];
     DWORD netmask_bufflen = 100;
     DWORD dwRetVal = 0;
-#if (_WIN32_WINNT >= 0x0601)  // Windows 7
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista and above
     ULONG converted_netmask;
     UINT netmask_bits;
     struct in_addr in_netmask;
@@ -3007,7 +3007,7 @@ psutil_net_if_addrs(PyObject *self, PyObject *args) {
                         pUnicast->Address.lpSockaddr;
                     intRet = inet_ntop(AF_INET, &(sa_in->sin_addr), buff,
                                        bufflen);
-#if (_WIN32_WINNT >= 0x0601)  // Windows 7
+#if (_WIN32_WINNT >= 0x0600) // Windows Vista and above
                     netmask_bits = pUnicast->OnLinkPrefixLength;
                     dwRetVal = ConvertLengthToIpv4Mask(netmask_bits, &converted_netmask);
                     if (dwRetVal == NO_ERROR) {
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -2911,9 +2911,18 @@ psutil_net_if_addrs(PyObject *self, PyObject *args) {
     unsigned int i = 0;
     ULONG family;
     PCTSTR intRet;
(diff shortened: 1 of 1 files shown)


giampaolo:
This is great, thanks. I only have one complaint. With this:

```
#if (_WIN32_WINNT >= 0x0601)  // Windows 7
```

...if the exe/wheel is compiled with Windows 7 the resulting exe/wheel will not be compatible with win < 7 (e.g. Vista). On Vista, the wheel / exe will "believe" it's running on Windows 7 when it's not. As such you should use some other strategy to determine the Windows version **at runtime** (a Windows API call).

ewedlund:
Ok, I'll try to find a solution to that.

ewedlund:
Actually, reading your consideration a second time I am not sure if I understand what you mean. Even if you are compiling on let's say Windows 7 for Windows Vista, _WIN32_WINNT should have the value indicating Windows Vista, right? 

The check I used #ifdef's for is needed for two reasons:

- OnLinkPrefixLength is available in IP_ADAPTER_UNICAST_ADDRESS as from Windows Vista

- ConvertLengthToIpv4Mask is also available as from Windows Vista

(as I side note I realised actually my check was excessive, I should check for >= Vista, not Windows 7)

In order to avoid the preprocessor command, I could call GetProcAddress in order to check whether the function is available at runtime, and I agree that it is a more correct solution, but checking for the presence of the struct member OnLinkPrefixLength at runtime would be more complicated. 

So basically, before trying a complicated solution for checking for OnLinkPrefixLength at runtime, I wanted to be sure that it is really needed. I am not very familiar with how the build procedure works, but AFAIK you would not compile for Windows Vista using Windows 7 libriaries? Or am I missing something?

giampaolo:
> Actually, reading your consideration a second time I am not sure if I understand what you mean.
> Even if you are compiling on let's say Windows 7 for Windows Vista, _WIN32_WINNT should
> have the value indicating Windows Vista, right?

No, it won't.
The problem is that the `#ifdef` check is "pre-processor code": as such it is executed once at the time of compilation and never again. The resulting exe/wheel binary has no `#ifdef` in it, just the C code block which was defined in the #ifdef block at the moment of compilation. That's why if you produce the exe/wheel on Windows 7 the resulting binary will be different than on Win XP or whatever.
I only discovered this recently, see:
https://github.com/giampaolo/psutil/issues/811#issuecomment-230636680  

> as I side note I realised actually my check was excessive, I should check for >= Vista, 
> not Windows 7

Are you sure? Does this functionality require Win >= Vista?
If so, then your PR should be good because we already assumed the exes/wheel do not work on Windows XP (but compiling from sources is fine).

ewedlund:
> Are you sure? Does this functionality require Win >= Vista?

According to 
https://msdn.microsoft.com/en-us/library/windows/desktop/aa366066(v=vs.85).aspx
and
https://msdn.microsoft.com/en-us/library/windows/desktop/bb408402(v=vs.85).aspx

it is from Windows Vista or Windows server 2008 of  and later. (Windows server 2008 has the same value of _WIN32_WINNT as Vista).

>If so, then your PR should be good because we already assumed the exes/wheel do not work on Windows XP (but compiling from sources is fine).

I will fix the version and commit.

giampaolo:
Looking back at this, if the change is Windows Vista + only I think this is good for merging.

ewedlund:
That would be great!

PR #880, opened 2016-09-01
Fixes race condition getting IPv4 TCP table on Windows.

AndreLouisCaron:
We've been getting rather frequent cases where the `psutil.Process().connections()` returns an empty list on Windows when we know for a fact that there are open TCP connections on the machine.

We traced down the problem to the fact that there is a race condition in the `GetExtendedTCPTable()` API: since we call it twice (once to get the size, once to read the data), it happends that new connections open between the two calls to the `GetExtendedTCPTable()` function and the table size is too small on the 2nd call.

The solution to this is to call the function in a loop and repeat until we manage to successfully read the data.

I couldn't run the tests on my machine, so I'm sending this as a preliminary patch.   I think the other calls (e.g. for IPv6, and probably UDP) also need this too.  Maybe some refactoring could help move this logic in common somewhere.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -402,3 +402,9 @@ I: 870
 N: Yago Jesus
 W: https://github.com/YJesus
 I: 798
+
+N: Andre Caron
+C: Montreal, QC, Canada
+E: andre.l.caron@gmail.com
+W: https://github.com/AndreLouisCaron
+I: 880
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -30,6 +30,7 @@ Bug tracker at https://github.com/giampaolo/psutil/issues
 - #869: [Windows] Process.wait() may raise TimeoutExpired with wrong timeout
   unit (ms instead of sec).
 - #870: [Windows] Handle leak inside psutil_get_process_data.
+- #880: [Windows] Handle race condition inside psutil_net_connections.
 
 
 4.3.0 - 2016-06-18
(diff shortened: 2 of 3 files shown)


giampaolo:
This looks reasonable. 
Given that this may loop different times I think it makes sense to release the GIL. You can do so by using `Py_BEGIN_ALLOW_THREADS` and `Py_END_ALLOW_THREADS` around `GetExtendedTCPTable`.
Also you should do the same for `GetExtendedUDPTable`.
Also please update HISTORY and CREDITS file.

AndreLouisCaron:
OK, thanks, will update shortly :-)

AndreLouisCaron:
OK.  I updated the code to affect TCP IPv6, UDP IPv4 and UDP Ipv6 too.  I also updated `HISTORY.rst` and `CREDITS`.

Tests are not passing now, still needs a bit of work.

AndreLouisCaron:
Added the code to release the GIL, as requested :-)

AndreLouisCaron:
@giampaolo I think I addressed everything you asked.  Let me know if this PR needs more work :-)

giampaolo:
Sorry I'm busy and still didn't have time to look at this but I promise I will (likely during the week end)

giampaolo:
Merged. Thanks.

AndreLouisCaron:
Thanks for merging this!

giampaolo:
Thanks for the PR. ;)
I will make a release soon.

PR #894, opened 2016-09-27
Quote script paths when invoking them from the shell

embray:
Several tests were erroring out on my system since my home directory on Windows contains spaces (though in principle this could apply on other platforms).


embray:
Again, one test failure on AppVeyor due to #896

giampaolo:
Makes sense. Merging. Thanks.

embray:
Thanks!

PR #895, opened 2016-09-27
Fix failing test_net_if_addrs when some network interfaces are disconnected

embray:
Trying to bind a socket to the address of an interface that's down can fail, so skip in that case.

### DIFF
--- a/psutil/tests/test_system.py
+++ b/psutil/tests/test_system.py
@@ -552,6 +552,8 @@ class TestSystemAPIs(unittest.TestCase):
         nics = psutil.net_if_addrs()
         assert nics, nics
 
+        nic_stats = psutil.net_if_stats()
+
         # Not reliable on all platforms (net_if_addrs() reports more
         # interfaces).
         # self.assertEqual(sorted(nics.keys()),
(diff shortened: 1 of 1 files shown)


embray:
A couple tests failed on AppVeyor due to the issue fixed by #897

giampaolo:
Can you point me to the failure or provide the traceback? Even if the interface is down the IP addresses should be consistent. If the failure is because the address cannot be binded it's ok (so we skip that), but if the test fails because the address is not a valid IPv4 address or something then we want a failure.

embray:
Ah, to be clear it's simply that the address can't be binded.  The error this is fixing look like this:

```
ERROR: test_net_if_addrs (test_system.TestSystemAPIs)
  File "C:\Users\Erik M. Bray\src\psutil-win\psutil\tests\test_system.py", line 574, in test_net_if_addrs
    s.bind((addr.address, 0))
OSError: [WinError 10049] The requested address is not valid in its context
```

I can reproduce this very simply outside psutil as well, simply by looping over the IP addresses of the interfaces and trying to bind a socket to them.  It fails with this error only on interfaces that are down.

giampaolo:
What the address looks like? Is it valid?

embray:
They have link-local addresses.  For example:

```
'Local Area Connection* 2': [
    snic(family=<AddressFamily.AF_LINK: -1>, address='90-2E-1C-47-65-7C', netmask=None, broadcast=None, ptp=None),
    snic(family=<AddressFamily.AF_INET: 2>, address='169.254.28.236', netmask='255.255.0.0', broadcast=None, ptp=None),
    snic(family=<AddressFamily.AF_INET6: 23>, address='fe80::d9b9:5207:4c85:1cec', netmask=None, broadcast=None, ptp=None)
]
```

giampaolo:
It probably is enough to just skip the `bind()` part if the interface is down - no?

embray:
I think I see what you're saying, but if it just skipped the `bind()` all that would be left would be initializing a socket with `addr.family`.  That doesn't seem worth testing since before that point it already asserts that `addr.family` is a valid, known value.

embray:
In other words, all this PR changes is skipping the part of the test that creates a socket and binds the address to it.  It doesn't skip any other assertions.  If it skips the bind there's not much point in creating the socket to bind to.

giampaolo:
I erroneously thought we were also checking address validity but we're not. Sorry about that. Merging.

embray:
Ok, no problem. Thanks!

giampaolo:
Thank you dude. ;)

Issue #896, opened 2016-09-27
Random Windows test failures on AppVeyor


embray:
Several of the AppVeyor test runs are randomly erroring out in `get_test_subprocess` with errors like:

```
======================================================================
ERROR: setUpClass (test_windows.WindowsSpecificTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\psutil\psutil\tests\test_windows.py", line 68, in setUpClass
    cls.pid = get_test_subprocess().pid
  File "c:\projects\psutil\psutil\tests\__init__.py", line 209, in get_test_subprocess
    os.remove(TESTFN)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\projects\\psutil\\$testfile'
```

I'm just making this placeholder issues for it to say that I'm working on fixing this; will add a PR when I'm fairly confident it's fixed.

giampaolo:
Oh yeah. That's a race condition / timing issue. 
We need to first `safe_remove()` the file before running the subprocess, then retry a bunch of times in case of error == 32.
A better alternative would be to use some other synchronization mechanism other than a file, like a signal, but I'm not sure whether that can be done on Windows as historically implements signals very poorly.

embray:
Instead of `safe_remove` I'm using `wait_for_file` which I modified slightly to allow for this case as well.

giampaolo:
I suppose this is fixed by now. Closing.

PR #897, opened 2016-09-27
Fix for #896

embray:
This is my attempt to work around #896.  I've run an appveyor build on the final commit in this branch a few times now and had it pass consistently, though no 100% promises.

This mostly works by eliminating some code duplication by reusing the existing `wait_for_file` helper function, with a few small modifications to it.

### DIFF
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -203,14 +203,9 @@ def get_test_subprocess(cmd=None, **kwds):
         pyline += "sleep(60)"
         cmd = [PYTHON, "-c", pyline]
         sproc = subprocess.Popen(cmd, **kwds)
-        stop_at = time.time() + 3
-        while stop_at > time.time():
-            if os.path.exists(TESTFN):
-                os.remove(TESTFN)
-                time.sleep(0.001)
-                break
-            time.sleep(0.001)
-        else:
+        try:
+            wait_for_file(TESTFN)
+        except RuntimeError:
             warn("couldn't make sure test file was actually created")
     else:
         sproc = subprocess.Popen(cmd, **kwds)
@@ -392,7 +387,7 @@ def wait_for_file(fname, timeout=GLOBAL_TIMEOUT, delete_file=True):
             if delete_file:
                 os.remove(fname)
             return data
-        except IOError:
+        except OSError:
             time.sleep(0.001)
     raise RuntimeError(
         "timed out after %s secs (couldn't read file)" % timeout)
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -387,7 +387,7 @@ def wait_for_file(fname, timeout=GLOBAL_TIMEOUT, delete_file=True):
             if delete_file:
                 os.remove(fname)
             return data
-        except OSError:
+        except (IOError, OSError):
             time.sleep(0.001)
     raise RuntimeError(
         "timed out after %s secs (couldn't read file)" % timeout)
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -204,7 +204,7 @@ def get_test_subprocess(cmd=None, **kwds):
         cmd = [PYTHON, "-c", pyline]
         sproc = subprocess.Popen(cmd, **kwds)
         try:
-            wait_for_file(TESTFN)
+            wait_for_file(TESTFN, empty=True)
         except RuntimeError:
             warn("couldn't make sure test file was actually created")
     else:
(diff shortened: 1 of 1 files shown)


embray:
It looks like maybe the travis builds failed for unrelated reasons, but I could be wrong.

psutil/tests/__init__.py
```diff
@@ -380,20 +375,26 @@ def wait_for_pid(pid, timeout=GLOBAL_TIMEOUT):
...
+                sleep_for *= 2
                 continue
             if delete_file:
                 os.remove(fname)
             return data
-        except IOError:
-            time.sleep(0.001)
+        except (IOError, OSError):
+            time.sleep(sleep_for)
+            sleep_for *= 2
```
giampaolo: This incremental timeout is a good change. I would put a limit though so that we sleep no more than, say, 0.01 secs per loop. That will make unit tests a bit faster.
`sleep_for *= 2` should become `sleep_for = min(sleep_for * 2, 0.01)`.
This way we'll wait:

```
0.001
0.002
0.004
0.008
0.01
0.01
...
```

psutil/tests/__init__.py
```diff
@@ -380,20 +375,26 @@ def wait_for_pid(pid, timeout=GLOBAL_TIMEOUT):
...
+            if not empty and not data:
+                time.sleep(sleep_for)
+                sleep_for *= 2
                 continue
             if delete_file:
                 os.remove(fname)
             return data
-        except IOError:
-            time.sleep(0.001)
+        except (IOError, OSError):
```
giampaolo: This is too broad. We should only `pass` in case of `OSError` + `ENOENT` (meaning file is still not there) and in case of windows also take `WindowsError` + `errno == 32` (file used by another process) into account.
This way we'll get an exception if, e.g., the file cannot be opened or removed because of permission issues (we want to be aware of that).

psutil/tests/__init__.py
embray: Okay

psutil/tests/__init__.py
embray: Okay, that makes sense.

embray:
Yeesh...I'll get this right eventually :)

embray:
Ok, Travis build now failing for a seemingly unrelated reason.

psutil/tests/__init__.py
```diff
@@ -380,20 +379,36 @@ def wait_for_pid(pid, timeout=GLOBAL_TIMEOUT):
...
             if delete_file:
                 os.remove(fname)
             return data
-        except IOError:
-            time.sleep(0.001)
+        except EnvironmentError as exc:
+            posix_errno = exc.errno
+            win_errno = getattr(exc, 'winerror', None)
+
+            if (posix_errno in (errno.ENOENT, errno.ENXIO, errno.EOPNOTSUPP) or
```
giampaolo: Why are ENXIO and EOPNOTSUPP necessary?

psutil/tests/__init__.py
```diff
@@ -62,6 +62,10 @@
 else:
     import imp as importlib
 
+if sys.platform.startswith('win'):
```
giampaolo: You can simply do `if WINDOWS:` instead

psutil/tests/__init__.py
embray: Have a look at https://travis-ci.org/giampaolo/psutil/builds/163674749

psutil/tests/__init__.py
embray: ah, nice

psutil/tests/__init__.py
giampaolo: What the...? I've never seen that! Thanks. Looks like this is good for merging.

giampaolo:
https://travis-ci.org/giampaolo/psutil/jobs/163704213 looks like this broke unicode tests somehow

giampaolo:
Merged for now. I will try to fix the remaining issues as a second step.

embray:
I was seeing those failures in the last Travis build, but assumed it was unrelated to this change.  I could certainly believe it's possible they're related but I don't immediately see how.  Maybe you have some better ideas.

giampaolo:
I think I will refactor the whole thing in order to extrapolate the "wait" logic (sleep(), etc) which is currently shared in 4 different places. A `retry` decorator looks like a sensible choice to me. Incidentally I think that will mean removing most of your work/logic (sorry).

embray:
That's fine--it makes sense to extract out to a general retry method.   I had the same thought at the time but mostly wanted to focus for now on the exact area that was causing failures for continuous integration (since I want that to be fully working so I can move on to adding Cygwin builds as well).

giampaolo:
Yeah, hopefully we should stabilize all tests soon.

giampaolo:
Note: for https://travis-ci.org/giampaolo/psutil/jobs/164018394 I will provide a fix soon so don't mind about it.

giampaolo:
I added the retry decorator here: https://github.com/giampaolo/psutil/pull/901/files
Also, tests should be more stable now.

PR #940, opened 2016-11-23
Correctly calculate CPU percentages on linux with guests

Foxlik:
Since guest is already included in user and guest_nice is included in nice, it got included twice in totals.

http://unix.stackexchange.com/questions/178045/proc-stat-is-guest-counted-into-user-time

Please doublecheck this, as I can test just on recent linux.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1671,11 +1671,16 @@ def cpu_percent(interval=None, percpu=False):
         raise ValueError("interval is not positive (got %r)" % interval)
 
     def calculate(t1, t2):
-        t1_all = sum(t1)
-        t1_busy = t1_all - t1.idle
+        # We have to substract guest and guest_nice, since those are already included in user
 
+        t1_all = sum(t1)
         t2_all = sum(t2)
+        if 'guest' in t1._fields and 'guest_nice' in t1._fields and 'guest' in t2._fields and 'guest_nice' in t2._fields:
+            t1_all = t1_all - t1.guest - t1.guest_nice
+            t2_all = t2_all - t2.guest - t2.guest_nice
+        t1_busy = t1_all - t1.idle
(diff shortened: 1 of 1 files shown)


giampaolo:
Uhm... good catch. I'm not completely certain the explanation given in that SO thread is correct though. The first message talks about times in `/proc/stat` (the **system-wide** CPU times) whereas who posted the first answer talks about `/proc/[pid]/stat`, which are **process times**.

giampaolo:
I'm trying to find an official reference which states that `guest` and `guest_nice` in /proc/stats are already included in `user` time.

Foxlik:
@giampaolo Good point. I think this is enough proof: https://github.com/torvalds/linux/blob/master/kernel/sched/cputime.c#L158

And for explanation see very last paragraph of https://lists.linuxfoundation.org/pipermail/virtualization/2009-August/013459.html

giampaolo:
Thanks, that's enough. =)
After further investigation it seems we are also calculating the busy time wrong (and hence the percentage is wrong).
Right now it's done as `busy = total - idle` but according to htop (see [here](http://stackoverflow.com/questions/5514119/accurately-calculating-cpu-utilization-in-linux-using-proc-stat) and [here](https://github.com/hishamhm/htop/blob/8af4d9f453ffa2209e486418811f7652822951c6/linux/LinuxProcessList.c#L867)) it should be `busy = total - idle - iowait`.

To push this even further, I'm not sure whether we should adjust the guest time calculation only in `cpu_percent()` or also in `cpu_times()` (I mean subtracting the `guest` and `guest_nice` from `user` and `nice`).

giampaolo:
This is now correctly fixed. Thanks for discovering the original issue. I had no idea the CPU percent related math was not 100% correct.

Foxlik:
Great. Thanks!

PR #974, opened 2017-02-11
Add new sensors_fans method

nicolargo:
Add a new method to grab the fans speed.

The method is called sensors_fans():

```
>>> import psutil
>>> psutil.sensors_fans()
defaultdict(<type 'list'>, {'dell_smm': [('Processor Fan', 0)]})
```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,13 @@
 
 *2017-02-07*
 
+5.1.4
+=====
+
+**Enhancements**
+
+- 971_: [Linux] Add sensors_fans method.
+
 5.1.3
 =====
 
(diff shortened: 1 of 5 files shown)


giampaolo:
Thanks for this PR.
I'm gonna merge this but I will apply changes to the return value so that:
- we return a `dict` instead of a `defaultdict`
- `dict` value will be a namedutple

giampaolo:
OK, here's the new API:

    >>> import psutil
    >>> psutil.sensors_fans()
    {'asus': [sfan(label='cpu_fan', current=3200)]}

giampaolo:
New 5.2.0 version is out on PYPI.

PR #988, opened 2017-03-09
Simplify bytes2human() symbols and prefix

cclauss:


cclauss:
Python 2.6 did not support dict comprehensions.

giampaolo:
That's not a dict comprehension, it's a tuple. I don't think this change is necessary. Closing.

Issue #82, opened 2014-05-23
Cygwin Support


giampaolo:
_From [jlo...@gmail.com](https://code.google.com/u/104710261032439836788/) on March 20, 2010 02:17:54_

```
Posting this on behalf of a user request I received directly via email. 

From: Yves Pouplard <yves.pouplard@gmail.com>
Date: Sun, Mar 7, 2010 at 12:50 PM
Subject: psutil and cygwin ?
To: jloden@gmail.com

Hi!

I am new in Python, so I have not been able to make psutil works in Cygwin,
but it would be fine if it would :)

fyi, Cygwin is a Linux environment that runs inside Windows, and I use it a
lot -- as many other people do, probably.

can you help?

thanks.

Yves Pouplard
```

_Original issue: http://code.google.com/p/psutil/issues/detail?id=82_

embray:
Would you accept a patch to add Cygwin support?

giampaolo:
I must admit I've never used it and don't know much about it. What would it take to add support? What benefits would it bring exactly (e.g. why would you want to use psutil from cygwin)?

embray:
> I must admit I've never used it and don't know much about it.

I know a decent amount about it--have spent some time (more than I would like to admit) in its internals.

> What would it take to add support?

From my initial experiments, the easiest way to add support would be to explicitly add it as a platform.  It's too different either from Linux, or plain Windows, to try to shoehorn support for it into either the existing Windows or Linux platform-specific modules.  However, it would reuse a decent amount of code from both, which might result in (a hopefully *small*) amount of refactoring.

Fortunately not a lot of new code would need to be written.  Most methods from the Linux module work--those that don't can be made to work with existing code from the Windows module.  It's just a matter of pulling the right bits in from the right places.

Additionally, a configuration would have to be added to the AppVeyor builds that includes Cygwin.  I haven't tried using Cygwin in AppVeyor yet, but it's something I'm working on.

> What benefits would it bring exactly (e.g. why would you want to use psutil from cygwin)?

The same benefits psutil has on any other platform :)

Ok, to be less glib, my primary motivation is work on SageMath (e.g. https://trac.sagemath.org/ticket/21598).  I'm working on getting Sage working on Windows, which for some time at least will continue to require Cygwin.  Sage is in the process of switching to psutil for all process-inspection related functionality (a move I strongly supported!)  However, I did not realize cygwin support was missing, so that permanently breaks my Cygwin builds.

[...]

PR #998, opened 2017-03-23
Prototype for Cygwin support module (not ready to merge)

embray:
Here is my prototype implementation for a Cygwin support module for psutil (#82).

This has been tested on Python 2.7 and 3.4 and all the tests pass on those versions (Pythons >= 3.5 are not supported on Cygwin yet).  There are a few tests that are skipped on Cygwin--mostly for corner cases that don't work on Cygwin the way they do, say, on Linux.

The main TODO items as I see it in order of effort / importance are:

* [ ] Refactoring--there are large chunks of code copied from the Linux and Windows modules to support different features on Cygwin.  The copied code has either no, or only minor modifications.  I'm thinking of moving such code around so that it can be shared by all modules that use it (to be clear, the full Windows module does *not* work on Cygwin and shouldn't be compiled there, but bits of it do; likewise for the Linux module).
* [ ] Documentation needs to be updated to reflect the new platform.
* [ ] Fix testing on Python 3.4 for Cygwin on AppVeyor
* [ ] There are still a number of compilation warnings that should probably be cleaned up if possible.


giampaolo:
This is mirable work, thanks a lot for doing this. As for refactoring, to me it looks like _pscygwin.py has more code in common with _pslinux.py rather than _pswindows.py.
As such, what I would recommend is to (whenever possible) keep Linux and Cygwin logic in _pslinux.py and use `if CYGWIN: ...` where necessary and when the logic differs too much just put it in _pscygwin.py. Something like this:

```
# _pscygwin.py

from . import _pslinux

# same logic
pids = _pslinux.pids

# new logic
def disk_io_counters():
      ...  
```

The trick is to be cautious about code which is executed at import time in _pslinux.py. Again, you can just use `if not CYGWIN: ...`, e.g.:

```
# _pslinux.py

if LINUX:
     PAGESIZE = os.sysconf("SC_PAGE_SIZE")
```

As for the C extension module, things may be more hairy. I took a look at your _psutil_cygwin.c and it seems it was a copy & paste of _psutil_windows.c with a lot of subsequent chances. I see a lot of differences but it looks like you moved a lot of stuff around (for no good reason?) but the main bits are pretty much the same, so I can't properly comment about that and would like to hear your input about it. 
Is it feasible to use the original _psutil_windows.c and use `#ifdef __CYGWIN__` where necessary? 

Last: I would like to try this but again, I have basically zero knowledge about cygwin. Can you provide some brief instructions on how to set this up on Linux so that I can play with it?

giampaolo:
Also, one question which may sound naive, but what target of users is this for? E.g. what is your use case?
I understand that in order to use psutil on cygwin you'll have to re/compile psutil from sources.
This also poses questions on whether we should provide exe / wheels binaries specific for cygwin, and that's something I'm not sure I want to do.

embray:
One comment I will make is that a lot more of the code duplication is actually in the C module, rather than the Python module.

embray:
> The trick is to be cautious about code which is executed at import time in _pslinux.py

That's the main thing I would worry about with this, and it could be annoying and/or fragile.  I think it makes (maybe) more sense to move common functionality into a common module.  Something like a `_pslinux_common.py` or `_pslinux_utils.py`.  But maybe you prefer not to litter functionality across too many individual files.

embray:
Looks like I've still got some work to do on the tests too.  I'll come back to that.

embray:
> Is it feasible to use the original _psutil_windows.c and use #ifdef __CYGWIN__ where necessary?

Yes, I think that might be possible.  That said, there are a number of bits in the `_psutil_windows.c` module that *won't* compile under Cygwin.  That suggests an argument (similar to the one I gave above) for a `_psutil_windows_common.c` or something.

As for the functions I copied + modified, I'd have to take a look at them again (a lot of this I did over several months, so I don't remember all the changes), but I think in most cases a single version of the function (perhaps with an additional argument added, or some ifdefs) can serve both cases.

embray:
> I would like to try this but again, I have basically zero knowledge about cygwin. Can you provide some brief instructions on how to set this up so that I can play with it?

I'll get back to you on that.  I'd need to come up with a minimal set of instructions.  But for the most part there's not much to it.

embray:
> Also, one question which may sound naive, but what target of users is this for? E.g. what is your use case?
I understand that in order to use psutil on cygwin you'll have to re/compile psutil from sources.
This also poses questions on whether we should provide exe / wheels binaries specific for cygwin, and that's something I'm not sure I want to do.

This is *originally* motivated by SageMath, which recently started using `psutil`, and relies on Cygwin for Windows support.  That said, I personally just thought it would be an interesting exercise.  I don't think you need to worry about supplying wheels; I would probably make a python-psutil package for Cygwin eventually anyways.

embray:
Been busy with other things but I'm still planning to come back to this and do some more cleanup.

embray:
Sorry for leaving this open for so long; I'm going to try to work some more on this now, to finally finish up the various loose ends.

giampaolo:
No rush. I appreciate the work you're doing as I can see it's taking a lot of effort.

KSSea:
Just curious if there's been any progress with this.  I'm very interested in having psutil available under Cygwin.

psutil/_common.py
```diff
@@ -32,10 +32,14 @@
 else:
     enum = None
 
+
+PY3 = sys.version_info[0] == 3
```
joaoe: I suggest `>=` else sometime in the future Python 4 is released and no one knows why psutil is broken in cygwin.

embray: Agree; thanks for pointing that out.  In fact instead of a `PY3` variable it'd be better to use `PY2` for Python < 3.

embray:
@KSSea Yes, I actually have some updates to this that I haven't pushed yet. I think last I left off I had rebased and done some refactoring and fixed some of my TODO items, but some of the tests weren't passing, and then I just got side-tracked.

giampaolo:
It turns out that apparently there's quite some demand for cygwin support:
https://www.google.it/search?q=psutil+cygwin&oq=psutil+cygwin&aqs=chrome..69i57j69i60l2j69i61j0.1753j0j7&sourceid=chrome&ie=UTF-8

embray:
Ugh, rebasing this on the current master was quite a mess.  It's going to take me a bit to untangle.

yselkowitz:
Any chance to finish this port?  I just came across something that requires psutil, and can certainly help get it the Cygwin distribution.

embray:
In fact one of the main issues that's been slowing this down is that discussed in #1247.  I'm working on a separate branch to kind of decouple some of the C code from Python in a way that makes refactoring easier.

That said, I know a lot of people are asking for this, so maybe if I can finish just some basic refactoring first the rest we can worry about later.  At this stage the biggest impediment is just getting caught up with master:   I started this work more than a year ago, and when I first posted this PR it had all tests passing, but not anymore.  

Incidentally I was just working on this last week.  It's not abandoned, just, low on my plate and something I poke at every now and then.  Which means it gets further behind master, and so on.... :)

giampaolo:
I may be wrong but it appears to me the main and first goal here should be avoiding to duplicate code from `_psutil_windows.c`, aka do not have `_psutil_cygwin.c` at all. IMO, `setup.py` should just do this (aka compile `_psutil_windows.c` and `_psutil_posix.c`):

```diff
diff --git a/setup.py b/setup.py
index fffefd2..0d55eb6 100755
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 sys.path.insert(0, os.path.join(HERE, "psutil"))
 
 from _common import BSD  # NOQA
+from _common import CYGWIN  # NOQA
 from _common import FREEBSD  # NOQA
 from _common import LINUX  # NOQA
 from _common import NETBSD  # NOQA
@@ -103,19 +104,32 @@ def silenced_output(stream_name):
         setattr(sys, stream_name, orig)
 
 
-if WINDOWS:
+if WINDOWS or CYGWIN:
     def get_winver():
-        maj, min = sys.getwindowsversion()[0:2]
-        return '0x0%s' % ((maj * 100) + min)
-
-    if sys.getwindowsversion()[0] < 6:
+        if WINDOWS:
+            maj, min = sys.getwindowsversion()[0:2]
+            return '0x0%s' % ((maj * 100) + min)
+        else:
+            import re
+            winver_re = re.compile(r'CYGWIN_NT-(?P<major>\d+)\.(?P<minor>\d+)')
+            verstr = os.uname()[0]
+            m = winver_re.search(verstr)
+            maj = int(m.group('major'))
+            min = int(m.group('minor'))
+            return '0x0%s' % ((maj * 100) + min)
+
+    if WINDOWS and sys.getwindowsversion()[0] < 6:
         msg = "warning: Windows versions < Vista are no longer supported or "
         msg = "maintained; latest official supported version is psutil 3.4.2; "
         msg += "psutil may still be installed from sources if you have "
         msg += "Visual Studio and may also (kind of) work though"
         warnings.warn(msg, UserWarning)
 
-    macros.append(("PSUTIL_WINDOWS", 1))
+    if WINDOWS:
+        macros.append(("PSUTIL_WINDOWS", 1))
+    else:
+        macros.append(("PSUTIL_CYGWIN", 1))
+
     macros.extend([
         # be nice to mingw, see:
         # http://www.mingw.org/wiki/Use_more_recent_defined_functions
```

How feasible would it be to restart from scratch and just compile `_psutil_windows.c` and `_psutil_posix.c`? With that in place I have a pretty clear idea on how to refactor the py code in order to reuse `_pslinux.py`.

embray:
I'm working on refactoring `_psutil_cygwin.c` so that there is less duplication.  However, it does not make sense to compile `_psutil_window.c` on Cygwin.  I don't think it will even build without significant `#ifdef`s, and even then much of the module is not applicable to Cygwin.

On the other hand, there is some Cygwin-specific code in `_psutil_cygwin.c` that really wouldn't make sense to have anywhere else, so it doesn't make sense to get rid of it entirely, with one exception: I have since released a package called pycygwin which includes basically the same Cygwin-specific functionality that I added in `_psutil_cygwin.c`.  If I could make that cygwin-only dependency of psutil then that code could be reused.  Otherwise it makes sense to copy it.  I should note that for pycygwin I used Cython, whereas the code in `_psutil_cygwin.c` of course just uses the C-API directly.

embray:
I think the biggest problem with this PR, and @giampaolo I think you'll agree, is that it was too ambitious.  I tried to get every, or almost every interface working just to see if I could.

But for the purposes of getting something that at least partially works as a first pass, I think I might start over again (although still reuse a lot) with my sights set a bit smaller.

The thing about Cygwin is that it is supposed to be *mostly* POSIX compatible.  There are a few POSIX APIs that are not implemented (except as stubs), and a few narrow areas where it is necessarily *not* POSIX compatible.  There are also a few non-POSIX interfaces borrowed from Linux.  So I believe that a lot of ground can be covered by just a little bit of refactoring to `_pslinux.py` (I would like to add something like a `_pslinux_common.py` that contains bits that can be used by other Linux-like platforms; alternatively it could just go in the existing `_common.py`).  Given that, I can add a simpler `_pscygwin.py` that only supports what is possible using bits from the posix and linux modules.

In general one should *not* use the Windows APIs on Cygwin; the whole point is that it's supposed to abstract away the Windows API.  But of course you *can* do it, and as my prototype demonstrates you can still add a lot of functionality if you dig in.    In some small cases it might still be unavoidable.  Though I also think a lot of code from psutil, as long as the license allows it, could form the basis for a new features in Cygwin to implement a lot of currently missing interfaces (I'm especially thinking `/proc/net` in this case).

giampaolo:
So, here's some comments in light of the recent discussion occurred in #1342. As I said in there the main goal here should be to avoid code duplication. Since you're going to restart from scratch I propose to proceed step by step. The goal of the first step IMO should be solving the import time issues and get to a point where `import psutil` just works. That will help us have a clearer view on how to proceed next. As part of the first step I suggest to leave the C alone and just focus on the python/import part. I suggest the following:

1) define a `psutil/_pscygwin.py` which will include the Python logic that differs from Linux (no Linux logic should be copied in here)
2) on Cygwin `psutil/__init__.py` will import `psutil/_pslinux.py`: 

```python
# __init__.py
if CYGWIN:
    from . import _pslinux as _psplatform
```

3) in `psutil/_pslinux.py` you can work with `if CYGWIN: ...` in order to exclude linux-specific logic which is loaded on import. More or less:

```python
# _pslinux.py

if CYGWIN:
    import _pscygwin
else:
    from . import _psutil_linux as cext

...

HAS_PRLIMIT = not CYGWIN and hasattr(cext, "linux_prlimit")
if not CYGWIN:
    try:
        set_scputimes_ntuple("/proc")
    except Exception:
        # Don't want to crash at import time.
        traceback.print_exc()
        scputimes = namedtuple('scputimes', 'user system idle')(0.0, 0.0, 0.0)

...

# at the end of the module, replace the APIs which differ
class Process:   # <-- the linux process
    ...

if CYGWIN:
    class Process(_pscygwin.Process):
        pass

virtual_memory = _pscygwin.virtual_memory
net_io_counters = _pscygwin.net_io_counters
...
```

* `psutil/_pscygwin.py` will more or less look like this

```python
import _psutil_cygwin as cext

def virtual_memory:
    ...

def net_io_counters():
    ...

class Process:
    def threads(self):
        return cext.proc_threads(self.pid)
...
```

The impact on `_pslinux.py` is affordable because you'll only change a couple of things which are loaded at import time. The rest will remain unchanged and clearly defined as it will keep living in the body functions/methods. These parts will never be invoked on Cygwin because you'll overwrite `_pslinux.py` namespace by injecting the definitions coming from `_pscygwin.py` at runtime. 

This IMO is a reasonable approach as long as we're not clear how the different bits should be moved around and refactored. 

Once you're done with that the next logical step will be identifying what C functions from `_psutil_windows.c`. That is more sensitive and less clear to me. We may proceed by reusing `_psutil_windows.c` and add `#ifdef`s all around, or we may have a separate `_psutil_cygwin.c` file. But that's not clear at this point (to me at least) so let's leave it for later and let's set the milestone for step 1 only.

@embray hope this helps and let me know if the above makes sense to you.

giampaolo:
Actually, on a second though I think it's better to do the other way around: to have `_pscygwin.py` import `_pslinux.py`:

`psutil/__init__.py`:

```python
if CYGWIN: 
    from . import _pscygwin as _psplatform
```

`psutil/_pslinux.py` (just solve the import time issues):

```python
if not CYGWIN:
    from . import _psutil_linux as cext

HAS_PRLIMIT = not CYGWIN and hasattr(cext, "linux_prlimit")
if not CYGWIN:
    try:
        set_scputimes_ntuple("/proc")
    except Exception:
        # Don't want to crash at import time.
        traceback.print_exc()
        scputimes = namedtuple('scputimes', 'user system idle')(0.0, 0.0, 0.0)
```

`psutil/_pscygwin.py` (redefine what's needed):

```python
from . import _pslinux
from . import _psutil_posix as cext_posix

net_if_addrs = cext_posix.net_if_addrs
disk_usage = _psposix.disk_usage

class Process(_pslinux.Process):

    def threads(self):
        return cext.proc_threads(self.pid)
```

embray:
@giampaolo I'm sorry you took the time to write all this, because I'm way ahead of you on most of that.  I have a different approach that I think you'll appreciate that was designed to avoid littering the code with `if (not) CYGWIN` blocks everywhere, which ends up looking quite a bit cleaner, for now.

I don't believe it's good to use `_pslinux.py` at all in the Cygwin implementation as there are too many differences between the platforms (Cygwin borrows from Linux in some areas that it likes, such as the layout of the `/proc` filesystem, but it is *not* Linux).  You'll just wind up with tons of the code in `_pslinux.py` bracketed by `if not CYGWIN`.  No good.

There is still a *tiny* bit of duplication for now, but not much, and it can be avoided with some other refactoring ideas I have.

I'm just making sure that the applicable tests pass, and that all other tests for interfaces that are not implemented yet are skipped as appropriate. Please in the meantime don't give it further thought until I post some actual code.

giampaolo:
You may have misunderstood what I have in mind. What I propose only adds a couple of `if (not) CYGWIN` in `_pslinux.py` in order to avoid crashing at import time, but all the rest is going to live in `_pscygwin.py` and reused via aliasing or subclassing. As such I don't aim at littering `_pslinux.py` body functions with `if (not) CYGWIN` (I for one would be against it). Judging from what I see in the current PR there are different things in common with Linux:

```
cpu_count_logical()
cpu_count_physical()
disk_partitions()
pids()
Process.name()
Process.terminal()
Process.cpu_times()
Process.wait()
Process.cpu_affinity()
Process.ppid()
```
I am not sure what you have in mind but FYI I am not keen on moving those parts out of `_pslinux.py` just for reusing them in `_pscygwin.py`. Linux is the major platform here (and in psutil in general), so it should not be altered that much just to integrate Cygwin, which really is a minor one by comparison.

giampaolo:
@embray after re reading your diff I noticed there are some tiny differences. I missed that on my first run looking back at this PR. I read your last message without giving it the proper attention and I was also a bit harsh in dismissing it. I am sorry about that.

embray:
No problem.  I'm still working on getting the tests in a sane state, but to spare some suspense what I'm doing is I created a `_pslinux_common.py` module, moved some of the common functionality there (there's not as much yet as you might assume), and both `_pscygwin.py` and `_pslinux.py` use it.

giampaolo:
Thanks for the update @embray. OK, this time I will shut up and wait to see some code and then we'll elaborate on that. FWIW, I Googled a bit and it seems Cygwin is not so "minor" as I thought. It appears there's quite some demand for it so I'm all for getting this in as a major feature. Should we close this PR?

embray:
Still making progress on this.  I've had various other fires to put out but I'm close to having the new prototype ready.  It's been mostly a progress of adding `skipIfs` to the right tests, as my current approach is to support just bare minimum functionality on Cygwin to start with so that we at least have a stub on which to add additional functionality (much of which can still be preserved from this PR, but added a bit at a time).

giampaolo:
Closing this out as outdated.

PR #1003, opened 2017-03-28
remove use of MA_RESERVED1 from SunOS module

### LINKED ISSUE #1002: Use of MA_RESERVED1 breaks on recent Solaris builds
This is against internal-to-Oracle Solaris, but recently `MA_RESERVED1` was renamed to `MA_CORE_NODATA`, as that value was given a specific use.  Currently, psutil translates that flag to `*`, but I can't find any documentation on what that's supposed to mean.  The documentation for the perms field in /proc/pid/maps in proc(5) only talks about `r`, `w`, `x`, `s`, and `p`, and not `R` and `*` as are also used in `_psutil_sunos.c`.  Since `MA_RESERVED1` was never used for anything (at least, not in recent memory), it's not clear that the `*` meant anything either, and it probably should just be dropped.  I can whip up a pull request for that, if desired.

dhduvall:
MA_RESERVED1 never meant anything, and the macro is going away in the
next release of Solaris.  MA_NORESERVE wasn't really mapped to anything
useful, either.  Removing the use of both macros shouldn't make any
material difference, and will be compatible across more versions of
Solaris.

Fixes #1002.

### DIFF
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -713,9 +713,7 @@ psutil_proc_memory_maps(PyObject *self, PyObject *args) {
         sprintf(perms, "%c%c%c%c%c%c", p->pr_mflags & MA_READ ? 'r' : '-',
                 p->pr_mflags & MA_WRITE ? 'w' : '-',
                 p->pr_mflags & MA_EXEC ? 'x' : '-',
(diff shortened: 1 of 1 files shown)


giampaolo:
Thanks.

Issue #1022, opened 2017-04-25
include pid entry into psutil.users()


alexanha:
Hi,

I need to have another entry ``'pid'=...`` in the output of ``psutil.users()``. As a proposal I uploaded a patch to [users_pid.txt](https://github.com/giampaolo/psutil/files/956232/users_pid.txt) with a solution I tested under Linux.

Cheers,
Alex

giampaolo:
Oh nice, this looks useful. Patch does not apply cleanly though. 
Could you either provide an updated patch or a PR?

PR #1023, opened 2017-04-25
make users() include pid into suser tuple

alexanha:
cf. #1022

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -441,3 +441,8 @@ I: 872
 N: Danek Duvall
 W: https://github.com/dhduvall
 I: 1002
+
+N: Alexander Hasselhuhn
+C: Germany
+W: https://github.com/alexanha
+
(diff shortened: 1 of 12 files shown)


giampaolo:
For consistency with the returned Python named tuple could you please always put the PID as the last element of the returned tuple in the C extension?

giampaolo:
...and while you're at it also update HISTORY and CREDITS files.

giampaolo:
Looks good, thanks. Merging.
Just one last thing. How should I describe this field in the doc? What is the PID referring to?

alexanha:
Awsome, thanks.
Concerning the pid: Formally speaking, since ``psutil.users()`` provides an interface to ``utmp`` it is the content of the ``ut_pid`` entry of each login entry in ``utmp`` (cf. ``man utmp``). Practically speaking it is the process id of the login process (like sshd, tmux, gdm-session-worker,...).

When can I expect this change to be released?

giampaolo:
OK, thanks. I updated doc in accordance.
I expect to make a new release pretty soon (days).

PR #1024, opened 2017-04-26
small bugfix in psutil_users

alexanha:
Sorry, made a little mistake yesterday when adapting the position of the pid in ``psutil_users`` (strange that the CI didn't find it).


giampaolo:
Yeah I noticed and fixed that already some days ago, that's why I'm not sure why your PR does not crash with the changes I introduced (which are the same).

alexanha:
strange; I could imagine, that this is no real conflict for git (which logically makes sense), but somehow git doesn't notice the tautology. Nevertheless, since the issues are fixed, I think we don't have to think about this anymore.

Issue #1034, opened 2017-04-29
Issues for mediocres


himanshub16:
I realise this project requires considerable knowledge and expertise.
Still, I like it and want to contribute to it. Are there any simple issues/guide where I can start from?

himanshub16:
I would recommend to label issues as 'bitesize', or as per difficulty level.

giampaolo:
Off the top of my head: I need to have a script that checks for broken links in the doc (this file: https://raw.githubusercontent.com/giampaolo/psutil/master/docs/index.rst). More or less it should:
* collect all urls by using a regular expression
* for each url make an HTTP request and fail it's a broken link (return code != 200 plus some other conditions I am probably missing right now).
The script should live under /scripts/internals and should be manually invoked from the Makefile every as `make check-broken-links`).
Not sure if this is something which may interest you but since I'm bad with regular expressions that would be useful to me.

Another thing which would be cool is https://github.com/giampaolo/psutil/issues/900 

Also you may want to take a look at this file:
https://github.com/giampaolo/psutil/blob/master/IDEAS
...although these are not easy tasks.

I will comment again if something else comes to mind.

giampaolo:
Closing this out. Just ping me when you're done with exams.

himanshub16:
:smile: I'm back for thrilling weekends.

giampaolo:
Hey dude. Sorry but I'm flying to Amsterdam tomorrow and I'll be back on Friday so be patient. I had something for you but I forgot. I'll search for inspiration by watching Amsterdam canals and let you know when I'm back. =)

PR #1036, opened 2017-04-29
Check broken links

himanshub16:
I've created a script to check for broken links as mentioned in #1034 
**Features of the [script](https://github.com/himanshub16/psutil/blob/check-broken-links/scripts/internal/check_broken_links.py)**:
* Takes file paths as command-line argument (absolute path).
* Uses HTTP HEAD request to minimize bandwidth requirement.
* Depends on [requests module](http://docs.python-requests.org/en/master/). Perhaps, it is now part of the Python standard library, as was available on Python2 on Ubuntu 17.04 (I haven't installed it).
  If not, then I would recommend keeping **requests** as a dependency, as it handles many ugly details as handling http, https, separately, handling redirects, status codes if one tries **urllib2** or **httplib**.
  Otherwise, the code is modular to support changing of background logic easily.

To use:
```
make check-broken-links
```
The list of files to be checked should be available in *Makefile* as [used here](https://github.com/himanshub16/psutil/blob/check-broken-links/Makefile#L278).

> There is much room for improvement as the change in choice of regex for finding links will change results. The proper [references](https://github.com/himanshub16/psutil/blob/check-broken-links/scripts/internal/check_broken_links.py#L25) used are mentioned for future reference.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,8 +275,7 @@ doc:
 	@echo "done; now manually upload doc.zip from here: https://pypi.python.org/pypi?:action=pkg_edit&name=psutil"
 
 # check whether the links mentioned in some files are valid.
-FILES_TO_CHECK_FOR_BROKEN_LINKS = $(PWD)/docs/index.rst \
-																	$(PWD)/DEVGUIDE.rst
-
 check-broken-links:
-	$(PYTHON) scripts/internal/check_broken_links.py $(FILES_TO_CHECK_FOR_BROKEN_LINKS)
+		git ls-files | grep \\.rst$ | xargs $(PYTHON) scripts/internal/check_broken_links.py
+# Alternate method, DOCFILES need to be defined
+# 	$(PYTHON) scripts/internal/check_broken_links.py $(DOCFILES)
--- a/Makefile
+++ b/Makefile
@@ -278,5 +278,3 @@ doc:
 # check whether the links mentioned in some files are valid.
 check-broken-links:
 		git ls-files | grep \\.rst$ | xargs $(PYTHON) scripts/internal/check_broken_links.py
-# Alternate method, DOCFILES need to be defined
-# 	$(PYTHON) scripts/internal/check_broken_links.py $(DOCFILES)
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ DEPS = \
 	setuptools \
 	sphinx \
 	twine \
-	unittest2
+	unittest2 \
+	requests
 
 # In not in a virtualenv, add --user options for install commands.
 INSTALL_OPTS = `$(PYTHON) -c "import sys; print('' if hasattr(sys, 'real_prefix') else '--user')"`
--- a/Makefile
+++ b/Makefile
@@ -273,3 +273,10 @@ doc:
 	cd docs && make html && cd _build/html/ && zip doc.zip -r .
 	mv docs/_build/html/doc.zip .
 	@echo "done; now manually upload doc.zip from here: https://pypi.python.org/pypi?:action=pkg_edit&name=psutil"
+
+# check whether the links mentioned in some files are valid.
+FILES_TO_CHECK_FOR_BROKEN_LINKS = $(PWD)/docs/index.rst \
+																	$(PWD)/DEVGUIDE.rst
+
+check-broken-links:
+	$(PYTHON) scripts/internal/check_broken_links.py $(FILES_TO_CHECK_FOR_BROKEN_LINKS)
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -47,8 +47,8 @@ import requests
 
 HERE = os.path.abspath(os.path.dirname(__file__))
 
-REGEX = r'(?:http|ftp|https)?://'
-REGEX += r'(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
+REGEX = r'(?:http|ftp|https)?://' \
+        r'(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
 
 
 def get_urls(filename):
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -57,7 +57,6 @@ def get_urls(filename):
     # fname = os.path.abspath(os.path.join(HERE, filename))
     # expecting absolute path
     fname = os.path.abspath(filename)
-    print(fname)
     text = ''
     with open(fname) as f:
         text = f.read()
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -20,6 +20,7 @@ Method:
 * Remove duplicates (because regex is not 100% efficient as of now).
 * Check validity of URL, using HEAD request. (HEAD to save bandwidth)
   Uses requests module for others are painful to use. REFERENCES[9]
+  Handles redirects, http, https, ftp as well.
 
 REFERENCES:
 Using [1] with some modificatons for including ftp
@@ -44,7 +45,6 @@ import requests
 
 
 HERE = os.path.abspath(os.path.dirname(__file__))
-print (HERE)
 
 URL_REGEX = '(?:http|ftp|https)?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
 
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -145,6 +145,7 @@ def main():
         for fail in fails:
             print(fail[1] + ' : ' + fail[0] + os.linesep)
         print('-' * 20)
+        sys.exit(1)
 
 
 if __name__ == '__main__':
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -65,19 +65,15 @@ def get_urls(filename):
     """
     # fname = os.path.abspath(os.path.join(HERE, filename))
     # expecting absolute path
-    fname = os.path.abspath(filename)
-    text = ''
-    with open(fname) as f:
-        text = f.read()
+    with open(filename) as fs:
+        text = fs.read()
 
     urls = re.findall(REGEX, text)
     # remove duplicates, list for sets are not iterable
     urls = list(set(urls))
     # correct urls which are between < and/or >
-    i = 0
-    while i < len(urls):
-        urls[i] = re.sub("[\*<>\(\)\)]", '', urls[i])
-        i += 1
+    for i, url in enumerate(urls):
+        urls[i] = re.sub("[\*<>\(\)\)]", '', url)
 
     return urls
 
@@ -140,7 +136,7 @@ def main():
             all_urls.append((fname, url))
 
     fails = parallel_validator(all_urls)
-    if len(fails) == 0:
+    if not fails:
         print("all links are valid. cheers!")
     else:
         print("total :", len(fails), "fails!")
--- a/scripts/internal/check_broken_links.py
+++ b/scripts/internal/check_broken_links.py
@@ -43,9 +43,8 @@ from __future__ import print_function
 import os
 import re
 import sys
-import time
 
-from concurrent.futures import ThreadPoolExecutor
+import concurrent.futures
 import requests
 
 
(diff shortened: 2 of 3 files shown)


giampaolo:
Wow this was fast, thanks a lot. I will take a look at this in a while and comment. In the meantime, since it seems you're eager to contribute, you may want to take at:
https://github.com/giampaolo/psutil/blob/master/DEVGUIDE.rst
In details, you want to `make setup-dev-env` first so that GIT hooks are installed and you will respect coding style (otherwise the continuous test integration fails).

Thanks again. More later.

Makefile
```diff
@@ -273,3 +273,10 @@ doc:
 	cd docs && make html && cd _build/html/ && zip doc.zip -r .
 	mv docs/_build/html/doc.zip .
 	@echo "done; now manually upload doc.zip from here: https://pypi.python.org/pypi?:action=pkg_edit&name=psutil"
+
+# check whether the links mentioned in some files are valid.
+FILES_TO_CHECK_FOR_BROKEN_LINKS = $(PWD)/docs/index.rst \
+																	$(PWD)/DEVGUIDE.rst
```
giampaolo: Please move this at the top of the file alongside other variables.
Also it would be better to use a shorter name (DOCFILES) and avoid to use the `$(PWD)` trick.

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
+
+# Author : Himanshu Shekhar < https://github.com/himanshub16 > (2017)
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
```
giampaolo: Please copy the license you see in other (all) py files.
Also, this file should be executable (chmod +x file)

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+[9] http://docs.python-requests.org/
+
+"""
+
+from __future__ import print_function
+
+import os
+import re
+import sys
+import requests
```
giampaolo: add an extra line to differentiate stdlib imports vs. third party libs imports

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+
+REGEX = r'(?:http|ftp|https)?://'
+REGEX += r'(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
+
```
giampaolo: you can do:

```
REGEX = r'(?:http|ftp|https)?://' + \
    r'(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
```

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+
+def get_urls(filename):
+    """Extracts all URLs available in specified filename
+    """
+    # fname = os.path.abspath(os.path.join(HERE, filename))
+    # expecting absolute path
+    fname = os.path.abspath(filename)
+    print(fname)
```
giampaolo: This is a utility function and as such it'd be better not to print anything from in here (the caller should). Usually prints should occur in main()

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+        text = f.read()
+
+    urls = re.findall(REGEX, text)
+    # remove duplicates, list for sets are not iterable
+    urls = list(set(urls))
+    # correct urls which are between < and/or >
+    i = 0
+    while i < len(urls):
+        urls[i] = re.sub("[<>]", '', urls[i])
+        i += 1
```
giampaolo: looks like this `i` variable is not used

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+        res = requests.head(url)
+        return res.ok
+    except requests.exceptions.RequestException:
+        return False
+
+
+def main():
+    """Main function
+    """
+    files = sys.argv[1:]
```
giampaolo: it'd be nice if this script would accept multiple files as in "script.py file1 file2" etc

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+    fails = []
+    for fname in files:
+        urls = get_urls(fname)
+        i = 0
```
giampaolo: this variable is not used

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+        last = len(urls)
+        for url in urls:
+            i += 1
+            if not validate_url(url):
+                fails.append((url, fname))
+            sys.stdout.write("\r " +
+                             fname + " : " + str(i) + " / " + str(last))
+            sys.stdout.flush()
```
giampaolo: why don't you just use `print()`?

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+
+    print()
+    if len(fails) == 0:
+        print("All URLs are valid. Cheers!")
```
giampaolo: just a matter of personal taste: I don't like capitalized messages ("All")

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,119 @@
...
+    else:
+        print("Total :", len(fails), "fails!")
+        print("Writing failed urls to fails.txt")
+        with open("../../fails.txt", 'w') as f:
+            for fail in fails:
+                f.write(fail[1] + ' : ' + fail[0] + os.linesep)
+            f.write('-' * 20)
+            f.write(os.linesep*2)
```
giampaolo: I prefer to print the broken urls on screen instead of writing them to a file;
also, if this script fails the process should exit with `sys.exit(1)`

giampaolo: This script can be highly optimized by retrieving urls in parallel. You can take a look at what I did here:
https://github.com/giampaolo/psutil/blob/7200a7c06af21aeb24adf28f4186f467ceb7314c/scripts/internal/download_exes.py#L156

Makefile
giampaolo: If this script accepts multiple files as in "script.py file1 file2" the Makefile could be smarter and pass all ".rst" file in one shot. You can do this to retrieve all rst files:
```
git ls-files | grep \\.rst$
```
This should do it:

```
git ls-files | grep \\.rst$ | xargs $(PYTHON) scripts/internal/check_broken_links.py
```

giampaolo (commented):
OK, here's all my comments. Please take a look at my comment about parallel execution.

scripts/internal/check_broken_links.py
giampaolo: actually nevermind; this is what it does =)

scripts/internal/check_broken_links.py
giampaolo: also, `requests` is a new dep which should be added to the Makefile (`DEPS` variable) and in scripts/internal/winmake.py.

scripts/internal/check_broken_links.py
himanshub16: Print is buffered. So, the progress does not update usually. `sys.stdout` provides option to flush the buffer.

scripts/internal/check_broken_links.py
himanshub16: Ok. The tester would then have to redirect the output if it becomes huge.

himanshub16: Yeah. I was about to ask your preference for the same.

scripts/internal/check_broken_links.py
giampaolo: also you should return an error if no args are passed:

```
if not files:
     sys.exit("usage: %s <FILES...>" % __name__)
```

scripts/internal/check_broken_links.py
giampaolo: There's `print(s, flush=True)` but I'm not sure whether it's actually necessary. At least, I never bumped into such a problem.

scripts/internal/check_broken_links.py
himanshub16: done

scripts/internal/check_broken_links.py
himanshub16: done

Makefile
himanshub16: copied the snippet to makefile

scripts/internal/check_broken_links.py
himanshub16: done

scripts/internal/check_broken_links.py
himanshub16: done

scripts/internal/check_broken_links.py
himanshub16: changed

himanshub16: commit available for review

himanshub16:
Everything seems fine. I wonder why coveralls is failing every time. `make coverage` on my system fails on some tests which don't appear to be a related to this branch.

giampaolo:
Don't worry about coveralls. Overall the PR looks fine, I only have a couple of minor concerns but I'm going to sleep now. I will comment in 6-7 hours.

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,154 @@
...
+    total = len(urls)
+    threads = []
+
+    with ThreadPoolExecutor() as executor:
+        for url in urls:
+            fut = executor.submit(validate_url, url[1])
+            threads.append((url, fut))
+
+        # wait for threads to progress a little
+        time.sleep(2)
```
giampaolo: mmm... code smell; I suppose if this is necessary is because you're doing something wrong somehow

giampaolo (changes requested):

scripts/internal/check_broken_links.py
giampaolo: This provides a nice example on how to do exactly this:
https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example
In details:
- it uses `concurrent.futures.as_completed` which AFAIU it yields you completed futures as soon as they are completed 
- uses a dict comprehension (future_to_url) to create a `{fut: url, ...}` mapping which will let you save to use the "threads" list 
- shows that you want to catch exceptions on `fut.result()` (just print it and continue)

Maybe with this you could avoid the `time.sleep()`

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,154 @@
...
+    return urls
+
+
+def validate_url(url):
+    """Validate the URL by attempting an HTTP connection.
+    Makes an HTTP-HEAD request for each URL.
+    Uses requests module.
+    """
+    try:
+        res = requests.head(url)
```
giampaolo: please use a timeout for the HTTP request (`head(url, timeout)`)
the timeout should be defined globally (`TIMEOUT`); I guess 30 secs is fine

scripts/internal/check_broken_links.py
```diff
@@ -0,0 +1,154 @@
...
+        # some websites deny 503, like Microsoft
+        # and some send 401, like Apple, observations
+        if (not res.ok) and (res.status_code in RETRY_STATUSES):
+            res = requests.get(url)
```
giampaolo: same here (use timeout)

scripts/internal/check_broken_links.py
himanshub16: I didn't want to call fut.result(), because it blocks the thread, losing the advantage of concurrency.

giampaolo: I think in terms of concurrency / time it doesn't matter whether you wait on a blocking `result()` call as the threads will keep running in background and do their job. 

The difference is that with `concurrent.futures.as_completed` you will only get tasks which are completed (or have errored) as soon as they finish.
This makes the main thread more flexible.
Incidentally, this means you can `print()` results in a synchronized fashion from the main thread and avoid the lock I was talking about.
Take a look at this commit which I just did: https://github.com/giampaolo/psutil/commit/a1517915cf0386942d0d0fbc5b101d7ae1c4aa2a

giampaolo: ...and this is how you should handle errors: e5f082e1410d319de124092a8507e8730b9c4270

himanshub16:
Wonder why the tests failed!
I ran into an issue and was not able to solve it after trying much.
On KeyboardInterrupt, the program ends but the threads run in the background. They are a bit tough to manage. I tried using a `try`-`catch`, and random hacks on `KeyboardInterrupt` and then `sys.exit`, which didn't work.

Makefile
```diff
@@ -273,3 +274,9 @@ doc:
 	cd docs && make html && cd _build/html/ && zip doc.zip -r .
 	mv docs/_build/html/doc.zip .
 	@echo "done; now manually upload doc.zip from here: https://pypi.python.org/pypi?:action=pkg_edit&name=psutil"
+
+# check whether the links mentioned in some files are valid.
+check-broken-links:
+		git ls-files | grep \\.rst$ | xargs $(PYTHON) scripts/internal/check_broken_links.py
+# Alternate method, DOCFILES need to be defined
+# 	$(PYTHON) scripts/internal/check_broken_links.py $(DOCFILES)
```
himanshub16: done

giampaolo:
Much better! Thanks. It was a good session. =)
Merging.

giampaolo:
> On KeyboardInterrupt, the program ends but the threads run in the background. They are a bit tough to stop. I tried using a try-catch, and random hacks on KeyboardInterrupt and then sys.exit, which didn't work.

Those are the threads hanging. The only way AFAIK is to exit abruptly with `os._exit(0)`.

himanshub16:
Thanks. I had a great weekend working on it.

giampaolo:
I'm glad you enjoyed it. I also had fun reviewing it and you were very good at addressing all my remarks.
Are you interested in pushing this further? I would like to also check broken links in py and c files but those are more difficult to parse because I split them on different lines in order to respect the 80 chars per line rule. If you want I will create a ticket where I describe exactly how I would address this problem.

Other than that I also have other tasks you may enjoy.

himanshub16:
Yeah. I'd like to make contribution.
A good idea will be to open issues with difficulty levels as label so that other people can view it.
Currently, I have exams approaching. I'll be back once they are over. :smiley:

PR #1038, opened 2017-04-29
fix issues with concurrency

himanshub16:
Changes:

* **The method of running threads**: [reference](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result)
  A call to fut.result() is blocking and makes the main thread wait for the other one to complete at that point. This is equivalent to synchronous behaviour, which has been modified here. First, start all workers. Give them some time to work. Then, start counting the chickens which have hatched.

* **Number of workers**: [reference](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)
  The default value of workers count in ThreadPoolExecutor is appreciable and more than the number of cores.

### DIFF
--- a/scripts/internal/download_exes.py
+++ b/scripts/internal/download_exes.py
@@ -20,6 +20,7 @@ import os
 import requests
 import shutil
 import sys
+from time import sleep
 
 from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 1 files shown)


giampaolo:
Can you please keep working in the other PR?

himanshub16:
Ok. I've made the necessary changes to #1036 
This was inspired by the same code, as it had similar behaviour.

giampaolo:
But it's not merged yet, =)
Do you want me to comment on the other PR before merging?

giampaolo:
Closing as outdated

PR #1057, opened 2017-05-09
Marked regexp patterns as raw strings and bytes to address the invalid escape warning on Python 3.6+

ygingras:
Starting with 3.6.0, invalid escapes in string and byte literals  trigger a DepreacationWarning.  This diff marks regexp patterns where such escapes occur as "raw".  There are probably other occurrences for non-Linux platforms, but since I can't test chances for such platforms I elected to not to make the changes.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -596,7 +596,7 @@ def cpu_count_logical():
         # https://github.com/giampaolo/psutil/issues/200
         # try to parse /proc/stat as a last resort
         if num == 0:
-            search = re.compile('cpu\d')
+            search = re.compile(r'cpu\d')
             with open_text('%s/stat' % get_procfs_path()) as f:
                 for line in f:
                     line = line.split(' ')[0]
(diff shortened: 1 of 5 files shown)


giampaolo:
Hello. Invalid because the have just one slash `\`?

psutil/_pslinux.py
```diff
@@ -1690,7 +1690,7 @@ def num_ctx_switches(self, _ctxsw_re=re.compile(b'ctxt_switches:\t(\d+)')):
             return _common.pctxsw(int(ctxsw[0]), int(ctxsw[1]))
 
     @wrap_exceptions
-    def num_threads(self, _num_threads_re=re.compile(b'Threads:\t(\d+)')):
+    def num_threads(self, _num_threads_re=re.compile(br'hreads:\t(\d+)')):
```
mjpieters: This line is now missing the initial `T`

mjpieters (commented):
Otherwise LGTM.

giampaolo:
How do you see the warnings? I tried `make test PYTHON=python3.7` but it prints no warnings.

giampaolo:
Also tried `warnings.simplefilter('always')` before running tests and still no warnings.

giampaolo:
Oh, I see, I was supposed to clean all pyc files first. LGTM except https://github.com/giampaolo/psutil/pull/1057#discussion_r115668395

ygingras:
Awesome guys!  Thank you for being so fast and thorough with the review.

Issue #1013, opened 2017-04-19
net_connections is returning incorrect pid


joeisca:
On FreeBSD, running psutil.net_connections() is returning connections with incorrect pids.

I tried to match a connection, by status='ESTABLISHED', and laddr=<PORT> and was given the expected connections, but the pid on one of the processes was reported as another processes pid.

As this was tied to a process timeout program, this gave unexpected and catastrophic results, depending on which process was inserted in place of the pid that should have been reported.

giampaolo:
Uhm, this is bad. For reference, on FreeBSD global and per-process connections are retrieved differently. The part associating the PID to the connection is here:
https://github.com/giampaolo/psutil/blob/b940247f6c85c307378bfa6b9c86d33f4b97a8d9/psutil/arch/bsd/freebsd_socks.c#L190

giampaolo:
Mmmm are you sure about this? Can you provide a script to reproduce the issue?
I've just tried to run netstat.py script and the PIDs look correct:

```
vagrant@freebsd/vagrant/psutil$ python scripts/netstat.py 
Proto Local address                  Remote address                 Status        PID    Program name
tcp6  :::22                          -                              LISTEN        756    sshd
udp6  :::514                         -                              NONE          557    syslogd
tcp   0.0.0.0:22                     -                              LISTEN        756    sshd
tcp   10.0.2.15:22                   10.0.2.2:54596                 ESTABLISHED   968    sshd
udp   0.0.0.0:514                    -                              NONE          557    syslogd
```

giampaolo:
OK, I've been lucky and by accident I added a test case which exposes this issue:

```
======================================================================
FAIL: test_connections.TestConnectedSocketPairs.test_unix
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/vagrant/psutil/psutil/tests/test_connections.py", line 223, in test_unix
    self.check_socket(server, conn=server_conn)
  File "/vagrant/psutil/psutil/tests/test_connections.py", line 110, in check_socket
    compare_procsys_connections(os.getpid(), cons)
  File "/vagrant/psutil/psutil/tests/test_connections.py", line 68, in compare_procsys_connections
    assert proc_cons == sys_cons, (proc_cons, sys_cons)
[...]

PR #1070, opened 2017-05-14
Fix socket to PID translation on FreeBSD.

glebius:
This file was derived from FreeBSD usr.bin/sockstat/sockstat.c. The
logic of socket to PID translation was copied incorrectly. The hash,
that sockstat(1) utility has, is completely internal feature, it isn't
part of FreeBSD API, it is just to speed things up. So, to use this
hash one actually needs to create it: declare array of buckets, populate
it with sockets. In the freebsd_socks.c this wasn't done. This fix
doesn't create the hash, instead it removes remnants of hashing that was
there in sockstat.c. It makes code more simple, but of course slower than
original sockstat(1) in case if machine is running zillions of sockets. I
decided to go this way simply because I am low on time to invest into
psutil, and also because better first provide correct and simple
implementation and then improve it, rather than jump for complexity.

### DIFF
--- a/psutil/arch/bsd/freebsd_socks.c
+++ b/psutil/arch/bsd/freebsd_socks.c
@@ -28,7 +28,6 @@
 #include "../../_psutil_common.h"
 #include "../../_psutil_posix.h"
 
-#define HASHSIZE 1009
 // a signaler for connections without an actual status
 static int PSUTIL_CONN_NONE = 128;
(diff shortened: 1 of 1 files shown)


giampaolo:
This PR fixes the PID mismatch issue but `net_connections` does not return all connections, see:
https://github.com/giampaolo/psutil/issues/1013#issuecomment-301759972

PR #1079, opened 2017-05-18
Fixes to net_connections() on FreeBSD.

glebius:
The problem was that the fd values were all initialized to -1, making pairs of locally connected sockets equal, thus truncated when added to Python set.


giampaolo:
Amazing! I wouldn't have made it. This was a high-priority one and also a new feature was added (the fd).  Thanks a lot Gleb.

Issue #595, opened 2015-02-19
open_files() on BSD reports null paths


giampaolo:
```
======================================================================
FAIL: test_open_files (__main__.TestProcess)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_psutil.py", line 1656, in test_open_files
    self.assertIn(TESTFN, filenames)
AssertionError: '/usr/home/user/psutil/$testfile' not found in ['/usr/local/bin/python2.7', '']

======================================================================
FAIL: test_open_files2 (__main__.TestProcess)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_psutil.py", line 1684, in test_open_files2
    self.assertEqual(path, fileobj.name)
AssertionError: '' != '/usr/home/user/psutil/$testfile'
```

This comes straight from kinfo_getfile(). The open file is there (you can tell from the fd) but the path is an empty string. Also procstat utility (`procstat -f {PID}`) which internally uses kinfo_getfile() has the same problem (also when running as root). Apparently the only tool which doesn't have this problem is lsof (but the code is a mess).

giampaolo:
Test disabled as of https://github.com/giampaolo/psutil/commit/b8f444d03f42f8b32f6743de36543326833c244a

giampaolo:
Note: OpenBSD has the same problem.

giampaolo:
@glebius is this something you may wanna look at?

giampaolo:
Actually I think this may be a kernel bug. Maybe we should report this to freebsd-dev.

glebius:
test_open_files2 no longer exists. What is the modern name of this test?

glebius:
Found it. It works :)

giampaolo:
Closing out as outdated.

PR #1080, opened 2017-05-18
It just works on FreeBSD 12.0 r317434.

glebius:


giampaolo:
I have FreeBSD 12 r316750 and it doesn't but I assume they fixed it later and it's gonna remain fixed in FreeBSD 12 final. The check should be stricter though. We should inspect FreeBSD version (12) and skip the test if it's minor than that (12). Not easy to do in pure python (I'll probably use `sysctl -n kern.osrelease`).

giampaolo:
re: #595

giampaolo:
Closing this out as outdated.

PR #1084, opened 2017-05-19
(Partial) Fix https://github.com/giampaolo/psutil/issues/694

alxchk:
I'm not really sure about your exception processing, so this code needs review. Also I may forget about some memory cleanup, who knows.

This works for me for Intel 32/64 on SunOS 10, and behavior is pretty much like pargs. Also code adds support for environment block. I'm not sure about your API, so there is some kind of issue with exception handling there.

### DIFF
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -370,17 +370,25 @@ class Process(object):
 
     def oneshot_enter(self):
         self._proc_name_and_args.cache_activate()
+        self._proc_environ.cache_activate()
         self._proc_basic_info.cache_activate()
         self._proc_cred.cache_activate()
 
     def oneshot_exit(self):
         self._proc_name_and_args.cache_deactivate()
+        self._proc_environ.cache_deactivate()
         self._proc_basic_info.cache_deactivate()
         self._proc_cred.cache_deactivate()
 
     @memoize_when_activated
     def _proc_name_and_args(self):
         return cext.proc_name_and_args(self.pid, self._procfs_path)
+
+    @memoize_when_activated
+    def _proc_environ(self):
+        args = cext.proc_environ(self.pid, self._procfs_path)
+        if args:
+            return dict(args)
 
     @memoize_when_activated
     def _proc_basic_info(self):
@@ -413,6 +421,10 @@ class Process(object):
     @wrap_exceptions
     def cmdline(self):
         return self._proc_name_and_args()[1].split(' ')
+
+    @wrap_exceptions
+    def environ(self):
+        return self._proc_environ()
 
     @wrap_exceptions
     def create_time(self):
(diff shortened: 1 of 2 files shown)


giampaolo:
Hello and thanks for this. I'm glad there's a SunOS internals expert out there as SunOS it's the least tested platform in psutil. With that said, to ease up the review process and trackability I would prefer two separate PRs. Do you think that is possible? Also the coding style needs some adjustments. I'm gonna start pointing them out here and now but it's totally OK to make a PR first and then I'll review that in details.

1. better to use 4 spaces indentation all over the place

2. do:

```
if (! something) {
     ...
}
else {
    ...
}
```

...rather than:

```
if (!something) {
     ...
} else {
    ...
}
```

...or other variants.

3. declare all variables at the top of the function (instead of later)

4. 1 variable declaration per line, e.g.

bad:

```
int i, x; 
```

good:

```
int i;
int x;
```

5. always check `malloc()`, `calloc()` etc. return values and raise `PyErr_NoMemory();` if needed

6. check all function return values (e.g. for `Py_BuildValue()` etc)

7. remember to free any allocated resource before returning or raising; tend to use `goto`s in order to do that (do `goto error` in case of error and do all your cleanup in the `error` block)

alxchk:
> I'm glad there's a SunOS internals expert
Well... I see SunOS for.. 1.5 days? Something like that. Sorry about that! :)

> I would prefer two separate PRs
You mean separate for env and cmdline? 

> Also the coding style needs some adjustments.
Sure. I did that in vim in that SunOS box. Horrible experience.

giampaolo:
> Well... I see SunOS for.. 1.5 days? Something like that. Sorry about that! :)

=)

> You mean separate for env and cmdline?

Yes please

> Sure. I did that in vim in that SunOS box. Horrible experience.

I can feel your pain. I've been doing the same for many years by emulating VMs with VirtualBox, setting up the VM (vim, git etc.) and switch between windows (horrible). Then I discovered Vagrant. With it I can start the VM, SSH into it and mount the local psutil directory into the VM. This way I can edit files locally (with my configured local editor) and run tests on the VM via SSH. Like this I switch between the (local) shell and the (local) editor and that's it, which is much better.
If you want to try to do the same you can try this:
* install Vagrant and Virtualbox
* copy this file in a directory and edit directory location: https://github.com/giampaolo/sysconf/blob/master/vagrant/sunos/Vagrantfile
* run `vagrant up` and `vagrant ssh`

giampaolo:
I realized and forgot to say that SunOS is the only Vagrant platform I don't use because I can't compile psutil with it: it uses CC compiler by default, I installed GCC, but I can't manage to compile properly, so maybe what I said above is not worth the effort, at least in case of SunOS.

alxchk:
Well, I'm using sshfs now, like in old good times :)
I almost refactored code. Is it ok to make 1 PR with 2 commits? Or you prefer 2 PRs (+merge)?

giampaolo:
2 separate PRs with as many commits as you want please. For simplicity you may want to push one, wait for review + merge and then start with the next one though.

PR #1089, opened 2017-05-20
Feature - parse environment (SunOS)

alxchk:
This code tries to read initial environment block and parse it. It may fail in case argv or envp were changed.
Depends on functions introduced in previous PR.

### DIFF
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -370,17 +370,23 @@ class Process(object):
 
     def oneshot_enter(self):
         self._proc_name_and_args.cache_activate()
+        self._proc_environ.cache_activate()
         self._proc_basic_info.cache_activate()
         self._proc_cred.cache_activate()
 
     def oneshot_exit(self):
         self._proc_name_and_args.cache_deactivate()
+        self._proc_environ.cache_deactivate()
         self._proc_basic_info.cache_deactivate()
         self._proc_cred.cache_deactivate()
 
     @memoize_when_activated
     def _proc_name_and_args(self):
         return cext.proc_name_and_args(self.pid, self._procfs_path)
+
+    @memoize_when_activated
+    def _proc_environ(self):
+        return cext.proc_environ(self.pid, self._procfs_path)
 
     @memoize_when_activated
     def _proc_basic_info(self):
@@ -413,6 +419,10 @@ class Process(object):
     @wrap_exceptions
     def cmdline(self):
         return self._proc_name_and_args()[1].split(' ')
+
+    @wrap_exceptions
+    def environ(self):
+        return self._proc_environ()
 
     @wrap_exceptions
     def create_time(self):
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -345,8 +345,8 @@ read_raw_env(psinfo_t info, const char *procfs_path, ssize_t *count) {
     }
 
     if (! info.pr_envp) {
-        PyErr_SetString(
-            PyExc_RuntimeError, "Process doesn't have environment block");
+        if (count)
+            *count = 0;
 
         return NULL;
     }
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -114,6 +114,238 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
         );
 }
 
+static int
+open_address_space(pid_t pid, const char *procfs_path) {
+    int fd;
+    char proc_path[PATH_MAX];
+
+    sprintf(proc_path, "%s/%i/as", procfs_path, pid);
+    fd = open(proc_path, O_RDONLY);
+    if (fd < 0)
+        PyErr_SetFromErrno(PyExc_OSError);
+
+    return fd;
+}
+
+static int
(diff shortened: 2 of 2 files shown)


giampaolo:
Nice.

What do you mean by "**tries** to read **initial** env block"? That the result is incomplete? Do unit tests pass?
 
How exactly does this fail if  argv or envp were changed? OSError? Something else? That would be a first for Process.environ() and I'm not sure whether it's acceptable to have an API which breaks after you play with `os.environ`. On the other hand, it's OK if it just doesn't reflect changes to `os.environ` made after the process is initialized (this already happens with a bunch of other platforms and it's documented).

That aside, given the amount of new code, this probably deserves to live in a separate `psutil/arch/sunos/environ.c` file (you can do this also later, first let's clarify the two points above).

alxchk:
Original pargs tries to find _environ symbol in address space and then read it. Because these values are placed in memory of process they can be altered by process in any way. So there is zero guaranty that they are actual values (same story on Linux). I don't want to put that elf parsing mess. Also in 99% cases values are not altered. But if they are - code simply stops parsing. There's no common way to know we're they modified or not. So code looks like it

giampaolo:
So if `os.environ` is altered it just returns the old "unaltered" version?

alxchk:
There is no such thing as "unaltered". There are two addresses - one which os knows it passed to "main" and second - _environ which is known to libc. This code reads first (whatever is located in that memory). pargs reads second and if fails read first.

giampaolo:
Let me rephrase it. Is this the way it's gonna behave?

```
a = psutil.Process().environ()
os.environ['XXX'] = '1'
b = psutil.Process().environ()
assert a == b
```

alxchk:
This shouldn't work, but honestly I didn't check

giampaolo:
We should know what happens in that case. Also did you run tests (`make setup-dev-env` and `make test`)? Do they pass?

alxchk:
I just runned make test, and there are a lot of failures there. I will check them one by one.

giampaolo:
Most of them are unrelated with this. I'm just interested in knowing whether environ related tests pass or fail.

alxchk:
So, here are current results - https://gist.github.com/alxchk/2f74d3b67e9bf029266edf2468df9cdb

alxchk:
```a = psutil.Process().environ()
os.environ['XXX'] = '1'
b = psutil.Process().environ()
assert a == b```

This is correct.

So far there are several cases.
1. Processes with environment accessible by pr_envp
2. Processes with garbage accessible by pr_envp
3. Processes where pr_envp is array with 1 elements, first one is NULL
4. Processes where pr_envp is NULL (kernel processes)

For 1 we return dict, for 3 - empty dict, for 2 - exception (ValueError), 4 - exception (RuntimeError)

**Comment by giampaolo** [comment]
OK forget about the rest but this has to do with your change:

```
    FAIL: test_environ (proc=psutil.Process(pid=5, name='vmtasks'))
      File "/psutil/psutil/tests/test_contracts.py", line 310, in test_fetch_all
        ret = attr(*args, **kwargs)
      File "/psutil/psutil/__init__.py", line 884, in environ
        return self._proc.environ()
      File "/psutil/psutil/_pssunos.py", line 339, in wrapper
      File "/psutil/psutil/_pssunos.py", line 425, in environ
        return self._proc_environ()
      File "/psutil/psutil/_common.py", line 338, in wrapper
        ret = cache[fun] = fun(self)
      File "/psutil/psutil/_pssunos.py", line 389, in _proc_environ
        return cext.proc_environ(self.pid, self._procfs_path)
    ValueError: Pointer block is truncated
```

**Comment by alxchk** [comment]
> OK forget about the rest but this has to do with your change

This is the case (4). What should be there?

**Comment by giampaolo** [comment]
Does the error go away if running as root?

**Comment by alxchk** [comment]
No. There is simply no environment (not empty, but no at all)

**Comment by alxchk** [comment]
Ok, I see. In Linux it returns {}, so I'll change to the same behavior

**Comment by alxchk** [comment]
https://gist.github.com/alxchk/dbfbd139902a976e9dca8b373bbd94b5

**Comment by giampaolo** [comment]
Yes, if the file is empty it makes sense to return an empty dict.
You still have one environ-related failure.
I think you can fix it by doing this:

```
diff --git a/psutil/tests/test_process.py b/psutil/tests/test_process.py
index cab5a2f..5fbc20e 100755
--- a/psutil/tests/test_process.py
+++ b/psutil/tests/test_process.py
@@ -1404,7 +1404,7 @@ class TestProcess(unittest.TestCase):
                 "__CF_USER_TEXT_ENCODING",
                 "VERSIONER_PYTHON_PREFER_32_BIT",
                 "VERSIONER_PYTHON_VERSION"])
-        if LINUX or OSX:
+        if LINUX or OSX or SUNOS:
             removes.extend(['PLAT'])
         if TOX:
             removes.extend(['HOME'])
```

giampaolo:
Also, once that is fixed, please move that stuff into `psutil/arch/sunos/environ.c`.

giampaolo:
Actually hang on, you're fixing also the `cmdline()` stuff in here. You should implement the `environ()` only. If you know these C routines are gonna be shared with `cmdline()` then put everything in a new `psutil/arch/solaris/process_info.c` file.

PR #1091, opened 2017-05-20
Add environment parsing

alxchk:


psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
+/*
+ * Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ *
+ * Functions specific to Sun OS Solaris platforms.
+ *
+ * Thanks to Justin Venus who originally wrote a consistent part of
+ * this in Cython which I later on translated in C.
```
giampaolo: Also feel free to add your name somewhere

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+    int i;
+    off_t end;
+    size_t len;
+    char buf[STRING_SEARCH_BUF_SIZE];
+    char *result = NULL;
+
+    if (lseek(fd, offset, SEEK_SET) == (off_t)-1) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        return NULL;
+    }
```
giampaolo: here the `fd` is not closed; I suggest you wrap all this cleanup logic in an `error:` goto; I use this paradigm all over the place in psutil

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+
+    // Search end of string
+    end = offset;
+    for (;; end += STRING_SEARCH_BUF_SIZE) {
+        r = read(fd, buf, sizeof(buf));
+
+        if (r == -1) {
+            PyErr_SetFromErrno(PyExc_OSError);
+            return NULL;
```
giampaolo: same here (fd is not closed -> use goto error;)

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+        return (char *) PyErr_NoMemory();
+
+    if (len) {
+        if (read_offt(fd, offset, result, len) < 0) {
+            free(result);
+            return NULL;
+        }
+    }
+
+    result[len] = '\00';
```
giampaolo: `result` is not `free`d

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+}
+
+static char **
+read_cstrings_block(int fd, off_t offset, size_t ptr_size, size_t count) {
+    char **result = NULL;
+    char *pblock;
+    size_t pblock_size;
+    int i;
+
+    assert(ptr_size == 4 || ptr_size == 8);
```
giampaolo: what does this do?

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+    int r;
+    char buf[8];
+    static const char zeros[8] = { 0, 0, 0, 0, 0, 0, 0, 0 };
+
+    assert(ptr_size == 4 || ptr_size == 8);
+
+    if (lseek(fd, offt, SEEK_SET) == (off_t)-1) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        return -1;
+    }
```
giampaolo: `fd` is not closed (use `goto error;`)

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+        r = read(fd, buf, ptr_size);
+
+        if (r < 0) {
+            PyErr_SetFromErrno(PyExc_OSError);
+            return -1;
+        }
+
+        if (r != ptr_size) {
+            PyErr_SetString(
+				PyExc_ValueError, "Pointer block is truncated");
```
giampaolo: I'd rather throw a `RuntimeError` here

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+    if (as < 0)
+        return NULL;
+
+    ptr_size = ptr_size_by_psinfo(info);
+
+    env_count = search_pointers_vector_size_offt(
+        as, info.pr_envp, ptr_size);
+
+	if (env_count >= 0 && count)
+		*count = env_count;
```
giampaolo: indentation

giampaolo (changes requested):
OK, I made an initial review but there's a lot to adjust here. In general:

* always indent with 4 spaces
* unless the function body is very small or you don't allocate any resource (e.g. a `malloc`, an open `fd` etc.) ALWAYS use `goto;`. As a general rule, every time you want to do `return NULL` think again and do `goto error;` instead. Forgetting to free resources every time you mean to return an error is very simple, it doesn't scale and that's the door which introduces memory leaks; also, if you want to test you made no mistakes, you can use `make test-memleaks`, which is not bullet proof as it doesn't cover all your code paths but it's something
* write docstrings describing what the C function does and be explicit in stating what the possible return values are and what they mean so that the users of the function know how to deal with them

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+
+static int
+open_address_space(pid_t pid, const char *procfs_path) {
+    int fd;
+    char proc_path[PATH_MAX];
+
+    sprintf(proc_path, "%s/%i/as", procfs_path, pid);
+    fd = open(proc_path, O_RDONLY);
+    if (fd < 0)
+        PyErr_SetFromErrno(PyExc_OSError);
```
alxchk: This function just opens /as (/mem alternative) and returns fd. So I set exception here to not to lost it whatever caller did. As function should return descriptor, there is zero sense to close it.

psutil/arch/solaris/process_as_utils.c
alxchk: Fd's are closed by caller, this simplifies descriptor management

psutil/arch/solaris/process_as_utils.c
alxchk: No need to free it. If there is no data to read, then just return empty C string (len = 0, alloced size = 1, result[0] = 0 == valid zero length C string)

psutil/arch/solaris/process_as_utils.c
alxchk: Explicitly checks that I didn't pass some invalid pointer size

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+    if (! result) {
+        PyErr_NoMemory();
+        goto lbPBlockError;
+    }
+
+    for (i=0; i<count; i++)
+        result[i] = read_cstring_offt(
+            fd, (ptr_size == 4?
+                ((uint32_t *) pblock)[i]:
+                ((uint64_t *) pblock)[i]));
```
alxchk: Generally it's not big deal if something wasn't readed. This is not regular structure, but data in some others address space. So in case some pointers are invalid we still can have some valid data.

psutil/arch/solaris/process_as_utils.c
alxchk: Caller's responsibility

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
...
+        return NULL;
+    }
+
+    as = open_address_space(info.pr_pid, procfs_path);
+    if (as < 0)
+        return NULL;
+
+    result = read_cstrings_block(
+        as, info.pr_argv, ptr_size_by_psinfo(info), info.pr_argc
+    );
```
alxchk: Whatever result is (NULL or non-NULL), it's returned to caller as is.

alxchk:
Updated

psutil/_psutil_sunos.c
```diff
@@ -154,6 +156,79 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
...
+
+        Py_DECREF(py_envname);
+        Py_DECREF(py_envval);
+    }
+
+    psutil_free_cstrings_array(env, env_count);
+    return py_retdict;
+
+ enomem:
+    PyErr_NoMemory();
```
giampaolo: This does not free resources. Get rid of this and just do:

```
PyErr_NoMemory();
goto error;
```

psutil/arch/solaris/process_as_utils.c
```diff
@@ -0,0 +1,302 @@
+/*
```
giampaolo: you missed this comment

psutil/arch/solaris/process_as_utils.c
giampaolo: oh you're right

psutil/arch/solaris/process_as_utils.c
giampaolo: I think it makes sense to also close it in here, in case the caller forgets to close it. An invalid fd is unusable anyway.

psutil/arch/solaris/process_as_utils.c
giampaolo: OK but what happens? A warning is print on stderr?

psutil/arch/solaris/process_as_utils.c
giampaolo: I don't like this, also because this function fully implements error handling, but if the caller does not check the return value then what's the point? Did you see this function fail?

psutil/arch/solaris/process_as_utils.c
giampaolo: what about this?

giampaolo:
Please split this in 2 files: `utils.c` and `environ.c`.

psutil/_psutil_sunos.c
alxchk: Which resources are not freed there?

psutil/arch/solaris/process_as_utils.c
alxchk: But why? All functions here are works only with address space and likely useless anywhere else. If some other code which works with process address space will appear it can land here.

psutil/_psutil_sunos.c
giampaolo: This, which you do in the `error` body later:

```
+    if (env && env_count >= 0)
 +        psutil_free_cstrings_array(env, env_count);
 +
 +    Py_XDECREF(py_envname);
 +    Py_XDECREF(py_envval);
 +    Py_XDECREF(py_retdict);
 +    return NULL;
```

As a general rule you should only define one "error" goto block, not different gotos, and implement the whole free resource logic in there (in 1 place only). Apply this construct everywhere. It's safer and also consistent with the existent code base.

alxchk: I can do this, but I can't get this...
After landing enomem codeflow get to error. So It do the same + set NoMemory exception.

psutil/arch/solaris/process_as_utils.c
giampaolo: If all these utilities are about the process environ() then this file should be called `environ.c`. I suggested the 2 different names (`environ.c` and `utils.c`) because in the previous PR where you addressed both environ() and cmdline() it seemed that some of these functions were shared by both environ and cmdline.

As a general rule, if a function is reused by multiple functionalities, then it's better to put them in a generic module, like `utils.c` or something. Anyway, in this case it seems like this should be called `environ.c`. We can move stuff around later.

alxchk:
This is my last bump. I'm tired of rebasing and not going to do anything else. Feel free to reject or to change as you like if you don't like the code

psutil/_psutil_sunos.c
giampaolo: Because before `malloc` you may already have allocated other resources. Or maybe you didn't, but you change your code later and open a fd or use another `malloc` before this one, or create a Python list or whatever.
The "free" logic should just live in one place (`error:`) so that every time you want to raise an exception you set the error type (`PyErr_NoMemory();` or whatever) and then just do `goto error`.
This way you don't have to reason about what resources you allocated thus far and you don't have to infer whether to use `goto errmem`, `goto oserror` etc.

giampaolo:
Dude, I appreciate what you are doing here, really, and I understand your frustration, but this is how code review works. Most of the nitpicks I provided in this review session are objectively correct and well motivated. psutil has very high code quality standards, pretty much the same as when you submit a PR for the cPython project (basically this is where I trained myself and I'm coming from being I'm a core-dev myself).
...And I'm picky because once your code gets in I am likely the one who will have to maintain it in the future, not you, because you are contributing something now, but you may disappear in the future, so I care to understand what you're doing and why (hence why I ask so many questions, like the "what does `assert` do?").
It is objectively better to stick with 4 spaces indentation, always check return codes, always free resources in a single point, provide docstring and choose C file names which make sense. These are basically the only complaints I gave you thus far and there were a lot of these. Don't get me wrong, the functionality you're providing is great per-se, and I will go as far as saying I wouldn't have been able to implement it myself otherwise it would have already been there, but the fact that I had to comment 57 times on this PR shows the code you provided may have worked but it wasn't top notch. In fact it's much better now.

giampaolo:
I will merge this now and fix the other few remaining nitpicks myself. I appreciate your contribution.

giampaolo:
I get this:

```
ERROR: psutil.tests.test_process.TestProcess.test_as_dict
  File "psutil/tests/test_process.py", line 1127, in test_as_dict
    d = p.as_dict()
  File "psutil/__init__.py", line 574, in as_dict
    ret = meth()
  File "psutil/__init__.py", line 883, in environ
    return self._proc.environ()
  File "psutil/_pssunos.py", line 339, in wrapper
  File "psutil/_pssunos.py", line 419, in environ
    return cext.proc_environ(self.pid, self._procfs_path)
RuntimeError: Dereferencing procfs pointers of 64bit process is not possible

ERROR: psutil.tests.test_process.TestProcess.test_pid_0
  File "psutil/tests/test_process.py", line 1366, in test_pid_0
    ret = meth()
  File "psutil/__init__.py", line 883, in environ
    return self._proc.environ()
  File "psutil/_pssunos.py", line 339, in wrapper
  File "psutil/_pssunos.py", line 419, in environ
    return cext.proc_environ(self.pid, self._procfs_path)
RuntimeError: Dereferencing procfs pointers of 64bit process is not possible

FAIL: psutil.tests.test_process.TestProcess.test_weird_environ
  File "psutil/tests/test_process.py", line 1445, in test_weird_environ
    self.assertEqual(p.environ(), {"A": "1", "C": "3"})
AssertionError: {'A': '1'} != {'A': '1', 'C': '3'}
- {'A': '1'}
+ {'A': '1', 'C': '3'}
```

alxchk:
First two happens when you try to get environment of 64 bit process from 32 bit. It happens because psinfo_t  has different fields size for 32 and 64 bit processes. So when you read psinfo_t  structure of 64 bit process from 32 you will get truncated invalid pointers. That's the check is_ptr_dereference_possible about.

alxchk:
What 3rd test is about I can't get, because as I understand http://pubs.opengroup.org/onlinepubs/9699919799/ it's not possible to not to have '=' sign.

giampaolo:
OK, got back from my vacation so looking back at this. There is another problem: when running as a limited user I'm able to query only the current process, whereas all other processes either raises AccessDenied (which is fine) or return an empty dict (which is not).

giampaolo:
OK I think I know why this happens.

giampaolo:
OK, I should have fixed it: https://github.com/giampaolo/psutil/pull/1097

PR #1093, opened 2017-05-20
SunOS: Fix .memory_maps(grouped=False)

alxchk:
Addresses were truncated at least on LP64

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -27,6 +27,7 @@
 
 **Bug fixes**
 
+- 1093_: [SunOS] memory_maps() shows wrong 64 bit addresses
 - 1007_: [Windows] boot_time() can have a 1 sec fluctuation between calls; the
   value of the first call is now cached so that boot_time() always returns the
   same value if fluctuation is <= 1 second.
--- a/CREDITS
+++ b/CREDITS
@@ -482,4 +482,4 @@ I: 1042, 1079
 
 N: Oleksii Shevchuk
 W: https://github.com/alxchk
-I: 1077
+I: 1077, 1093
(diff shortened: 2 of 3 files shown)


giampaolo:
What's the issue here? Wrong numbers were returned?

alxchk:
Yes. 

Before:
```
python -c 'import psutil; print psutil.Process().memory_maps(grouped=False)[-1]'
pmmap_ext(addr='x17a0000-x179c000', perms='r-x-', path='/pupy/client/sources-linux/buildenv-sunos/build/lib/python2.7/lib-dynload/time.so', rss=16384, anonymous=0, locked=0)
```

giampaolo:
OK, please update HISTORY and CREDITS and it's good to go.

alxchk:
But there is no bug? What to write there? Maybe you? :)

alxchk:
So I need to issue new bug to write something to HISTORY? How this works?

giampaolo:
Just take a look at HISTORY.rst and see how it's done. Just reference this PR number and describe what changed after your patch, e.g.:

```
- 1093_: [SunOS] Process.memory_maps() was returning incorrect numbers due to inappropriate conversion of C types. (patch by YOURNAME)
``` 

In CREDITS just add your name, website, etc. by following the existing convention.

alxchk:
Done

HISTORY.rst
```diff
@@ -27,6 +27,7 @@
 
 **Bug fixes**
 
+- 1093_: [SunOS] memory_maps() shows wrong 64 bit addresses
```
giampaolo: Was there also a mismatch with the other numbers?  You change the type of all of them.

alxchk: I didn't see the mismatch with other numbers (because usually they are quite low, and there is no chances to get signed integer overflow), but semantic should be the same - these numbers can not be negative.

giampaolo: Makes sense. Thanks.

Issue #605, opened 2015-03-08
support AIX


wiggin15:
I see it's in the TODO file, filed under "lower priority", but I wanted a ticket to be open because we actually encountered this:

```
root@aixio003 (AIX 7.1.0.0) ➜  ~  easy_install psutil                                                                                                                                                                                                         15-03-08 17:46
Searching for psutil
Reading http://pypi.python.org/simple/psutil/
Best match: psutil 2.2.1
Downloading https://pypi.python.org/packages/source/p/psutil/psutil-2.2.1.tar.gz#md5=1a2b58cd9e3a53528bb6148f0c4d5244
Processing psutil-2.2.1.tar.gz
Writing /tmp/easy_install-GaoOfH/psutil-2.2.1/setup.cfg
Running psutil-2.2.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-GaoOfH/psutil-2.2.1/egg-dist-tmp-9X2hV0
error: Setup script exited with platform aix7 is not supported
```

giampaolo:
Yes, AIX is not supported. Considering that this is such a rarely used platform and that I don't have access to any AIX box it's very unlikely this is gonna get pushed forward unless someone steps in and decides to provide a patch.

wiggin15:
I started working on this here: https://github.com/Infinidat/psutil/commit/2d037eb595ecabe139b2158234641c6aff431628
There are still some missing functions, but if I open a pull request will you consider merging?

mffrench:
will be interested on that too... :)

giampaolo:
@wiggin15 what's the status of that branch of yours? What APIs are not implemented?

wiggin15:
We're using it here at my company. I think it is mostly stable. There are 10 API functions not implemented (5 of them are methods of the Process class) but I added documentation and disabled them for AIX in the latest commit. See https://github.com/Infinidat/psutil/commits/aix

(Please note that the AIX branch is also rebased on top of the branch in pull request #610 which is also not merged yet)

giampaolo:
I'm sorry but IMO 10 unsupported APIs means this is not acceptable/ready for inclusion yet. It would mean over complicating the existent code with `if`s, `unittest.skipIf(AIX)` etc., and make the doc less readable.
Please note that it is also OK to parse the output of CLI tools if the functionality is too hard to implement in C. We currently do this to provide `swap_memory` stats on Solaris, if I'm not mistaken.

giampaolo:
[...]

Issue #558, opened 2014-12-15
Allow specifying /proc and other needed folders


wojons:
The use case of this is running monitoring tools or scripts built around psutil that run inside docker continents. With docker the admin can specify that the /proc of the host machine is read only and mounted in a location on the sub continent the issue for this is that /proc room the host maybe /proc_host and then you need to read things from there for monitoring. An issue that comes up is CoreOS and a few other docker OS dont have package managers so there is no way to really install tools there its expected that the user installs the tools in a docker container.

ggueret:
+1 with the following use case : mount the host rootfs in a docker container and get some metrics from psutil, like cAdvisor "--volume=/:/rootfs:ro".

giampaolo:
OK, I'm convinced there's an use case which justifies this change. 
The problem is how to expose this feature in terms of API. 
The Linux implementation lives in the `psutil._pslinux` namespace:
https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py
...which psutil imports as `"_psplatform"`:
https://github.com/giampaolo/psutil/blob/999dc8535ab0273cdcb369b1ee31082ae632275d/psutil/__init__.py#L62

The first thing which pops into my mind is to define a `PROCFS_PATH` constant in `_pslinux.py` which the user can change like this:

```
import psutil
psutil._psplatform.PROCFS_PATH
```

I'm not too happy about this though because we are accessing a private module. I can't think of any other solution though (also renaming the module doesn't fully convince me).
Thoughts? @jloden @josiahcarlson

josiahcarlson:
I hate to say it, but what about a module property in `psutil`? Dirty, but then you aren't hiding it in a private module. Make it do the right thing on all platforms (like an exception on platforms where it doesn't exist).

Or don't even make it a property, make it a constant, and make `psutil._psplatform.*` check `psutil.PROCFS_PATH` whenever it is looking for the value, and it can just be a global like any other. You can import `psutil` in `_psplatform` functions that need it, or reference it directly from `sys.modules` based on `__name__`. Stuff the module property in `_psplatform.PROCFS_PATH`, which checks the attribute in `psutil`.

GIampaolo, you always have fun programming problems :)

P.S. At least one of those suggestions is a joke, can you guess which one(s)? Trick or Treat ;)

giampaolo:
Thanks for chiming in Josiah. =)

> I hate to say it, but what about a module property in psutil?

I knew you'd say that. =)

[...]

PR #1123, opened 2017-09-05
AIX support

wiggin15:
Following #605, this is the proposed branch for AIX support.

The following methods and methods are not supported in AIX in this branch:

    psutil.Process.memory_maps
    psutil.Process.num_ctx_switches

Not all unit tests pass on AIX at this time (I didn't make any changes to unit tests yet).


psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+The following functions and methods are UNSUPPORTED on the AIX platform:
+
+    psutil.Process.memory_maps
+    psutil.Process.num_ctx_switches
+
+
+Process.io_counters read count is always 0
+
+
+TestSystemAPIs.test_pid_exists_2           there are pids in /proc that don't really exist
```
giampaolo: You mean, say, a `/proc/54` directory exists but there's no such PID?

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+TestProcess.test_name                      isolated python calls execve which changes process name
```
giampaolo: What does this mean?

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+TestProcess.test_num_fds                   opening a socket doesn't create fd in /proc/pid/fd (until data is sent??)
```
giampaolo: Does this mean `num_fds()` is unreliable?

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+TestProcess.test_open_files                /dev/null shows in open_files but it isn't a file
```
giampaolo: Can't you just skip it?

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+TestProcess.test_pid_0                     pid 0 doesn't have a name on AIX
```
giampaolo: Does `ps` list it? If not is there some other task manager on AIX that does? If not I would advice to return `"kernel"`.

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+TestProcess.test_prog_w_funky_name         funky name test doesn't work on UNIX environments
+TestProcess.test_zombie_process            trying to create a zombie process doesn't create a zombie process on AIX
```
giampaolo: Do zombie processes exist on AIX? Is `status()` able to detect them?

psutil/TODO.aix
```diff
@@ -0,0 +1,30 @@
...
+
+TestProcess.test_memory_maps               missing API
+TestProcess.test_num_ctx_switches          missing API
+TestExampleScripts.test_pmap               missing API
+
+TestProcess.test_send_signal               unknown
+TestFetchAllProcesses.test_fetch_all       unknown
+TestProcess.test_Popen                     unknown (flaky)
+TestProcess.test_children                  unknown (flaky)
+TestProcess.test_children_recursive        unknown (flaky)
```
giampaolo: `children()` is a crucial method so I would like to see these pass. Since `children()` really is a utility method at its core, if they don't it's likely that `ppid()` implementation is broken.

psutil/__init__.py
```diff
@@ -82,6 +82,7 @@
 from ._common import POSIX  # NOQA
 from ._common import SUNOS
 from ._common import WINDOWS
+from ._common import AIX
```
giampaolo: Please sort all imports by name.

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+from collections import namedtuple
+
+from . import _common
+from . import _psposix
+from . import _psutil_posix as cext_posix
+from . import _psutil_aix as cext
+from ._common import socktype_to_enum, sockfam_to_enum
+from ._common import NIC_DUPLEX_FULL, NIC_DUPLEX_HALF, NIC_DUPLEX_UNKNOWN
+from ._common import usage_percent
+from ._compat import PY3
```
giampaolo: Please sort all imports by name. Also one import per line (avoid commas)

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+net_io_counters = cext.net_io_counters
+
+
+def virtual_memory():
+    total, free, pinned, inuse = cext.virtual_mem()
+    total = total * PAGE_SIZE
+    avail = free * PAGE_SIZE
+    used = inuse * PAGE_SIZE
+    percent = usage_percent((total - avail), total, _round=1)
+    return svmem(total, avail, percent, used, free)
```
giampaolo: This is one of the most important functions so there should be a specific test for this which tests the result against a command line utility (does AIX have `free`?)? As such there should be a `psutil/tests/test_aix.py` file.

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+    p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    stdout, stderr = p.communicate()
+    if PY3:
+        stdout, stderr = [x.decode(sys.stdout.encoding)
+                          for x in (stdout, stderr)]
+    if p.returncode != 0:
+        raise RuntimeError("%r command error\n%s" % (cmd, stderr))
+    processors = stdout.strip().splitlines()
+    return len(processors)
```
giampaolo: Looks like you want to add a test for this in `psutil/tests/test_aix.py`

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+        ntuple = _common.sdiskpart(device, mountpoint, fstype, opts)
+        retlist.append(ntuple)
+    return retlist
+
+
+def net_connections(kind, _pid=-1):
+    """Return socket connections.  If pid == -1 return system-wide
+    connections (as opposed to connections opened by one process only).
+    """
+    cmap = _common.conn_tmap.copy()
```
giampaolo: Looks like there's no need to `copy()` the dict

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+    def __init__(self, pid):
+        self.pid = pid
+        self._name = None
+        self._ppid = None
+
+    def oneshot_enter(self):
+        pass
+
+    def oneshot_exit(self):
+        pass
```
giampaolo: Looks like `cext.proc_basic_info()` returns multiple info so we can speed things up via `oneshot`().

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+            raise
+
+    @wrap_exceptions
+    def nice_set(self, value):
+        if self.pid in (2, 3):
+            # Special case PIDs: internally setpriority(3) return ESRCH
+            # (no such process), no matter what.
+            # The process actually exists though, as it has a name,
+            # creation time, etc.
+            raise AccessDenied(self.pid, self._name)
```
giampaolo: I don't like the hard coded `(2, 3)` thing. Looks like you should use `pid_exists()` instead.

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+
+    @wrap_exceptions
+    def uids(self):
+        real, effective, saved, _, _, _ = cext.proc_cred(self.pid)
+        return _common.puids(real, effective, saved)
+
+    @wrap_exceptions
+    def gids(self):
+        _, _, _, real, effective, saved = cext.proc_cred(self.pid)
+        return _common.puids(real, effective, saved)
```
giampaolo: These 2 (uids() and gids()) also look like something which can be sped up via `oneshot()`.

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+        return None
+
+    @wrap_exceptions
+    def cwd(self):
+        try:
+            return os.readlink("/proc/%s/cwd" % self.pid).rstrip('/')
+        except OSError as err:
+            if err.errno == errno.ENOENT:
+                os.stat("/proc/%s" % self.pid)
+                return None
```
giampaolo: There's no precedent of `cwd()` returning None, but there is for `exe()` returning `""` so I would return an empty string here.

psutil/_psaix.py
```diff
@@ -0,0 +1,488 @@
...
+            retlist.append(_common.popenfile(path, int(fd)))
+        return retlist
+
+    @wrap_exceptions
+    def num_fds(self):
+        return len(os.listdir("/proc/%s/fd" % self.pid))
+
+    @wrap_exceptions
+    def num_ctx_switches(self):
+        return _common.pctxsw(*cext.proc_num_ctx_switches(self.pid))
```
giampaolo: I thought this wasn't supported? (it's not from `psutil/__init__.py` at least)

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+    if (thread_count <= 0) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    /* Allocate enough memory */
+    threadt = (perfstat_thread_t *)calloc(thread_count,
+        sizeof(perfstat_thread_t));
+    if (threadt == NULL) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: You want `PyErr_NoMemory()` here.

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+}
+
+
+/*
+ * Return users currently connected on the system.
+ */
+static PyObject *
+psutil_users(PyObject *self, PyObject *args) {
+    struct utmpx *ut;
+    PyObject *ret_list = PyList_New(0);
```
giampaolo: In order to follow existing convention I recommend using a `py_` prefix for all python related variable names, so this should be `py_retlist`.

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+    PyObject *tuple = NULL;
+    PyObject *user_proc = NULL;
```
giampaolo: Same here (`py_tuple` and `py_user_proc`)

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+    while (NULL != (ut = getutxent())) {
+        if (ut->ut_type == USER_PROCESS)
+            user_proc = Py_True;
+        else
+            user_proc = Py_False;
+        tuple = Py_BuildValue(
+            "(sssfO)",
+            ut->ut_user,              // username
+            ut->ut_line,              // tty
+            ut->ut_host,              // hostname
```
giampaolo: Strings should be handled with care (it's a major change I introduced in 5.3.0). This is how they should be treated:
https://github.com/Infinidat/psutil/blob/c895a4aa1a08e7dac7213ab1c2458b0a28cea343/psutil/_psutil_linux.c#L469-L477

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    mt = getmntent(file);
+    while (mt != NULL) {
+        py_tuple = Py_BuildValue(
+            "(ssss)",
+            mt->mnt_fsname,   // device
+            mt->mnt_dir,    // mount point
```
giampaolo: New way to handle strings:
https://github.com/Infinidat/psutil/blob/c895a4aa1a08e7dac7213ab1c2458b0a28cea343/psutil/_psutil_linux.c#L469-L477

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+        goto error;
+    }
+    if (tot < 0) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+    statp = (perfstat_netadapter_t *)
+        malloc(tot * sizeof(perfstat_netadapter_t));
+    if (statp == NULL) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: Use `PyErr_NoMemory`

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+    ncpu = perfstat_cpu(NULL, NULL, sizeof(perfstat_cpu_t), 0);
+    if (ncpu <= 0){
+        PyErr_SetFromErrno(PyExc_OSError);
+        return NULL;
+    }
+
+    /* allocate enough memory to hold the ncpu structures */
+    cpu = (perfstat_cpu_t *) malloc(ncpu * sizeof(perfstat_cpu_t));
+    if (cpu == NULL) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: Use `PyErr_NoMemory`

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+            goto error;
+        Py_DECREF(py_cputime);
+    }
+    free(cpu);
+    return py_retlist;
+
+error:
+    Py_XDECREF(py_cputime);
+    Py_DECREF(py_retlist);
+    free(cpu);
```
giampaolo: I think you should free() it only if it hasn't been malloc()'ed before. You can set `cpu = NULL;` earlier and do `if (cpu != NULL) {free(cpu);}` here

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+    if (disk_count <= 0) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    /* Allocate enough memory */
+    diskt = (perfstat_disk_t *)calloc(disk_count,
+        sizeof(perfstat_disk_t));
+    if (diskt == NULL) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: PyErr_NoMemory

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,777 @@
...
+            diskt[i].__rxfers,
+            diskt[i].xfers - diskt[i].__rxfers,
+            diskt[i].rblks * diskt[i].bsize,
+            diskt[i].wblks * diskt[i].bsize,
+            diskt[i].rserv / 1000 / 1000,  // from nano to milli secs
+            diskt[i].wserv / 1000 / 1000   // from nano to milli secs
+        );
+        if (py_disk_info == NULL)
+            goto error;
+        if (PyDict_SetItemString(py_retdict, diskt[i].name,
```
giampaolo: Correct way to handle strings:
https://github.com/Infinidat/psutil/blob/c895a4aa1a08e7dac7213ab1c2458b0a28cea343/psutil/_psutil_linux.c#L469-L477

psutil/_psutil_aix.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2009, Giampaolo Rodola'. All rights reserved.
```
giampaolo: Please also add your name. Also do the same in other new files (copy license and add your name).

psutil/_psutil_posix.c
```diff
@@ -688,7 +692,7 @@ void init_psutil_posix(void)
     PyObject *module = Py_InitModule("_psutil_posix", PsutilMethods);
 #endif
 
-#if defined(PSUTIL_BSD) || defined(PSUTIL_OSX) || defined(PSUTIL_SUNOS)
+#if defined(PSUTIL_BSD) || defined(PSUTIL_OSX) || defined(PSUTIL_SUNOS) || defined(_AIX)
```
giampaolo: You should define `PSUTIL_AIX` in setup.py and use it in here

giampaolo (changes requested):
Here's an initial review. Overall, really great work! 
There are more issues but the ones I gave thus far are enough to keep you busy for quite a while.

psutil/TODO.aix
wiggin15: Yes, AIX is weird like that.

psutil/TODO.aix
wiggin15: On our environment we're using a custom version of Python, which on AIX uses a trick utilizing execve. This means the test will work on regular environments but not on ours specifically. I'm not sure this file should really be included in the repo (I'm also not sure it's up-to-date), this file was created for note taking when debugging the tests.

psutil/TODO.aix
wiggin15: Hmm, I guess so

psutil/TODO.aix
wiggin15: Yes. There are probably many changes to the tests I can do to make them pass. I deliberately did not change the unit tests for now to see what breaks - but I should before this will be really ready to be merged.

psutil/TODO.aix
wiggin15: "ps" calls PID 0 "swapper". I can return this name specifically for the case of pid=0.

psutil/TODO.aix
wiggin15: AFAIK they do exist and can be detected, but the trick we use to create one with `fork` doesn't work

psutil/TODO.aix
wiggin15: This test actually didn't fail on the last test runs I ran, I don't know why it failed and if this still happens. Like I said, this file may be a little out-of-date.

psutil/_psaix.py
wiggin15: This is the same implementation as in _pssunos.py

psutil/_psaix.py
wiggin15: This is the same implementation as in _pssunos.py

psutil/_psaix.py
wiggin15: This is the same implementation as in _pssunos.py

psutil/_psaix.py
wiggin15: Right. this should be removed.

psutil/TODO.aix
giampaolo: That's *really* unfortunate. How about doing this then?

```
def pids():
    """Returns a list of PIDs currently running on the system."""
    return [int(x) for x in os.listdir('/proc') if x.isdigit() \
            and  _psposix.pid_exists(int(x))]
```

Does this test pass?

psutil/TODO.aix
giampaolo: I'm not sure what you mean by this then. Simply make sure `name()` returns the expected string (in case of a Python process it should be "python".

psutil/TODO.aix
giampaolo: Sounds good. Also make sure `pids()` return PID 0 and `pid_exists(0)` return True. If they don't by default then you can simply hard-code it. `ps` should be used as the leading reference and psutil should behave the same as ps.

psutil/TODO.aix
giampaolo: What happens when you call this test function?
https://github.com/giampaolo/psutil/blob/8b8da39e0c62432504fb5f67c418715aad35b291/psutil/tests/__init__.py#L335
Does it crash? Does it return a process but it's status is not zombie (what is it?)?

psutil/_psaix.py
giampaolo: Then that should be changed but we can ignore it for now, If `pid_exists()` is reliable in this case (it should) then let's just use that.

psutil/_psaix.py
giampaolo: Oh I see. I don't remember why I did that but let's just return None for now.

psutil/TODO.aix
wiggin15: This fixes `test_pid_exists_2` but causes `test_pids` to break because `ps` still returns the processes that "don't really exist" and then we get this:

      File "/root/psutil/psutil/tests/test_posix.py", line 297, in test_pids
         self.fail("difference: " + str(difference))
    AssertionError: difference: [131076, 196614, 1048608, 1114146, 1179684, 1376298, 1638568]

psutil/TODO.aix
wiggin15: execve causes the name to change so `sys.executable` and `name` don't match. 

    AssertionError: ('python2.7', 'python2.7.bin')

This is implementation detail in our environment and can be ignored.
I found out that the set of tests that fail is very different from before (it's more stable now too, i.e. less "flaky" tests). This file will be updated in the next commit.

psutil/TODO.aix
wiggin15: Apparently this doesn't fail any more

psutil/TODO.aix
wiggin15: This creates a process, but its status is 0 (SNONE) and not SZOMB. Then we crash here:

    call_until(lambda: zproc.status(), "ret == psutil.STATUS_ZOMBIE")

I'm not sure what that means.

psutil/_psaix.py
wiggin15: No `free`, but a command called `svmem`. Will do.

psutil/_psaix.py
wiggin15: I am already using a shell command to determine number of processors (lsdev). I'm not sure what I can test this against.

psutil/_psutil_aix.c
wiggin15: It looks like PyUnicode_DecodeFSDefault is available on Python 3 only? When using it my code doesn't compile. Looks like `#ifdef`s are required.

psutil/_psutil_aix.c
wiggin15: We can't get to `error` without allocating `cpu`, so this is safe.

wiggin15:
I added a new commit. Still no AIX-specific tests and no unicode string handling, but I fixed some broken functions and started changing the unit tests a bit. I think this looks better.

psutil/_psaix.py
giampaolo: You're right.

giampaolo: So please just add a `# TODO: rewrite this in C` comment

psutil/_psaix.py
```diff
@@ -0,0 +1,525 @@
...
+
+    @wrap_exceptions
+    def connections(self, kind='inet'):
+        ret = net_connections(kind, _pid=self.pid)
+        # The underlying C implementation retrieves all OS connections
+        # and filters them by PID.  At this point we can't tell whether
+        # an empty list means there were no connections for process or
+        # process is no longer active so we force NSP in case the PID
+        # is no longer there.
+        if not ret:
```
giampaolo: I suggest to remove the `if not ret` line. There may be some connections but the process may have disappeared beforehand so we basically want to always call `os.stat`

psutil/_psaix.py
```diff
@@ -0,0 +1,525 @@
...
+
+    @wrap_exceptions
+    def cpu_times(self):
+        cpu_times = cext.proc_cpu_times(self.pid)
+        return _common.pcputimes(*cpu_times)
+
+    @wrap_exceptions
+    def terminal(self):
+        psinfo = self._proc_basic_info()
+        ttydev = psinfo[-1]
```
giampaolo: I think you better use a "fixed" integer (5, 6 or whatever) because if you add a new element to the returned list this will break

psutil/_psaix.py
```diff
@@ -0,0 +1,525 @@
...
+    def open_files(self):
+        # TODO rewrite without using procfiles (stat /proc/pid/fd/* and then
+        # find matching name of the inode)
+        p = subprocess.Popen(["/usr/bin/procfiles", "-n", str(self.pid)],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = p.communicate()
+        if PY3:
+            stdout, stderr = [x.decode(sys.stdout.encoding)
+                              for x in (stdout, stderr)]
+        if "no such process" in stderr:
```
giampaolo: perhaps do `stderr.lower()`?

psutil/_psaix.py
```diff
@@ -0,0 +1,525 @@
...
+            raise NoSuchProcess(self.pid, self._name)
+        procfiles = re.findall("(\d+): S_IFREG.*\s*.*name:(.*)\n", stdout)
+        retlist = []
+        for fd, path in procfiles:
+            path = path.strip()
+            if path.startswith("//"):
+                path = path[1:]
+            if path == "Cannot be retrieved":
```
giampaolo: perhaps do `path.lower()`?

psutil/_psutil_aix.c
giampaolo: I backported provided a "Python 2 like" version of `PyUnicode_DecodeFSDefault`  in `_psutil_common.c` so you just have to `#include "_psutil_common.h"` at the top of this C file in order to use it.

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,815 @@
...
+    }
+    endutxent();
+
+    return py_retlist;
+
+error:
+    Py_XDECREF(py_tuple);
+    Py_DECREF(py_retlist);
+    if (ut != NULL)
+        endutent();
```
giampaolo: You use `endutxent();` on line 288, not this. Why?

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,815 @@
...
+
+    setutxent();
+    while (NULL != (ut = getutxent())) {
+        if (ut->ut_type == BOOT_TIME) {
+            boot_time = (float)ut->ut_tv.tv_sec;
+            break;
+        }
+    }
+    endutxent();
+    if (boot_time != 0.0) {
```
giampaolo: Are there cases where time can be 0.0? Perhaps we need a comment here?

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,815 @@
...
+        return NULL;
+    }
+
+    return Py_BuildValue(
+        "IIIII",
+        cpu.pswitch,
+        cpu.syscall,
+        cpu.devintrs,
+        cpu.softintrs,
+        cpu.traps
```
giampaolo: In TODO.aix you state all of these numbers are 0, correct?

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+        }
+        lport = (int)ntohs(inp.inp_lport);
+
+        inet_ntop(fam, laddr, laddr_str, sizeof(laddr_str));
+
+        if (raddr != NULL) {
+            inet_ntop(fam, raddr, raddr_str, sizeof(raddr_str));
+            return Py_BuildValue("(iii(si)(si)ii)", fd, fam,
+                                 s.so_type, laddr_str, lport, raddr_str,
+                                 rport, state, pid);
```
giampaolo: I suspect returning 2 tuples in a single shot like this (`(si)`) will leak memory. You need 3 tuples, 2 per addresses and 1 per connection. It's tedious work but you can pretty much copy & paste from here:
https://github.com/giampaolo/psutil/blob/f435c2b6d308c12d0df33b10d828d97df303a614/psutil/_psutil_osx.c#L1332-L1349

giampaolo: Note: you can use `make test-memleaks` to make sure of this.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    if (Kd < 0) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    /* Read the process table */
+    msz = (size_t)(PROCSIZE * PROCINFO_INCR);
+    processes = (struct procentry64 *)malloc(msz);
+    if (!processes) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: `PyErr_NoMemory()`

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    p = processes;
+    while ((i = getprocs64(p, PROCSIZE, (struct fdsinfo64 *)NULL, 0, &pid,
+                 PROCINFO_INCR))
+    == PROCINFO_INCR) {
+        np += PROCINFO_INCR;
+        if (np >= Np) {
+            msz = (size_t)(PROCSIZE * (Np + PROCINFO_INCR));
+            processes = (struct procentry64 *)realloc((char *)processes, msz);
+            if (!processes) {
+                PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: `PyErr_NoMemory()`

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+        if (requested_pid != -1 && requested_pid != pid)
+            continue;
+        if (p->pi_state == 0 || p->pi_state == SZOMB)
+            continue;
+
+
+        if (!fds) {
+            fds = (struct fdsinfo64 *)malloc((size_t)FDSINFOSIZE);
+            if (!fds) {
+                PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: `PyErr_NoMemory()`

setup.py
```diff
@@ -254,6 +264,8 @@ def get_ethtool_macro():
         if platform.release() == '5.10':
             posix_extension.sources.append('psutil/arch/solaris/v10/ifaddrs.c')
             posix_extension.define_macros.append(('PSUTIL_SUNOS10', 1))
+    if AIX:
```
giampaolo: `elif` is better

giampaolo:
Good work on addressing the first set of comments! I'm gonna add another "big" request.
Right now "/proc" is hard coded but in the py and the C file. It should be possible to change that via `psutil.PROCFS_PATH`. In order to do that.
Copy this in `_psaix.py`:
https://github.com/Infinidat/psutil/blob/cac89d522d64571a53011cd0914a654a9c5e2869/psutil/_pslinux.py#L215
...then store it in the `Process` class like this:
https://github.com/Infinidat/psutil/blob/cac89d522d64571a53011cd0914a654a9c5e2869/psutil/_pslinux.py#L1392
...and pass that string to the underlying C functions that need it. 

On the C side of things you'll now have to accept 2 args, the PID and /proc path prefix, like this:
https://github.com/Infinidat/psutil/blob/cac89d522d64571a53011cd0914a654a9c5e2869/psutil/_psutil_sunos.c#L96-L101

Not sure if I've been clear (I hope =)).

psutil/_psaix.py
wiggin15: This is again the same as in _pssunos.py

psutil/_psutil_aix.c
wiggin15: Looks like you're right.
Again, this is the same as in the sunos implementation.

psutil/_psutil_aix.c
wiggin15: Not sure. This too is copied from _psutil_sunos.c.

psutil/_psutil_aix.c
wiggin15: Not all of them, only `syscall` and `traps`. I'm not sure why.

psutil/arch/aix/net_connections.c
wiggin15: What would be the result of `make test-memleaks` if there is a memory leak? Running this command returned 3 failed test:

    test_memory_leaks.TestProcessObjectLeaks.test_num_ctx_switches
    test_memory_leaks.TestTerminatedProcessLeaks.test_io_counters
    test_memory_leaks.TestTerminatedProcessLeaks.test_num_ctx_switches

The failures don't seem to be due to memory leaks.

psutil/_psaix.py
wiggin15: I tried hard to find how to rewrite this in C and couldn't find any way to do that. Some functions I did find returned wrong results. I ran truss (the equivalent of strace) to find out how `lsdev` itself returns the results, and it looks like it iterates over the devices from the ODM (the equivalent of Windows registry). I don't think reading from the ODM in psutil is better than just parsing the output of `lsdev`.

wiggin15:
What's the reason for making `/proc` changeable? Is this configurable on the operating system?

giampaolo:
> What's the reason for making /proc changeable?

https://github.com/giampaolo/psutil/issues/558

psutil/arch/aix/net_connections.c
giampaolo: Let's think about memory leaks later, after you fixed unicode strings and /proc fs location.

wiggin15:
I think, instead of moving the `get` function to `_common`, we can move this function to the main `__init__.py` file and then we won't have to use the `sys.modules` trick. Something like this in `__init__`:

    if LINUX or SOLARIS or AIX:
        # This is a public writable constant which is read from platform-specific code
        PROCFS_PATH = "/proc"

        def get_procfs_path():
            return PROCFS_PATH

What do you think?
In addition, I noticed that the BSD code is still using hard-coded `/proc`. Is it not relevant in that platform? Maybe we can change to `if POSIX` instead of Linux/Solaris/AIX if the BSD module is changed.

psutil/_psutil_aix.c
wiggin15: `boot_time` will be `0.0` if the loop breaks without finding `BOOT_TIME`. We can add a comment, but again this is common implementation with Solaris.

giampaolo:
How can you use that from `_ps*.py` modules if it's defined in `__init__.py`? It's `__init__` which imports those module not vice versa. Anyway it looks like something which can be dealt with in another PR so for the time being you can also define it in `_psaix.py` and then we can move it later.

> In addition, I noticed that the BSD code is still using hard-coded /proc. Is it not relevant in that platform? 

It's not, BSD systems do not rely on /proc.

psutil/_psutil_aix.c
giampaolo: Oh yeah. I now realize that sunos has different "minor" issues like this one (and others I mentioned here). I will stop being picky and just focus on more important aspects. =)

psutil/_psutil_aix.c
giampaolo: OK then. If it's an OS problem there's nothing we can do.

wiggin15:
> How can you use that from _ps*.py modules if it's defined in __init__.py?

It's a matter of import order. If you define it in `__init__` before importing the platform-specific modules, they will succeed in importing from `__init__`. Maybe not the best way to do it though.


> BSD systems do not rely on /proc.

`_psbsd.py` has a few `readlink`s and `open`s from `/proc` (the C extension doesn't, but still..)

psutil/_psutil_aix.c
wiggin15: There are definitely some comments that are relevant and should be fixed in Solaris too. There are also a number of tests that fail in Solaris as well as on AIX that can be fixed in both. I'm keeping track of your comments and may submit a separate PR to fix them after this one.

psutil/_psaix.py
wiggin15: I'm having trouble checking this function against the shell command because there memory usage changes between the call to `svmon` and the call to `psutil.virtual_memory`. I think checking whether they are "close" can be inconsistent. This function has basic tests in `test_system.py`...

wiggin15:
Added more commits :)

psutil/TODO.aix
```diff
@@ -0,0 +1,25 @@
...
+    test_prog_w_funky_name      funky name tests don't work, name is empty (also on Solaris)
+    test_setup_script           TBD (also on Solaris)
+    test_cmdline                long args are cut from cmdline in /proc/pid/psinfo and getargs
+    test_io_counters            io_counters read count is always 0
+    test_zombie_process         trying to create a zombie process doesn't create a zombie process on AIX
+    test_create_zombie_proc     trying to create a zombie process doesn't create a zombie process on AIX
+    test_procinfo               missing API
+    test_cpu_stats              cpu_stats always returns ctx_switches=0
+    test_disk_usage             df returns "-" for all procfs fields but API returns something real
+    test_pid_exists_2           there are pids in /proc that don't really exist
```
giampaolo: This still worries me and should be discussed. I suppose I'll have a better idea of what this means once I look at the test results.

psutil/__init__.py
```diff
@@ -890,11 +898,12 @@ def num_handles(self):
             """
             return self._proc.num_handles()
 
-    def num_ctx_switches(self):
-        """Return the number of voluntary and involuntary context
-        switches performed by this process.
-        """
-        return self._proc.num_ctx_switches()
+    if not AIX:
```
giampaolo: I suggest using `if hasattr(_psplatform.Process, "num_ctx_switches"):` instead

docs/index.rst
```diff
@@ -2110,7 +2120,7 @@ Constants
 .. _const-procfs_path:
 .. data:: PROCFS_PATH
 
-  The path of the /proc filesystem on Linux and Solaris (defaults to
+  The path of the /proc filesystem on Linux, Solaris and AIX (defaults to
   ``"/proc"``).
   You may want to re-set this constant right after importing psutil in case
   your /proc filesystem is mounted elsewhere or if you want to retrieve
```
giampaolo: Please also add the new `AIX` constant in the doc and mark it as `.. versionadded:: 5.4.0`.

psutil/arch/aix/ifaddrs.c
```diff
@@ -0,0 +1,125 @@
...
+    ecp = ccp + ifsize;
+
+    while (ccp < ecp) {
+
+        ifr = (struct ifreq *) ccp;
+        ifsize = sizeof(ifr->ifr_name) + SIZE(ifr->ifr_addr);
+        fam = ifr->ifr_addr.sa_family;
+
+        if (fam == AF_INET || fam == AF_INET6) {
+            cifa = (struct ifaddrs *) calloc(1, sizeof(struct ifaddrs));
```
giampaolo: It seems you need to check `calloc` return value

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    int Kd,             /* kernel memory file descriptor */
+    KA_T addr,          /* kernel memory address */
+    char *buf,          /* buffer to receive data */
+    size_t len) {       /* length to read */
+    int br;
+
+    if (lseek64(Kd, (off64_t)addr, L_SET) == (off64_t)-1)
+        return(1);
+    br = read(Kd, buf, len);
+    return((br == len) ? 0 : 1);
```
giampaolo: If I understand this correctly you return `1` (meaning error) in case of len/size mismatch. Whereas the previous `lseek` and `read` calls set errno, this line (53) won't, so I think you need to do this:

```
if (br != len) {
    PyErr_SetString(PyExc_RuntimeError, 
                    "size mismatch when reading kernel memory fd");
    return 1;
}
return 0;
```

Also, since you set the exception in here, you should do the same for the 2 calls above (`lseek` and `read`) and do `PyErr_SetFromErrno(PyExc_OSError);`.
Of course the callers of this function should just return NULL.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+
+    uo = (int)(mb.m_hdr.mh_data - unp_addr);
+    if ((uo + sizeof(struct sockaddr)) <= sizeof(mb))
+        ua = (struct sockaddr_un *)((char *)&mb + uo);
+    else {
+        if (mb.m_hdr.mh_data
+        &&  !psutil_kread(Kd, (KA_T)mb.m_hdr.mh_data, (char *)&un, sizeof(un))
+        ) {
+        ua = &un;
+        }
```
giampaolo: Can you please indent it like this (easier to read)?:

```
        if (mb.m_hdr.mh_data && !psutil_kread(
                Kd, (KA_T)mb.m_hdr.mh_data, (char *)&un, sizeof(un))) {
            ua = &un;
        }
```

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+            PyErr_SetFromErrno(PyExc_OSError);
+            return NULL;
+        }
+        if ((KA_T) f.f_data != (KA_T) unp.unp_socket) {
+            PyErr_SetFromErrno(PyExc_OSError);
+            return NULL;
+        }
+
+        if (unp.unp_addr) {
+            read_unp_addr(Kd, unp.unp_addr, unix_laddr_str,
```
giampaolo: Shouldn't you check return value? `read_unp_addr` uses `psutil_kread` so it seems you should.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+                sizeof(unix_laddr_str));
+        }
+
+        if (unp.unp_conn) {
+            if (psutil_kread(Kd, (KA_T) unp.unp_conn, (char *)&unp,
+                sizeof(unp))) {
+                PyErr_SetFromErrno(PyExc_OSError);
+                return NULL;
+            }
+            read_unp_addr(Kd, unp.unp_addr, unix_raddr_str,
```
giampaolo: Shouldn't you check return value? `read_unp_addr` uses `psutil_kread` so it seems you should.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    if (!s.so_type) {
+        return NO_SOCKET;
+    }
+
+    if (!s.so_proto ||
+        psutil_kread(Kd, (KA_T)s.so_proto, (char *)&p, sizeof(p))) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        return NULL;
```
giampaolo: Mmmm code smell. If `!s.so_proto` condition is true `errno` will never be set and you'll return an "empty" exception.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    if (!p.pr_domain
```
giampaolo: Same here

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    ||  psutil_kread(Kd, (KA_T)p.pr_domain, (char *)&d, sizeof(d))) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        return NULL;
+    }
+
+    fam = d.dom_family;
+    if (fam == AF_INET || fam == AF_INET6) {
+        /* Read protocol control block */
+        if (!s.so_pcb
+        ||  psutil_kread(Kd, (KA_T) s.so_pcb, (char *) &inp, sizeof(inp))) {
```
giampaolo: And here

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+
+    if (fam == AF_UNIX) {
+        if (psutil_kread(Kd, (KA_T) s.so_pcb, (char *)&unp, sizeof(unp))) {
+            PyErr_SetFromErrno(PyExc_OSError);
+            return NULL;
+        }
+        if ((KA_T) f.f_data != (KA_T) unp.unp_socket) {
+            PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: Mmmm I suppose `errno` will not be set here, so this will fail with an "empty" exception. Perhaps you want RuntimeError?

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+    PyObject *py_tuple = NULL;
+    KA_T fp;
+    int Kd = -1;
+    int i, np;
+    struct procentry64 *p;
+    struct fdsinfo64 *fds = (struct fdsinfo64 *)NULL;
+    size_t msz;
+    pid32_t requested_pid;
+    pid32_t pid;
+    int Np = 0;          /* number of processes */
```
giampaolo: Why not call it  `num_procs`? It would be less confusing also because there's another `np` variable.

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+                    /* the process table */
+
+    if (py_retlist == NULL)
+        goto error;
+    if (! PyArg_ParseTuple(args, "i", &requested_pid))
+        goto error;
+
+    Kd = open(KMEM, O_RDONLY, 0);
+    if (Kd < 0) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo: Looks like you can use `PyErr_SetFromErrnoWithFilename` instead

psutil/arch/aix/net_connections.c
```diff
@@ -0,0 +1,328 @@
...
+                PyErr_NoMemory();
+                goto error;
+            }
+            Np += PROCINFO_INCR;
+        }
+        p = (struct procentry64 *)((char *)processes + (np * PROCSIZE));
+    }
+
+    if (i > 0)
+        np += i;
```
giampaolo: Is this on purpose?

giampaolo (changes requested):
Great work with unicode handling and /proc changes! 
I added some comments about the C code. 
When those are fixed I suppose we can start looking at the the output of `make test` and `make test-memleaks`.

psutil/arch/aix/net_connections.c
wiggin15: This is from lsof code, which is pretty terrible:
https://fossies.org/dox/lsof_4.89_src/aix_2dproc_8c_source.html
Interestingly both `np` and `Np` are for the number of processes in `processes`. `Np` is the number allocated and `np` is the number read. I know these are not good names but I prefer to stick to the original names for comparing to the original code (it helped just now).

psutil/arch/aix/net_connections.c
wiggin15: Also from [lsof code](https://fossies.org/dox/lsof_4.89_src/aix_2dproc_8c_source.html), line 476. This code is tricky but it's correct. We're reading the processes into `p` in `PROCINFO_INCR` increments. We `realloc` the next "block" for `processes` every time we read a full `PROCINFO_INCR` block. The last call can return less than `PROCINFO_INCR` processes, but we still read that block and need to go over it, so `np` must be increased by the number of processes in the last block - which is in `i`. See the [docs for `getprocs64`](https://www.ibm.com/support/knowledgecenter/en/ssw_aix_61/com.ibm.aix.basetrf1/getprocs.htm)
Confusing, I know.

docs/index.rst
wiggin15: The docs have a section called `.. _const-oses:`. Is it ok if I add `AIX` to this section with:

    .. versionchanged:: 5.4.0 added AIX

in order to keep this const in the same group as the rest?

psutil/arch/aix/net_connections.c
giampaolo: I see. OK then. And yes, I remember lsof source code: I remember I gave up because I didn't understand it. =)

wiggin15:
Added a new commit and uploaded the latest test results to https://gist.github.com/wiggin15/1d76d1dc1ad50e6665fe4dc5c6fa898a
please ignore 2 tests that fail with `python2.7.bin`, this is caused by a trick we're using on our environment.

giampaolo:
```
  File "/root/psutil/scripts/procinfo.py", line 228, in run
    print_("ctx-switches", str_ntuple(pinfo['num_ctx_switches']))
KeyError: 'num_ctx_switches'
```

This is easy: just change scripts/procinfo.py so that it won't print this info on AIX

```
  File "/root/psutil/psutil/tests/test_posix.py", line 368, in df
    total = int(fields[1]) * 1024
ValueError: invalid literal for int() with base 10: '-'
```

This means `df -k` output is different on AIX compared to other POSIXes. You can simply skip thi test on AIX.

```
  File "/root/psutil/psutil/tests/test_contracts.py", line 541, in cwd
    self.assertIsInstance(ret, str)
AssertionError: None is not an instance of <type 'str'>
```

Just return an empty string in cwd() method of _psaix.py instead of None.

```
FAIL: psutil.tests.test_misc.TestMisc.test_setup_script
  File "/root/psutil/psutil/tests/test_misc.py", line 369, in test_setup_script
    self.assertRaises(SystemExit, module.setup)
AssertionError: SystemExit not raised
```

Just skip this one.

```
FAIL: psutil.tests.test_misc.TestProcessUtils.test_create_zombie_proc
  File "/root/psutil/psutil/tests/test_misc.py", line 950, in test_create_zombie_proc
    zpid = create_zombie_proc()
  File "/root/psutil/psutil/tests/__init__.py", line 366, in create_zombie_proc
    call_until(lambda: zproc.status(), "ret == psutil.STATUS_ZOMBIE")
  File "/root/psutil/psutil/tests/__init__.py", line 593, in wrapper
    return fun(*args, **kwargs)
  File "/root/psutil/psutil/tests/__init__.py", line 643, in call_until
    assert eval(expr)
AssertionError
```

OK, you said `create_zombie_proc()` doesn't work on AIX. Have you checked whether `ps` is able to detect it though? I just want to be sure it's not `status()` method which is broken. If we make sure it's not and we don't find a way to correctly create a zombie process via unit tests then let' just skip this. 

```

  File "/root/psutil/psutil/tests/test_process.py", line 720, in test_cmdline
    ' '.join(cmdline))
AssertionError: '/root/python/bin/python2.7 -c' != '/root/python/bin/python2.7 -c import time; time.sleep(60)'
```

I would just change the test and do a comparison of the first 30 chars if you're on AIX, else just do what we're currently doing.

```
  File "/root/psutil/psutil/tests/test_process.py", line 322, in test_io_counters
    self.assertGreater(io2.read_count, io1.read_count)
AssertionError: 0L not greater than 0L
```

Same here. If we know io_counters()'s read_count is always zero just skip this specific test line when on AIX.

```
FAIL: psutil.tests.test_process.TestProcess.test_prog_w_funky_name
  File "/root/psutil/psutil/tests/test_process.py", line 755, in test_prog_w_funky_name
    self.assertEqual(p.cmdline(), cmdline)
AssertionError: Lists differ: [''] != ['/root/psutil/$testfnfoo bar ...
```

This is bad and should be fixed. The test may also not pass but we should never have `['']` returned.

```

FAIL: psutil.tests.test_system.TestSystemAPIs.test_cpu_stats
  File "/root/psutil/psutil/tests/test_system.py", line 746, in test_cpu_stats
    self.assertGreater(value, 0)
AssertionError: 0 not greater than 0
```

Again, if we know 'ctx_switches' and 'interrupts' are always 0 just skip this test on solaris (just add a comment / note)


```
FAIL: psutil.tests.test_system.TestSystemAPIs.test_pid_exists_2
  File "/root/psutil/psutil/tests/test_system.py", line 238, in test_pid_exists_2
    self.fail(pid)
AssertionError: 260
```

This is bad. It means pid_exists() is broken. Does AIX have /proc/{pid}/status? Linux uses a specific logic for pid_exists() as it takes the Tid (thread ID) into account. I wonder if AIX does the same.


What about memory leak tests?

giampaolo:
Note: it may also be `pids()` returning a PID which does not exist, not necessarily `pid_exists` which is broken. Anyway, it's either one or the other. `ps` is your friend and should tell you who's wrong.

wiggin15:
Added a new commit

> Just return an empty string in cwd() method of _psaix.py instead of None.

I noticed that the test for `cwd` already checks whether the return value is `None` (which can also happen on BSD and Solaris), but the test checks if it's `None` after making sure it's a string, which means it's a bug in the test. I fixed the test instead. I also fixed some other failures in `test_fetch_all`

> test_setup_script

I realized why it failed. It's because I used `python setup.py test` instead of `make test` to run the unit tests. Running the former makes argv ok for setup.setup() so it doesn't raise SystemExit with usage. I'll use `make test` from now on.

> OK, you said create_zombie_proc() doesn't work on AIX. Have you checked whether ps is able to detect it though? I just want to be sure it's not status()

OK, so it turns out you *can* create zombie processes and there was indeed a bug in `status`. It's not simple to determine actual process state from /proc, but I think I added the code required.

> I would just change the test and do a comparison of the first 30 chars if you're on AIX

The problem isn't that we cut to 30 characters (according to the struct in `/proc` the max should be 80), but that there is a single argument `import time; time.sleep(60)` which is too long so it's somehow truncated from `psinfo`. If I add another argument to argv after this argument, it will show in psinfo while "skipping" the long argument.

> It means pid_exists() is broken

The PIDs in question do exist, in `ps` and in `/proc`. They have `/proc/<pid>/psinfo`, and even `psutil.Process(pid)` works on them. Yet pid_exists returns False, because the "send a signal" check, even when trying with Python's `os.kill`, raises `OSError: no such process`. I noticed these processes are all named `wait` and it appears that there is 1 per logical CPU. It's an AIX thing...

> The test may also not pass but we should never have [''] returned.

I traced this back to an issue with our environment, again. `create_exe` used to compile C code for the test executable but now it's using `sys.executable` instead (also, there's a bug in that function where the default code is never used). Apparently `sys.executable` is not copy-able on our environment (again the `execve python2.7.bin` trick) so when running it it immediately exits, and that's why we get the empty cmdline. When changing `create_exe` (not committed, of course), those tests pass except `test_prog_w_funky_name` which, again, doesn't pass on Solaris either.

> What about memory leak tests?

Those tests pass now after a few small changes.

New test results in: https://gist.github.com/wiggin15/d1054f0a4a172a572dd610aad649dfa7

giampaolo:
> The PIDs in question do exist, in ps and in /proc. They have /proc/<pid>/psinfo, and even psutil.Process(pid) works on them. Yet pid_exists returns False, because the "send a signal" check, even when trying with Python's os.kill, raises OSError: no such process. I noticed these processes are all named wait and it appears that there is 1 per logical CPU. It's an AIX thing...

OK, so I suppose the `os.kill` trick is not reliable. What about `return os.path.exitsts("/proc/{pid}")`? Does it work? Last possibility is doing `return pid in pids()` but that's slow.... :-\
I will reply to other comments later (need to sleep =)).

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,870 @@
...
+    PyObject *py_retdict = PyDict_New();
+    PyObject *py_ifc_info = NULL;
+
+    if (py_retdict == NULL)
+        return NULL;
+
+    /* check how many perfstat_netinterface_t structures are available */
+    tot = perfstat_netinterface(NULL, NULL, sizeof(perfstat_netinterface_t), 0);
+    if (tot == 0) {
+        PyErr_SetString(PyExc_RuntimeError, "no network interface found");
```
giampaolo: I think it's legitimate to have a box without any NIC installed. We probably better return an empty dict.

psutil/_psutil_aix.c
giampaolo: Please just add 2 XXX comments to signal that syscall and traps fields are always 0.

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,870 @@
...
+
+    for (i = 0; i < disk_count; i++) {
+        py_disk_info = Py_BuildValue(
+            "KKKKKK",
+            diskt[i].__rxfers,
+            diskt[i].xfers - diskt[i].__rxfers,
+            diskt[i].rblks * diskt[i].bsize,
+            diskt[i].wblks * diskt[i].bsize,
+            diskt[i].rserv / 1000 / 1000,  // from nano to milli secs
+            diskt[i].wserv / 1000 / 1000   // from nano to milli secs
```
giampaolo: Please add a XXX comment telling the field which is always set to 0

psutil/arch/aix/ifaddrs.h
```diff
@@ -0,0 +1,33 @@
...
+#undef  ifa_broadaddr
+#define ifa_broadaddr ifa_dstaddr
+
+struct ifaddrs {
+  struct ifaddrs  *ifa_next;
+  char            *ifa_name;
+  unsigned int     ifa_flags;
+  struct sockaddr *ifa_addr;
+  struct sockaddr *ifa_netmask;
+  struct sockaddr *ifa_dstaddr;
```
giampaolo: Please indent by 4

psutil/tests/test_contracts.py
```diff
@@ -282,6 +284,9 @@ def test_fetch_all(self):
             'send_signal', 'suspend', 'resume', 'terminate', 'kill', 'wait',
             'as_dict', 'parent', 'children', 'memory_info_ex', 'oneshot',
         ])
+        if AIX:
+            # "<exiting>" processes really don't have names
+            excluded_names.add('name')
```
giampaolo: I wouldn't do this. Instead I would change the `name` method (defined later) like this:

```
    def name(self, ret, proc):
        self.assertIsInstance(ret, str)
        if not AIX:
            assert ret
        # ...else on AIX "<exiting>" processes (zombies) don't have names
```

psutil/tests/test_memory_leaks.py
```diff
@@ -288,6 +289,7 @@ def test_num_fds(self):
         self.execute(self.proc.num_fds)
 
     @skip_if_linux()
+    @unittest.skipIf(AIX, "num_ctx_switches not available on AIX")
```
giampaolo: Please instead define a HAS_NUM_CTX_SWITCHES global in `tests/__init__.py` and use it here. I prefer that way so I can keep track of supported things in a single place, in case their support is added later.

giampaolo (changes requested):
Some more comments. We're almost there.

giampaolo:
Also, can you please make sure all C lines are less than 80 chars long?

psutil/_psutil_aix.c
```diff
@@ -0,0 +1,870 @@
...
+            mt->mnt_type,   // fs type
+            mt->mnt_opts);  // options
+        if (py_tuple == NULL)
+            goto error;
+        if (PyList_Append(py_retlist, py_tuple))
+            goto error;
+        Py_DECREF(py_dev);
+        Py_DECREF(py_mountp);
+        Py_DECREF(py_tuple);
+        mt = getmntent(file);
```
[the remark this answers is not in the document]
wiggin15: This is just like the Linux and Solaris implementations and I think it's ok.

psutil/_psaix.py
```diff
@@ -0,0 +1,592 @@
...
+            if not os.path.isabs(exe):
+                # if cwd has changed, we're out of luck - this may be wrong!
+                exe = os.path.abspath(os.path.join(self.cwd(), exe))
+            if (os.path.isabs(exe) and
+               os.path.isfile(exe) and
+               os.access(exe, os.X_OK)):
+                return exe
+            # not found, move to search in PATH using basename only
+            exe = os.path.basename(exe)
+        # search for exe name PATH
```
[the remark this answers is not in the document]
wiggin15: There's a comment at the top of the function, "guessing is more complex than what's in the wrapping class". I don't really remember this, but it appears that we have to search in `PATH` which the parent class doesn't do.

psutil/_psutil_aix.c
wiggin15: Ugh. My bad, there was a bug here (I don't like it when I have bugs :/). The values are ulonglong, so should be built with `K` instead of `I`. The values are _not_ 0.

psutil/_psutil_aix.c
wiggin15: These values (disk_io_counters) are ok, it's `proc_io_counters` that's returning 0s. I will add `XXX` comments there.

psutil/_psaix.py
wiggin15: Why return `None` from `cpu_count_physical` instead of 0?

psutil/_psaix.py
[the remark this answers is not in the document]
giampaolo: Yeah that's what I don't like as it seems too risky. You may have two exes with the same name in different PATHs. Plus it may be a bit too expensive / slow.

psutil/_psutil_aix.c
giampaolo: Perfect. Just update the doc/TODO.aix then

psutil/_psaix.py
giampaolo: It's by design and documented: when cannot be determined we're supposed to return None. `os.cpu_count()` on Python 3 behaves the same.

wiggin15:
Added a new commit with AIX-specific tests

giampaolo:
As far as I'm concerned this looks good enough to be merged at this point so just tell me when you want me to do it.

wiggin15:
I think that this is ready now, except I'm not sure TODO.aix should be where it is (or stay at all), and maybe you want to update the main docs about the state of support for AIX (experimental?)

giampaolo:
It's OK, I will update doc / README, etc. before the next release, once I'm back from China. In the meantime, thanks a lot. You really did an amazing work. At first I was skeptical about adding support for a platform I can't test against, but the quality of the code you provided is excellent and you were great at addressing all my remarks.  I hope you will keep hacking on psutil as a main contributor.

PR #1130, opened 2017-09-24
Solaris: Use the correct set/get/end functions for utmp

wiggin15:
boot_time() and users() use utmp (getutxent) on Solaris, but don't call "set" and don't use the correct "end" variant (for utmpx)

Theoretically this can confuse psutil if utmp is already consumed, like so:

    >>> import ctypes
    >>> import psutil
    >>> libc = ctypes.cdll.LoadLibrary("/usr/lib/64/libc.so")
    >>> while libc.getutent():
    ...     pass
    ...
    >>> psutil.boot_time()
      File "/root/python/lib64/python2.7/site-packages/psutil-5.2.0.1-py2.7-solaris-2.10-i86pc.64bit.egg/psutil/__init__.py", line 2281, in boot_time
        return _psplatform.boot_time()
      File "/root/python/lib64/python2.7/site-packages/psutil-5.2.0.1-py2.7-solaris-2.10-i86pc.64bit.egg/psutil/_pssunos.py", line 285, in boot_time
        return cext.boot_time()
    RuntimeError: can't determine boot time

(I don't think this is a real enough scenario to make a unit test out of this, but this is an example of why it's important)

### DIFF
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -548,6 +548,7 @@ psutil_users(PyObject *self, PyObject *args) {
     if (py_retlist == NULL)
         return NULL;
 
+    setutxent();
     while (NULL != (ut = getutxent())) {
         if (ut->ut_type == USER_PROCESS)
             py_user_proc = Py_True;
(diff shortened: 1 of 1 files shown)


giampaolo:
I am not sure I understand the practical implications of this. Can you describe that as an entry in HISTORY.rst? If it's just a theoretical issue than it's also fine to merge this as-is.

wiggin15:
It's just theoretical. It's just the "right way" to use the functions. It's not important enough to be an entry in HISTORY.rst.

Issue #1129, opened 2017-09-22
No such file or directory: '/sys/class/hwmon/hwmon2/fan4_input'


ssaip:
I face the following issue running psutil (5.3.1) used by glances v2.11.1. Might be related to #990 .
glances -d gives the following:

```
2017-09-22 02:28:48,483 -- CRITICAL -- Error while initializing the sensors plugin ([Errno 2] No such file or directory: '/sys/class/hwmon/hwmon2/fan4_input')
2017-09-22 02:28:48,484 -- ERROR -- Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/glances/stats.py", line 117, in _load_plugin
    self._plugins[name] = plugin.Plugin(args=args)
  File "/usr/local/lib/python2.7/dist-packages/glances/plugins/glances_sensors.py", line 53, in __init__
    self.glancesgrabsensors = GlancesGrabSensors()
  File "/usr/local/lib/python2.7/dist-packages/glances/plugins/glances_sensors.py", line 245, in __init__
    self.sfans = psutil.sensors_fans()
  File "/usr/local/lib/python2.7/dist-packages/psutil/__init__.py", line 2298, in sensors_fans
    return _psplatform.sensors_fans()
  File "/usr/local/lib/python2.7/dist-packages/psutil/_pslinux.py", line 1193, in sensors_fans
    current = int(cat(base + '_input'))
  File "/usr/local/lib/python2.7/dist-packages/psutil/_pslinux.py", line 298, in cat
    with open_binary(fname) if binary else open_text(fname) as f:
  File "/usr/local/lib/python2.7/dist-packages/psutil/_pslinux.py", line 190, in open_binary
    return open(fname, "rb", **kwargs)
IOError: [Errno 2] No such file or directory: '/sys/class/hwmon/hwmon2/fan4_input'
```

Output in python:

	>>> psutil.sensors_fans()
	Traceback (most recent call last):
	File "<stdin>", line 1, in <module>
	File "/usr/local/lib/python2.7/dist-packages/psutil/__init__.py", line 2298, in sensors_fans
		return _psplatform.sensors_fans()
	File "/usr/local/lib/python2.7/dist-packages/psutil/_pslinux.py", line 1193, in sensors_fans
		current = int(cat(base + '_input'))
	File "/usr/local/lib/python2.7/dist-packages/psutil/_pslinux.py", line 298, in cat
		with open_binary(fname) if binary else open_text(fname) as f:
[...]

PR #1141, opened 2017-09-27
1129: have sensors_fans() on Linux skip entry on IOError

ssaip:
@giampaolo your commit for #1129 (943367f) fixed `sensors_temperatures()` but my problem was in `sensors_fans()` 

Cheers


giampaolo:
Oh you're right, my bad.

Issue #1127, opened 2017-09-14
Invalid management of reference counting in open_files (OSX)


jakub-bacic:
While iterating over file descriptors of the given process, it's possible to get `EPERM` error while calling `proc_pidfdinfo`:
https://github.com/giampaolo/psutil/blob/c4c6e22ca4dad687d2a8505f9e0207d8ecd64358/psutil/_psutil_osx.c#L1137-L1141

The errors checking part will raise exception and goto `error` label:
https://github.com/giampaolo/psutil/blob/c4c6e22ca4dad687d2a8505f9e0207d8ecd64358/psutil/_psutil_osx.c#L1151-L1152

If such situation happens after the first iteration of the loop (so `py_path` variable is not NULL), it will cause `py_path` reference count to be decremented even though it shouldn't be:
https://github.com/giampaolo/psutil/blob/c4c6e22ca4dad687d2a8505f9e0207d8ecd64358/psutil/_psutil_osx.c#L1180

I was able to discover that issue while testing psutil on Mac OS High Sierra beta release. Issue is consistently reproducible (at least on my machine) when I try to iterate over all processes and list open files for them. Example problematic process is called `lsd` and `EPERM` error is also visible while using command-line `lsof` utility:

```
hsierras-MacBook-Pro:psutil jbacic$ lsof -p 325
COMMAND PID   USER   FD   TYPE DEVICE   SIZE/OFF       NODE NAME
lsd     325 jbacic  cwd    DIR   1,15       1024          2 /
lsd     325 jbacic  txt    REG   1,15      19136 4298624763 /usr/libexec/lsd
lsd     325 jbacic  txt    REG   1,15      32768 4298764710 /private/var/db/mds/messages/502/se_SecurityMessages
lsd     325 jbacic  txt    REG   1,15     376816 4298543466 /System/Library/Frameworks/Security.framework/Versions/A/PlugIns/csparser.bundle/Contents/MacOS/csparser
lsd     325 jbacic  txt    REG   1,15   26704160 4298629438 /usr/share/icu/icudt59l.dat
lsd     325 jbacic  txt    REG   1,15    7331840 4298779644 /private/var/folders/35/wbfsx81n0wlfd1165gphvvw00000gp/0/com.apple.LaunchServices-219-v2.csstore
lsd     325 jbacic  txt    REG   1,15     837248 4298618806 /usr/lib/dyld
[...]

PR #1144, opened 2017-10-10
#1127 Fix reference counting bug in open_files impl on OSX

jakub-bacic:
I refactored the code using the following template:

```c
static PyObject *
example_function(PyObject *self, PyObject *args) {
    //1 declare return variable called py_ret in the first line
    PyObject *py_ret = NULL;
    
    //2 declare other python variables (with prefix py_) for easier management of reference counting
    PyObject *py_tuple = NULL;
    PyObject *py_path = NULL;

    //3 declare C variables for easier mapping malloc/free
    struct proc_fdinfo *fds_pointer = NULL;

    //4 other variables here

    //5 main code - in case of error, goto except block
    py_ret = PyList_New(0);
    if (!py_ret)
        goto except;

    // ...

    //6 finally
    goto finally;

except:
    Py_XDECREF(py_ret);
    py_ret = NULL;

finally:
    // call Py_XDECREF on every variable declared in section 2
    Py_XDECREF(py_tuple);
    Py_XDECREF(py_path);

    // call free on every variable declared in section 3 (calling free on NULL is no-op)
    free(fds_pointer);

    return py_ret;
}
```

Additionally, variables reused in a loop should be properly DECREFed and assigned NULL at the end of the loop (so the presented template works properly in case of flow interruption due to the error).

### DIFF
--- a/psutil/_psutil_osx.c
+++ b/psutil/_psutil_osx.c
@@ -1129,7 +1129,6 @@ psutil_proc_open_files(PyObject *self, PyObject *args) {
     iterations = (pidinfo_result / PROC_PIDLISTFD_SIZE);
 
     for (i = 0; i < iterations; i++) {
-        py_tuple = NULL;
         fdp_pointer = &fds_pointer[i];
 
         if (fdp_pointer->proc_fdtype == PROX_FDTYPE_VNODE) {
@@ -1167,7 +1166,9 @@ psutil_proc_open_files(PyObject *self, PyObject *args) {
             if (PyList_Append(py_retlist, py_tuple))
                 goto error;
             Py_DECREF(py_tuple);
+            py_tuple = NULL;
             Py_DECREF(py_path);
+            py_path = NULL;
             // --- /construct python list
         }
     }


giampaolo:
It seems to me like you're changing too much for no good reason. There's no need to redefine variable names, rename "error" goto to "except", introduce "finally" goto etc.
AFAICT the only change which is needed is this one:

```
              Py_DECREF(py_tuple);
 +            py_tuple = NULL;
              Py_DECREF(py_path);
 +            py_path = NULL;
 ```

Am I right?

**Comment by jakub-bacic** [comment]
Indeed, the minimal change needed to fix this particular issue will look like this.

I introduced more changes only to suggest a cleaner way of dealing with such C code. I assume you want to minimize the change scope as much as possible?

**Comment by giampaolo** [comment]
Correct, I prefer to minimize the change. Assigning NULL after DECREF is fine.

**Comment by jakub-bacic** [comment]
Example script to test (it should result in segfault - must be run on Mac OS High Sierra):
```python
import psutil

def test_open_files():
    try:
      proc.open_files()
    except psutil.Error:
      pass

test_open_files()
```

PR #1145, opened 2017-10-16
fix or skip remaining AIX unit tests

wiggin15:


psutil/tests/__init__.py
```diff
@@ -690,15 +690,6 @@ def create_exe(outpath, c_code=None):
...
             raise ValueError("gcc is not installed")
-        if c_code is None:
-            c_code = textwrap.dedent(
-                """
-                #include <unistd.h>
-                int main() {
-                    pause();
-                    return 1;
-                }
-                """)
```
giampaolo: Mmm why?

SethMichaelLarson: Unreachable. Compares `c_code` as a bool and then compares `c_code is None` in the same branch.

giampaolo: Oh I see. Yes, it is unreachable indeed. It should not be removed though. 
Please change it to:

```
if isinstance(c_code, bool):
    c_code = textwrap.dedent(
        """
        #include <unistd.h>
        int main() {
            pause();
            return 1;
        }
        """)

assert isinstance(c_code, str), c_code
```

wiggin15: From what I see `c_code` is either `None` (use python executable) or `str` (code to compile - used in one place). We're not using this "default" `c_code`, anyway.

giampaolo: The use case I want to support is `c_code=True`, as a way for telling "create an executable by using the C compiler". It is true this is currently not used (it was) but I want to keep it around, just in case.

wiggin15:
Updated pull request.

giampaolo:
Thanks Arnon.

PR #1161, opened 2017-10-28
Fix test asserts due to leftover test subprocesses

adpag:
Two cases where a test subprocess is not cleaned up after the test, leading to asserts in following tests.


giampaolo:
The missing `reap_children` is interesting. I get occasional failures on both Travis and Appveyor which I think may be related to this. Thanks.

PR #1164, opened 2017-10-30
AIX: implement num_ctx_switches

wiggin15:


psutil/arch/aix/num_ctx_switches.c
```diff
@@ -0,0 +1,45 @@
...
+            (long long) p->pi_ru.ru_nvcsw,    /* voluntary context switches */
+            (long long) p->pi_ru.ru_nivcsw);  /* involuntary */
+        free(processes);
+        return py_tuple;
+    }
+
+    /* finished iteration without finding requested pid */
+    free(processes);
+    return NoSuchProcess();
+}
```
giampaolo: Do you mind putting this in `_psutil_aix.c`? It's a short function, I don't think it's worth a new file just for it.

psutil/arch/aix/psutil_aix_common.c
```diff
@@ -0,0 +1,79 @@
+/*
```
giampaolo: I would just call this file `common.c`. The `aix` bit is redundant because you're already in `arch/aix` subdir. The `psutil` bit is only relevant in the `psutil` dir.

docs/index.rst
```diff
@@ -1376,8 +1376,6 @@ Process class
     The number voluntary and involuntary context switches performed by
     this process (cumulative).
 
-    Availability: all platforms except AIX
-
```
giampaolo: Please add `.. versionchanged:: 5.4.1 added AIX support`

psutil/arch/aix/net_connections.c
```diff
@@ -14,6 +14,8 @@
 */
 
 #include "net_connections.h"
+#include "psutil_aix_common.h"
+#include "../../_psutil_common.h"
```
giampaolo: Does it break anything if we move "local" includes like these ones after "standard" includes. I mean `#include <fcntl.h>` first (and others), one empty line, then "our" includes.

giampaolo (changes requested):
Cool! Glad to see this.

psutil/arch/aix/net_connections.c
wiggin15: Unfortunately switching the order does break compilation (redefinitions of macros between `pyconfig.h` which is included from `Python.h` with definitions from `/usr/include`. The latter does `ifndef`s and `undef`s so this order is ok).

wiggin15: I can make switch the order if I explicitly import `Python.h` first. I'll do that.

wiggin15:
Updated both commits.

PR #1167, opened 2017-11-08
Including non-unicast packets in packet count calculation

matray:
A relatively simple change to the windows net_io_counter code to include non-unicast packets in the packet count of the interface (both sending and receiving).


giampaolo:
Sweet! Thanks.

PR #1174, opened 2017-11-12
Move exceptions to separate file

wiggin15:
I'm not sure if there was a reason that the exceptions are not in a separate file. It looks simpler to import them instead of defining them as "None" in the platform modules and then overriding them from the main file, plus it cleans up the main file a bit.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -85,6 +85,12 @@ from ._common import POSIX  # NOQA
 from ._common import SUNOS
 from ._common import WINDOWS
 
+from ._exceptions import Error
+from ._exceptions import NoSuchProcess
+from ._exceptions import ZombieProcess
+from ._exceptions import AccessDenied
+from ._exceptions import TimeoutExpired
+
 if LINUX:
     # This is public API and it will be retrieved from _pslinux.py
     # via sys.modules.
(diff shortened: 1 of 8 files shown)


giampaolo:
Well, the reason may sound silly, but it's because of `repr`. I don't ilke the "_psutil.exceptions" prefix:

```    p = psutil.Process(2345)
  File "/home/giampaolo/svn/psutil/psutil/__init__.py", line 303, in __init__
    self._init(pid)
  File "/home/giampaolo/svn/psutil/psutil/__init__.py", line 343, in _init
    raise NoSuchProcess(pid, None, msg)
psutil._exceptions.NoSuchProcess: psutil.NoSuchProcess no process found with pid 2345
```

A user seeing that traceback will try to import NSP as:

```from psutil._exceptions import NoSuchProcess```

Yes, also `from psutil import NoSuchProcess` would work, but you know that only if you read psutil doc, and we would have 2 ways of doing the same thing, which goes against a Python zen.

wiggin15:
I updated "repr" to use hard-coded `psutil.` instead of `__module__.`. I still think this is a better way because the override is confusing, and AFAIK using a separate exceptions file is a common practice.

giampaolo:
Sweet! I remember trying doing something like this years ago but didn't succeeded for some reason. This way is definitively better.

PR #1177, opened 2017-11-19
OSX: implement sensors_battery

wiggin15:


psutil/_psutil_osx.c
```diff
@@ -1788,6 +1790,93 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    if (!power_info) {
+        PyErr_SetString(PyExc_RuntimeError,
+            "Failed to get power sources info");
+        goto error;
+    }
+
+    power_sources_list = IOPSCopyPowerSourcesList(power_info);
+    if (!power_sources_list) {
+        PyErr_SetString(PyExc_RuntimeError,
+            "Failed to get power sources list");
```
giampaolo: I would change the err message to "IOPSCopyPowerSourcesList() syscall failed"

psutil/_psutil_osx.c
```diff
@@ -1788,6 +1790,93 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    if (!CFNumberGetValue(time_to_empty_ref,
+                          kCFNumberIntType, &time_to_empty)) {
+        /* This value is recommended for non-Apple power sources, so it's not
+         * an error if it doesn't exist. We'll return -1 for "unknown" */
+        /* A value of -1 indicates "Still Calculating the Time" also for
+         * apple power source */
+        time_to_empty = -1;
+    }
+
+    py_retlist = Py_BuildValue("Iii",
```
giampaolo: I think ` Py_BuildValue("Iii"` produces a tuple, not a list.

psutil/_psutil_osx.c
```diff
@@ -1788,6 +1790,93 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    if (!power_sources_list) {
+        PyErr_SetString(PyExc_RuntimeError,
+            "Failed to get power sources list");
+        goto error;
+    }
+
+    /* Should only get one source. But in practice, check for > 0 sources */
+    if (!CFArrayGetCount(power_sources_list)) {
+        PyErr_SetString(PyExc_RuntimeError,
+            "Power sources list is empty");
```
giampaolo: I checkout out your branch and bumped into this error on my virtualized OSX box.
It seems this happens where there is not battery so we probably have to return None here.

giampaolo (changes requested):
Cool stuff. Some comments inline. Also another thing. In the future we'll probably want to make `sensors_battery` return info about multiple batteries as a list. Are these APIs able to do that?

Also, it would be good to have a specific test comparing results with a CLI tool. See here:
http://osxdaily.com/2015/12/10/get-mac-battery-life-info-command-line-os-x/

psutil/_psutil_osx.c
wiggin15: I'm following the style of the rest of psutil. the name `py_retlist` is used for tuples as well as lists everywhere in the code.
There is no variable currently named "py_rettuple" or "py_retobj" etc. What name would you suggest?

giampaolo: > I'm following the style of the rest of psutil. the name py_retlist is used for tuples as well as lists everywhere in the code.

Really? I didn't realize that. Anyway, I suggest `py_tuple` in this case.

psutil/_psutil_osx.c
```diff
@@ -1788,6 +1790,93 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    CFStringRef ps_state_ref = NULL;
+    uint32_t capacity;     /* units are percent */
+    int time_to_empty;     /* units are minutes */
+    int is_power_plugged;
+
+    power_info = IOPSCopyPowerSourcesInfo();
+
+    if (!power_info) {
+        PyErr_SetString(PyExc_RuntimeError,
+            "Failed to get power sources info");
```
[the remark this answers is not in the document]
wiggin15: I was following the style of "psutil_disk_io_counters" where errors for "IO*" functions set strings like "unable to get the disk properties" - but I will change this.

psutil/_psutil_osx.c
wiggin15: Already done :)

wiggin15:
> Are these APIs able to do that?

Since the API returns "PowerSourcesList", I assume this can include more than one source, but I will have no way to test this.

giampaolo:
> Since the API returns "PowerSourcesList", I assume this can include more than one source, but I will have no way to test this.

OK, let's ignore this for now. But just FYI my plan is to have:

`sensors_battery(all=True) -> [ntuple(...), ...]`

psutil/_psutil_osx.c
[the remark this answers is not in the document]
giampaolo: OK, I will change those ones then. In general I will probably want to apply this paradigm everywhere.

wiggin15:
Updated commit

psutil/_psosx.py
```diff
@@ -207,6 +207,29 @@ def disk_partitions(all=False):
...
+
+
+def sensors_battery():
+    """Return battery information.
+    """
+    try:
+        percent, minsleft, power_plugged = cext.sensors_battery()
+    except NotImplementedError:
+        # no power source - return None according to interface
+        return None
```
giampaolo: nice solution, I like it

psutil/tests/__init__.py
```diff
@@ -70,7 +70,7 @@
     'VERBOSITY',
     "HAS_CPU_AFFINITY", "HAS_CPU_FREQ", "HAS_ENVIRON", "HAS_PROC_IO_COUNTERS",
     "HAS_IONICE", "HAS_MEMORY_MAPS", "HAS_PROC_CPU_NUM", "HAS_RLIMIT",
-    "HAS_SENSORS_BATTERY", "HAS_BATTERY""HAS_SENSORS_FANS",
+    "HAS_SENSORS_BATTERY", "HAS_BATTERY", "HAS_SENSORS_FANS",
```
giampaolo: good catch

psutil/tests/test_osx.py
```diff
@@ -285,6 +286,18 @@ def test_net_if_stats(self):
...
+
+    @unittest.skipIf(not HAS_BATTERY, "no battery")
+    def test_sensors_battery(self):
+        out = sh("pmset -g batt")
+        percent = re.search("(\d+)%", out).group(1)
+        drawing_from = re.search("Now drawing from '([^']+)'", out).group(1)
+        power_plugged = drawing_from == "AC Power"
+        psutil_result = psutil.sensors_battery()
+        self.assertEqual(psutil_result.power_plugged, power_plugged)
+        self.assertEqual(psutil_result.percent, int(percent))
```
giampaolo: sweet!

Issue #1183, opened 2017-11-30
Process.children() is slow


pitrou:
```
>>> import psutil
>>> p = psutil.Process()
>>> p.children()
[]
>>> %timeit p.children()
36.1 ms ± 268 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

It seems half of that is calling `process_iter()`:
```
>>> %timeit l = list(psutil.process_iter())
20.5 ms ± 232 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

giampaolo:
Hey Antoine! =)
`Process.children()` is inevitably slow because it has to obtain `pid` and `ppid` for all processes:
https://github.com/giampaolo/psutil/blob/3a3598e433adec73e2a9c4d5f08e658516fb1d32/psutil/__init__.py#L862-L864
We can't get away from that.

With that said, it is true that we're doing more than just that because we also call `create_time()`, even more than once ([here](https://github.com/giampaolo/psutil/blob/3a3598e433adec73e2a9c4d5f08e658516fb1d32/psutil/__init__.py#L330) and [here](https://github.com/giampaolo/psutil/blob/3a3598e433adec73e2a9c4d5f08e658516fb1d32/psutil/__init__.py#L867)

We can probably avoid that and define a utility function similar to what we have on windows:
https://github.com/giampaolo/psutil/blob/3a3598e433adec73e2a9c4d5f08e658516fb1d32/psutil/__init__.py#L851-L854
...which just returns a `{pid; ppid, ...}` map for all processes in one shot.

pitrou:
Does that function have to be consistent? The problem is if the processes change while iterating on e.g. `/proc`.

giampaolo:
A process which "changes" while iterating means it will disappear, in which case we'll just skip it.
They will also be skipped in case of `AccessDenied` (doc doesn't mention this - I will file a ticket for it).

pitrou:
I see, so we can just scan quickly over `/proc`?

giampaolo:
Yes. Same goes for all other platforms except Windows which already does this [in C](https://github.com/giampaolo/psutil/blob/3a3598e433adec73e2a9c4d5f08e658516fb1d32/psutil/_psutil_windows.c#L2980).

giampaolo:
Addresed in #1185 and #1186. Closing.

PR #1184, opened 2017-11-30
Issue #1183: implement ppid_map for Linux

pitrou:
This makes Process.children() 3x faster when there are no children.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1356,6 +1356,24 @@ def pid_exists(pid):
             return pid in pids()
 
 
+def parse_stat_file(procfs_path, pid):
+    """Parse /proc/{pid}/stat file. Return a list of fields where
+    process name is in position 0.
+    Using "man proc" as a reference: where "man proc" refers to
+    position N, always substract 2 (e.g starttime pos 22 in
+    'man proc' == pos 20 in the list returned here).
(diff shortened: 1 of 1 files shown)


pitrou:
Note that Process.children(recursive=True) doesn't really benefit as it still creates a Process instance for every process on the system.

pitrou:
@giampaolo

pitrou:
I've got a patch to also make `children(recursive=True)`. Should I push it in this PR or create a separate PR for it?

giampaolo:
That won't work as it's Linux only. The change is much simpler: just define a `_ppid_map()` utility function in `__init__.py` which relies on the per-module `Process` class directly. That way all platforms can benefit from the speedup.

pitrou:
> The change is much simpler: just define a _ppid_map() utility function in __init__.py which relies on the per-module Process class directly.

Not sure what you mean by that? The point here is to avoid Process instantiations when we're only interested in a couple of them at the end.

giampaolo:
Instantiating "private" (e.g. look into _pslinux.py) `Process` classes should be cheap as they define `__slots__` (contrarily from the `Process` class in `__init__`).

giampaolo:
https://github.com/giampaolo/psutil/pull/1185

PR #1185, opened 2017-11-30
Speedup Process.children() 

giampaolo:
See #1183.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1367,6 +1367,8 @@ def ppid_map():
             with open_binary("%s/%s/stat" % (procfs_path, pid)) as f:
                 data = f.read()
         except EnvironmentError as err:
+            # Note: we should be able to access /stat for all processes
+            # so we won't bump into EPERM, which is good.
             if err.errno not in (errno.ENOENT, errno.ESRCH,
                                  errno.EPERM, errno.EACCES):
                 raise
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1356,6 +1356,27 @@ def pid_exists(pid):
             return pid in pids()
 
 
+def ppid_map():
+    """Obtain a {pid: ppid, ...} dict for all running processes in
+    one shot. Used to speed up Process.children().
+    """
+    ret = {}
+    procfs_path = get_procfs_path()
+    for pid in pids():
+        try:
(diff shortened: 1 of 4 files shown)


giampaolo:
```
BEFORE PATCH

~/svn/psutil {master}$ python -m timeit -s "import psutil; p = psutil.Process()" "p.children()"
100 loops, best of 3: 12.2 msec per loop

AFTER PATCH

~/svn/psutil {master}$ python -m timeit -s "import psutil; p = psutil.Process()" "p.children()"
100 loops, best of 3: 5.49 msec per loop
```

pitrou:
Here are the numbers I get:
* git master:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children())"
10 loops, best of 3: 35.3 msec per loop
```
* this PR:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children())"
100 loops, best of 3: 15.5 msec per loop
```
* PR #1184:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children())"
100 loops, best of 3: 13.6 msec per loop
```

So my approach is slightly faster, at the expense of being Linux-only. Of course, one could argue that WIndows and Linux are the dominant platforms ;-)

giampaolo:
> Of course, one could argue that WIndows and Linux are the dominant platforms ;-)

Makes sense =)
Here's a variation of your patch: https://github.com/giampaolo/psutil/pull/1185/commits/78f8cbcf0a93b2b075803ad9578bc317ce599fd7.

psutil/_pslinux.py
```diff
@@ -1356,6 +1356,27 @@ def pid_exists(pid):
...
+    one shot. Used to speed up Process.children().
+    """
+    ret = {}
+    procfs_path = get_procfs_path()
+    for pid in pids():
+        try:
+            with open_binary("%s/%s/stat" % (procfs_path, pid)) as f:
+                data = f.read()
+        except EnvironmentError as err:
+            if err.errno not in (errno.ENOENT, errno.EPERM, errno.EACCES):
```
pitrou: Apparently also `ESRCH` (according to `wrap_exceptions`)?

giampaolo:
Merged. Thanks for reporting this. I had no idea such a speedup could be possible.

PR #1186, opened 2017-12-01
Faster Process.children(recursive=True)

pitrou:
Before:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children(recursive=True))"
10 loops, best of 3: 29.6 msec per loop
```

After:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children(recursive=True))"
100 loops, best of 3: 12.4 msec per loop
```

For reference, non-recursive:
```
$ python -m timeit -s "import psutil; p = psutil.Process()" "list(p.children())"
100 loops, best of 3: 12.2 msec per loop
```

Issue #1183.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -888,33 +888,33 @@ class Process(object):
                     except (NoSuchProcess, ZombieProcess):
                         pass
         else:
-            # construct a dict where 'values' are all the processes
-            # having 'key' as their parent
-            table = collections.defaultdict(list)
+            # Construct a {pid: [child pids]} dict
+            reverse_ppid_map = collections.defaultdict(list)
             for pid, ppid in ppid_map.items():
-                try:
-                    p = Process(pid)
-                    table[ppid].append(p)
-                except (NoSuchProcess, ZombieProcess):
-                    pass
-            # At this point we have a mapping table where table[self.pid]
-            # are the current process' children.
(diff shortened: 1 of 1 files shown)


pitrou:
@giampaolo

pitrou:
Basically the idea is that instead of calling `Process` for all processes on the system and then selecting those that are candidate (grand)children of `self`, we prune the search space and only call `Process` for candidate (grand)children.

pitrou:
(there seems to be an unrelated failure on AppVeyor)

psutil/__init__.py
```diff
@@ -888,33 +888,33 @@ def children(self, recursive=False):
...
-                for child in table[pid]:
+                reverse_ppid_map[ppid].append(pid)
+            # Recursively traverse that dict, starting from self.pid,
+            # such that we only call Process() on actual children
+            seen = set()
+            stack = [self.pid]
+            while stack:
+                pid = stack.pop()
+                if pid in seen:
+                    # Since pids can be reused while the ppid_map is
```
pitrou: Note I'm not sure this check is necessary, as the `create_time` test may be sufficient to avoid such potential cycles.

giampaolo:
Thanks, this is awesome. I'll try to take a look at this over the week end.

giampaolo:
Looks good. Merging.

Issue #775, opened 2016-02-23
Windows - psutil does not detect mounted folders without drive letters as disks


MattGirolami:
Windows supports assigning NTFS disks a mounted folder/mount point value instead of (or in addition to) a drive letter. Psutil does not recognize drives that only have a mount point and not a drive letter.

In the example below, I've attached and formatted a hard drive, but instead of assigning it it's own drive letter, I've assigned it a mount point path of C:\mountedfolder\ . I can iterate the drive in powershell/wmi:

```
C:\>  Get-WmiObject -namespace "root/cimv2" -query "SELECT Name, Capacity, FreeSpace, deviceid FROM Win32_Volume WHERE DriveType = 3" | select deviceid,name,capacity | format-list


deviceid : \\?\Volume{343e9a8f-c1de-11e4-b98f-806e6f6e6963}\
name     : \\?\Volume{343e9a8f-c1de-11e4-b98f-806e6f6e6963}\
capacity : 523235328

deviceid : \\?\Volume{07ec83a6-0000-0000-0000-010000000000}\
name     : C:\mountedfolder\
capacity : 1070592000

deviceid : \\?\Volume{58dafacc-c28b-11e4-95bb-3863bbba0e0f}\
name     : C:\
capacity : 253377900544

deviceid : \\?\Volume{25083962-0000-0000-0000-c01d3b000000}\
name     : D:\
capacity : 2148532224`
```

If I attempt to iterate the disks in psutil, I do not see my disk at C:\mountedfolder\

```
C:\> py
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:54:25) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> psutil.disk_partitions()
[sdiskpart(device='C:\\', mountpoint='C:\\', fstype='NTFS', opts='rw,fixed'), 
 sdiskpart(device='D:\\', mountpoint='D:\\', fstype='FAT32', opts='rw,fixed')]
>>>
```

If I assign a drive letter ( like E:) to the same disk at c:\mountedfolder\ I can list it in powershell 

```
C:\>  Get-WmiObject -namespace "root/cimv2" -query "SELECT Name, Capacity, FreeSpace, deviceid FROM Win32_Volume WHERE DriveType = 3" | select deviceid,name,capacity | format-list


[...]

PR #1192, opened 2017-12-11
Add mount points to disk_partitions() in Windows (#775)

jomann09:
So after taking a second look I think this will work. I did not end up doing a utility function since it's barely repeated anymore - I fixed the way it was looping. I am also breaking out of the loop on error instead of just jumping right down to error, allowing it to finish up with the volume that it is currently working on.

Also, shows mount points without `all=True` to match Linux implementation.

### DIFF
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -2599,7 +2599,8 @@ psutil_disk_partitions(PyObject *self, PyObject *args) {
                             opts);
 
                         if (!py_tuple || PyList_Append(py_retlist, py_tuple) == -1) {
-                            break;
+                            FindVolumeMountPointClose(mp_h);
+                            goto error;
                         }
 
                         Py_DECREF(py_tuple);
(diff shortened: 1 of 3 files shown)


psutil/_psutil_windows.c
```diff
@@ -2573,6 +2578,39 @@ psutil_disk_partitions(PyObject *self, PyObject *args) {
...
+                        py_tuple = Py_BuildValue(
+                            "(ssss)",
+                            drive_letter,
+                            mp_path,
+                            fs_type, // Typically NTFS
+                            opts);
+
+                        if (!py_tuple || PyList_Append(py_retlist, py_tuple) == -1) {
+                            break;
+                        }
```
giampaolo: Aren't you supposed to `goto error;` here?

jomann09: I was unsure about doing the `goto error;` because it would skip the rest of the volume's functionality (like adding the actual volume) but if you think it's fine to exit there, I can add that there.

giampaolo: It's correct to skip: errors should not pass unnoticed.

giampaolo: s/correct/not correct

jomann09: Updated with `goto error;` and closed the handler before going.

PR #1207, opened 2018-01-09
Include license file in the generated wheel packages

jdufresne:
The wheel package format supports including the license file. This is
done using the [metadata] section in the setup.cfg file. For additional
information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

Helps satisfy the project's license:

> Redistributions in binary form must reproduce the above copyright
> notice, this list of conditions and the following disclaimer in the
> documentation and/or other materials provided with the distribution.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[metadata]
+license_file = LICENSE


giampaolo:
Isn't there a way to do this without adding another file? The root dir is already pretty crowded.

jdufresne:
I'm not aware of any other method.

Are you perhaps thinking of `MANIFEST.in`? That file does not affect wheels, only source distributions. 

If you have some other suggestion, let me know, but I don't think one exists.

PR #1208, opened 2018-01-11
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -338,6 +338,7 @@ def main():
     )
     if setuptools is not None:
         kwargs.update(
+            python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
             test_suite="psutil.tests.get_suite",
             tests_require=tests_require,
             extras_require=extras_require,


giampaolo:
Mmm interesting. What happens if, say, I try to install psutil with Python 3.2?

jdufresne:
If you're using a modern version of pip, pip will install the highest version that satisfies the `python_requires` marker. As an example, the most recent [Pillow](https://github.com/python-pillow/Pillow) version 5.0 does not support Python 3.3. But observe the behave when I install with Python 3.3:

```
$ virtualenv --python=python3.3 reqtest
Running virtualenv with interpreter /usr/bin/python3.3
Using base prefix '/usr'
New python executable in /home/jon/reqtest/bin/python3.3
Not overwriting existing python script /home/jon/reqtest/bin/python (you must use /home/jon/reqtest/bin/python3.3)
Installing setuptools, pip, wheel...done.

$ . reqtest/bin/activate

(reqtest) $ pip install Pillow
Collecting Pillow
Collecting olefile (from Pillow)
Installing collected packages: olefile, Pillow
Successfully installed Pillow-4.3.0 olefile-0.44
```

As Pillow 5.0 isn't compatible with the Python version, pip found the next available version that is.

This will be important when psutil drops Python 2.6 and other EOL support. Those users can continue to get a usable version of psutil with the command `pip install psutil` without the need to pin to a specific version. It'll just work (provided a modern version of pip).

giampaolo:
Clear. This is extremely useful then, and will make 2.6 support removal less painful. Thanks.

giampaolo:
Actually, while we're at it, isn't this the same?

```
python_requires=">=2.6, >=3.3.*",
```

If so I prefer, 'cause it's shorter and cleaner.

PR #1213, opened 2018-01-30
24line can't excute

whoamishuaige:
my env is centos7.0,python2.7,psutil2.0.0，that couldn't excute 24line.so i change this line as this

### DIFF
--- a/scripts/who.py
+++ b/scripts/who.py
@@ -21,7 +21,10 @@ import psutil
 def main():
     users = psutil.users()
     for user in users:
-        proc_name = psutil.Process(user.pid).name() if user.pid else ""
+        if hasattr(user,'pid'):
+                proc_name = psutil.Process(user.pid).name()
+        else:
+                proc_name = ""
         print("%-12s %-10s %-10s %-14s %s" % (
             user.name,
             user.terminal or '-',


giampaolo:
I don't understand the difference. This:
```
        proc_name = psutil.Process(user.pid).name() if user.pid else ""
```
is supposed to do the same thing no? (get name() if user.pid is true)

giampaolo:
Actually I realize that if user.pid is 0 for some reason that won't work.

giampaolo:
Anyway, can you paste the traceback? Are you getting an `AttributeError`?

whoamishuaige:
when i run python who.py,i get a AttributeError like this AttributeError: 'suser' object has no attribute 'pid'.so i want to know the what is python,for that i enter python terminal,excute the command like this import psutil;
psutil.users()
then i got result
[suser(name='root', terminal='pts/2', host='220.249.22.170', started=1517387776.0)]
so that pid is undefined in psutil.users object.my env is python 2.7.5,psutil is 2.0.0,os is centos7.x.

giampaolo:
psutil 2.0.0 is ancient! You should upgrade it (we're at 5.4.3).

whoamishuaige:
thank you,sloving this problem for me

giampaolo:
To be precise:  =)
psutil-0.2.0: 2010-11-13
https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#020

Issue #1210, opened 2018-01-17
CPU steal stuck at 100%


vfuse:
Occasionally psutil returns a CPU steal time of 100%, the only way to get this back to the correct value is by rebooting the system.

```
cpu:
{
"0": {
"guest": 0.0, 
"guest_nice": 0.0, 
"idle": 0.0, 
"iowait": 0.0, 
"irq": 0.0, 
"nice": 0.0, 
"softirq": 0.0, 
"steal": 100.0, 
"system": 0.0, 
"user": 0.0
}, 
"1": {
"guest": 0.0, 
"guest_nice": 0.0, 
"idle": 0.0, 
"iowait": 0.0, 
"irq": 0.0, 
"nice": 0.0, 
"softirq": 0.0, 
"steal": 100.0, 
"system": 0.0, 
"user": 0.0
}
}
```
```

top - 10:25:55 up 46 days, 20:48, 1 user, load average: 0,34, 0,19, 0,15
Tasks: 120 total, 1 running, 119 sleeping, 0 stopped, 0 zombie
%Cpu(s): 0,0 us, 0,0 sy, 0,0 ni, 99,7 id, 0,3 wa, 0,0 hi, 0,0 si, 0,0 st
KiB Mem : 8173956 total, 496288 free, 6969612 used, 708056 buff/cache
KiB Swap: 2097148 total, 349612 free, 1747536 used. 894700 avail Mem

```
Running version 5.4.3 on AWS Ubuntu 16.04 xenial.

lbguilherme:
I currently own the vps affected by this problem, in case something need to be tested on the machine itself.

giampaolo:
Mmm can you try this and paste the output?

```
import time, psutil
for x in range(10):
      print(psutil.cpu_times())
      time.sleep(1)
```

lbguilherme:
```
$ python
Python 2.7.12 (default, Nov 20 2017, 18:23:56) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import time, psutil
>>> for x in range(10):
...       print(psutil.cpu_times())
...       time.sleep(1)
... 
scputimes(user=316284.26, nice=144.61, system=64877.16, idle=7577169.14, iowait=16110.07, irq=0.0, softirq=6806.69, steal=14416055402.18, guest=0.0, guest_nice=0.0)
[...]

PR #1214, opened 2018-02-06
Ignore negative deltas in cpu times when calculating percentages

wiggin15:
See discussion in  #1210

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1655,6 +1655,27 @@ def _cpu_busy_time(times):
     return busy
 
 
+def _cpu_times_deltas(t1, t2):
+    assert t1._fields == t2._fields, (t1, t2)
+    field_deltas = []
+    for field in _psplatform.scputimes._fields:
+        field_delta = getattr(t2, field) - getattr(t1, field)
+        # CPU times are always supposed to increase over time
+        # or at least remain the same and that's because time
+        # cannot go backwards.
+        # Surprisingly sometimes this might not be the case (at
+        # least on Windows and Linux), see:
+        # https://github.com/giampaolo/psutil/issues/392
+        # https://github.com/giampaolo/psutil/issues/645
+        # https://github.com/giampaolo/psutil/issues/1210
+        # Trim negative deltas to zero to ignore decreasing fields.
+        # top does the same. Reference:
+        # https://gitlab.com/procps-ng/procps/blob/v3.3.12/top/top.c#L5063
+        field_delta = max(0, field_delta)
+        field_deltas.append(field_delta)
+    return _psplatform.scputimes(*field_deltas)
+
+
 def cpu_percent(interval=None, percpu=False):
     """Return a float representing the current system-wide CPU
     utilization as a percentage.
@@ -1697,18 +1718,11 @@ def cpu_percent(interval=None, percpu=False):
         raise ValueError("interval is not positive (got %r)" % interval)
 
     def calculate(t1, t2):
-        t1_all = _cpu_tot_time(t1)
-        t1_busy = _cpu_busy_time(t1)
+        times_delta = _cpu_times_deltas(t1, t2)
 
-        t2_all = _cpu_tot_time(t2)
-        t2_busy = _cpu_busy_time(t2)
+        all_delta = _cpu_tot_time(times_delta)
+        busy_delta = _cpu_busy_time(times_delta)
 
-        # this usually indicates a float precision issue
-        if t2_busy <= t1_busy:
-            return 0.0
-
-        busy_delta = t2_busy - t1_busy
-        all_delta = t2_all - t1_all
         try:
             busy_perc = (busy_delta / all_delta) * 100
         except ZeroDivisionError:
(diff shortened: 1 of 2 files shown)


giampaolo:
Uhm... scary (because these are very old critical parts). =) 
Do you think you can add a test case which simulates this issue?
If it's too much effort I can do it for you.

wiggin15:
Yes, this is a little scary :grimacing:. A careful review of the logic is needed.
I added a test and checked that it fails before the fix.

giampaolo:
I don't want to rush this. I will take a deep look at this when I have more free time (may be weeks or months).

giampaolo:
I modified your unit tests so that it uses mock instead of playing with PROCFS_PATH:

```python
    def test_cpu_steal_decrease(self):
        # Test cumulative cpu stats decrease. We should ignore this.
        # See issue #1210.

        def open_mock(name, *args, **kwargs):
            if name == "/proc/stat":
                return io.BytesIO(textwrap.dedent("""\
                    cpu   0 0 0 0 0 0 0 1 0 0
                    cpu0  0 0 0 0 0 0 0 1 0 0
                    cpu1  0 0 0 0 0 0 0 1 0 0
                    """).encode())
            else:
                return orig_open(name, *args, **kwargs)

        orig_open = open
        patch_point = 'builtins.open' if PY3 else '__builtin__.open'

        with mock.patch(patch_point, create=True, side_effect=open_mock) as m:
            # first call to "percent" functions should read the new stat file
            # and compare to the "real" file read at import time - so the
            # values are meaningless
            psutil.cpu_percent()
            assert m.called
            psutil.cpu_percent(percpu=True)
            psutil.cpu_times_percent()
            psutil.cpu_times_percent(percpu=True)

        def open_mock(name, *args, **kwargs):
            if name == "/proc/stat":
                return io.BytesIO(textwrap.dedent("""\
                    cpu   1 0 0 0 0 0 0 0 0 0
                    cpu0  1 0 0 0 0 0 0 0 0 0
                    cpu1  1 0 0 0 0 0 0 0 0 0
                    """).encode())
            else:
                return orig_open(name, *args, **kwargs)

        with mock.patch(patch_point, create=True, side_effect=open_mock) as m:
            # Increase "user" while steal goes "backwards" to zero.
            cpu_percent = psutil.cpu_percent()
            assert m.called
            cpu_percent_percpu = psutil.cpu_percent(percpu=True)
            cpu_times_percent = psutil.cpu_times_percent()
            cpu_times_percent_percpu = psutil.cpu_times_percent(percpu=True)
            self.assertNotEqual(cpu_percent, 0)
            self.assertNotEqual(sum(cpu_percent_percpu), 0)
            self.assertNotEqual(sum(cpu_times_percent), 0)
            self.assertNotEqual(sum(cpu_times_percent), 100.0)
            self.assertNotEqual(sum(map(sum, cpu_times_percent_percpu)), 0)
            self.assertNotEqual(sum(map(sum, cpu_times_percent_percpu)), 100.0)
            self.assertEqual(cpu_times_percent.steal, 0)
            self.assertNotEqual(cpu_times_percent.user, 0)
```

psutil/__init__.py
```diff
@@ -1698,18 +1718,11 @@ def cpu_percent(interval=None, percpu=False):
...
 
     def calculate(t1, t2):
-        t1_all = _cpu_tot_time(t1)
-        t1_busy = _cpu_busy_time(t1)
+        times_delta = _cpu_times_deltas(t1, t2)
 
-        t2_all = _cpu_tot_time(t2)
-        t2_busy = _cpu_busy_time(t2)
+        all_delta = _cpu_tot_time(times_delta)
+        busy_delta = _cpu_busy_time(times_delta)
```
giampaolo: This I understand and it looks correct. It's also better than the previous version.

psutil/__init__.py
```diff
@@ -1778,28 +1791,14 @@ def cpu_times_percent(interval=None, percpu=False):
...
-        all_delta = _cpu_tot_time(t2) - _cpu_tot_time(t1)
-        for field in t1._fields:
-            field_delta = getattr(t2, field) - getattr(t1, field)
-            try:
-                field_perc = (100 * field_delta) / all_delta
-            except ZeroDivisionError:
-                field_perc = 0.0
+        times_delta = _cpu_times_deltas(t1, t2)
+        all_delta = _cpu_tot_time(times_delta)
+        scale = 100.0 / max(1, all_delta)
```
giampaolo: I'm having troubles with understanding this part. Can you add a comment explaining what it does?

psutil/__init__.py
```diff
@@ -1656,6 +1656,26 @@ def _cpu_busy_time(times):
     return busy
 
 
+def _cpu_times_deltas(t1, t2):
```
giampaolo: It wouldn't be bad to add:
```python
assert t1._fields == t2._fields, (t1, t2)
```

psutil/__init__.py
wiggin15: I just moved the calculation of `100 / all_delta` that was inside the loop before to outside of the loop. So `(100 * field_delta) / all_delta` becomes `field_delta * scale` where `scale` is "pre-calculated". The `max` bit is to avoid ZeroDivisionError in the case of `all_delta=0`. In this case `field_delta` will also be 0 so the percentages will zero out anyway. I took this from `top.c` (https://gitlab.com/procps-ng/procps/blob/v3.3.12/top/top.c#L5083) but I can revert to the old logic if you prefer.

wiggin15:
Thanks for the review. I will update the unit tests, add the assert and try to write a comment for the `scale` variable :)

giampaolo:
Thanks. This is great BTW. That logic has been "uncertain" for a long time, as you may have seen from my code comments.

wiggin15:
Updated commit

PR #1220, opened 2018-02-17
Try to extract argv[] from process address space first

alxchk:
Bump https://github.com/giampaolo/psutil/issues/1092

### DIFF
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -129,7 +129,7 @@ cstrings_array_to_string(char ** array, size_t count, char dm) {
     char *last = NULL;
 
     if (!array)
-        return NULL;
+        return strdup("");
 
     for (i=0; i<count; i++) {
         if (!array[i])
@@ -144,8 +144,10 @@ cstrings_array_to_string(char ** array, size_t count, char dm) {
     }
 
     result = malloc(total_length);
-    if (!result)
+    if (!result) {
+        PyErr_NoMemory();
         return NULL;
+    }
 
     result[0] = '\0';
     last = result;
@@ -191,6 +193,8 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
     if (!py_name)
         goto error;
 
+    /* SunOS truncates arguments to length PRARGSZ, very likely args are truncated.
+     * The only way to retrieve full command line is to parse process memory */
     if (info.pr_argc && strlen(info.pr_psargs) == PRARGSZ-1) {
         argv = psutil_read_raw_args(info, procfs_path, &argc);
         if (argv) {
@@ -198,18 +202,26 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
             if (argv_plain) {
                 py_args = PyUnicode_DecodeFSDefault(argv_plain);
                 free(argv_plain);
-            }
+            } else
+                goto error;
+
             psutil_free_cstrings_array(argv, argc);
         }
     }
 
+    /* If we can't read process memory or can't decode the result
+     * then return args from /proc. */
     if (!py_args)
         py_args = PyUnicode_DecodeFSDefault(info.pr_psargs);
+
+    /* Both methods has been failed. */
     if (!py_args)
         goto error;
+
     py_retlist = Py_BuildValue("OO", py_name, py_args);
     if (!py_retlist)
         goto error;
+
     Py_DECREF(py_name);
     Py_DECREF(py_args);
     return py_retlist;
--- a/psutil/_psutil_sunos.c
+++ b/psutil/_psutil_sunos.c
@@ -116,6 +116,54 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
         );
 }
 
+/*
+ * Join array of C strings to C string with delemiter dm.
+ * Omit empty records.
+ */
+static char*
+cstrings_array_to_string(char ** array, size_t count, char dm) {
+    int i;
+    size_t total_length = 0;
+    size_t item_length = 0;
+    char *result = NULL;
+    char *last = NULL;
+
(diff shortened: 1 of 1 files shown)


alxchk:
Passed arguments which you can find from proc is truncates in sunos. So if you have privileges you can parse them out from process memory (this is how their standard toolz are doing). This was in PR for environment code, but after split never get ro the master branch

psutil/_psutil_sunos.c
```diff
@@ -116,6 +116,54 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
...
+        total_length += item_length;
+    }
+
+    if (!total_length) {
+        return strdup("");
+    }
+
+    result = malloc(total_length);
+    if (!result)
+        return NULL;
```
giampaolo: You should `PyErr_NoMemory` here

psutil/_psutil_sunos.c
```diff
@@ -139,7 +190,21 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
...
+    if (info.pr_argc && strlen(info.pr_psargs) == PRARGSZ-1) {
+        argv = psutil_read_raw_args(info, procfs_path, &argc);
+        if (argv) {
+            argv_plain = cstrings_array_to_string(argv, argc, ' ');
+            if (argv_plain) {
+                py_args = PyUnicode_DecodeFSDefault(argv_plain);
+                free(argv_plain);
+            }
+            psutil_free_cstrings_array(argv, argc);
+        }
```
giampaolo: There should be an `else` here in case `psutil_read_raw_args` fails I guess.

psutil/_psutil_sunos.c
```diff
@@ -139,7 +190,21 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
...
     if (!py_name)
         goto error;
-    py_args = PyUnicode_DecodeFSDefault(info.pr_psargs);
+
+    if (info.pr_argc && strlen(info.pr_psargs) == PRARGSZ-1) {
+        argv = psutil_read_raw_args(info, procfs_path, &argc);
+        if (argv) {
+            argv_plain = cstrings_array_to_string(argv, argc, ' ');
+            if (argv_plain) {
+                py_args = PyUnicode_DecodeFSDefault(argv_plain);
```
giampaolo: You should check return code and raise exception if necessary

psutil/_psutil_sunos.c
```diff
@@ -139,7 +190,21 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
     py_name = PyUnicode_DecodeFSDefault(info.pr_fname);
     if (!py_name)
         goto error;
-    py_args = PyUnicode_DecodeFSDefault(info.pr_psargs);
+
```
giampaolo: Please add a comment explaining what this section is for, e.g. that you first attempt reading process memory then parse /proc which may truncate args.

alxchk:
Bump

psutil/_psutil_sunos.c
```diff
@@ -139,12 +190,37 @@ psutil_proc_name_and_args(PyObject *self, PyObject *args) {
...
+     * The only way to retrieve full command line is to parse process memory */
+    if (info.pr_argc && strlen(info.pr_psargs) == PRARGSZ-1) {
+        argv = psutil_read_raw_args(info, procfs_path, &argc);
+        if (argv) {
+            argv_plain = cstrings_array_to_string(argv, argc, ' ');
+            if (argv_plain) {
+                py_args = PyUnicode_DecodeFSDefault(argv_plain);
+                free(argv_plain);
+            } else {
+                PyErr_NoMemory();
```
giampaolo: You should `PyErr_NoMemory` from `cstrings_array_to_string`, not here

alxchk: Well, I can. But isn't it better to leave the utility function clean? Right now it returns NULL only in one case (let's say like strcat), and I handle this case from function which works with python objects.
If you don't like I will change this of course

giampaolo: The rationale is that if the utility function changes and need to raise something different than `PyErr_NoMemory` then the caller has no way to know what to do

psutil/_psutil_sunos.c
```diff
@@ -116,6 +116,58 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
...
+static int
+cstrings_array_to_string(char **joined, char ** array, size_t count, char dm) {
+    int i;
+    size_t total_length = 0;
+    size_t item_length = 0;
+    char *result = NULL;
+    char *last = NULL;
+
+    if (!array || !joined)
+        return 0;
```
giampaolo: You should set a proper Python exception here. Later on, when you invoke this function, you arbitrarily `goto error` but I think that this way it will generate a void "python exception not set" error. I think something like this should be good:
```
PyErr_SetString(PyExc_ValueError, "array size mismatch");
```

psutil/_psutil_sunos.c
```diff
@@ -116,6 +116,58 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
...
+    for (i=0; i<count; i++) {
+        if (!array[i])
+            continue;
+
+        item_length = strlen(array[i]) + 1;
+        total_length += item_length;
+    }
+
+    if (!total_length) {
+        return 0;
```
giampaolo: same here (raise proper exception - I'd say ValueError with a proper message set)

giampaolo (changes requested):
Sorry for the delay. I added a couple of inline comments then it's good to go.

psutil/_psutil_sunos.c
giampaolo: I see. Fine then.

Issue #1223, opened 2018-02-20
boot_time() returns incorrect value on XP


xiaolingbao:
Build _pswindows.pyd with _WIN32_WINNT set to '0x501' and PSAPI_VERSION=1, and run:

c:\window> python -c "import psutil; print psutil.boot_time()"
4.74414813e+18

Cause:
uptime is defined as double with the above settings (https://github.com/giampaolo/psutil/blob/dd19ba7b5bbf5ea67a4d386baa3fabd13b118533/psutil/_psutil_windows.c#L216), Py_BuildValue() later on expects either ULONGLONG or LONGLONG. There is memory layout mismatch between the expected value format and the supplied.

This is a regression as the previous version uses double in both places.

giampaolo:
Can you submit a PR?

PR #1225, opened 2018-02-20
Make uptime type consitent to fix boot time error.

xiaolingbao:
https://github.com/giampaolo/psutil/issues/1223


giampaolo:
The downside of this is that psutil cannot be compiled on Windows < Vista. Can you leave the pre-processor stuff in place?

xiaolingbao:
Thanks for your comment. Sorry I don't have system < Vista at the moment and I cannot see where is the offending line in the change. Can you please paste the compile error if possible?

giampaolo:
I also don't have < Vista but last time I checked (maybe 1 year ago) psutil could be compiled from sources on Windows XP. If possible I would like to keep it that way.

xiaolingbao:
I reviewed the code several times and I think it should still compile, unless I missed something :)
The conditional compile seems unnecessary. I'll try to find a system < vista to try later. Thank you again.

giampaolo:
`ULONGLONG` does not exist on < Vista so as it stands it should not compile.

PR #1228, opened 2018-03-04
Remove additional workarounds and support for Python 3.2 & 3.3

jdufresne:
Support for Python 3.3 was dropped in version 5.4.1. Support for Python 3.2 was dropped earlier. Remove all references to these unsupported versions including documentation, scripts, workarounds, etc. Eases maintenance as fewer workarounds are used for unsupported environments.


giampaolo:
Cool stuff! Thanks.

psutil/tests/__main__.py
```diff
@@ -32,8 +32,6 @@
     TEST_DEPS.extend(["ipaddress", "unittest2", "argparse", "mock==1.0.1"])
 elif sys.version_info[:2] == (2, 7) or sys.version_info[:2] <= (3, 2):
```
hugovk: That `or X` looks like a bug, it means this statement is also true for Python 2.6.

Python 2.6 then has both `mock==1.0.1` and `mock` in `TEST_DEPS`.

See PR #1237.

PR #1235, opened 2018-03-06
In tox, move flake8 to own target

jdufresne:
Allows running static analysis (lint/flake8) separately from testing the library. Reduces the redundant times static analysis runs. Previously ran after every test environment, but only needs to run once, requiring less tests resources.

For lint, use "skip_install = True" as installing psutil isn't required to do static analysis. For details on this configuration, see:

https://tox.readthedocs.io/en/latest/config.html#confval-skip_install=BOOL

> skip_install=BOOL
>
> Do not install the current package. This can be used when you need the virtualenv management but do not want to install the current package into that environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,11 +5,10 @@
 # directory.
 
 [tox]
-envlist = py26, py27, py34, py35, py36
+envlist = py26, py27, py34, py35, py36, lint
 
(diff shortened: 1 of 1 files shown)


giampaolo:
Cool stuff. Thanks.

PR #1237, opened 2018-03-08
Simplify Python version checks, fix a bug

hugovk:
The one in psutil/tests/\_\_main__.py looks like an unintentional bug:

```python
if sys.version_info[:2] == (2, 6):
    TEST_DEPS.extend(["ipaddress", "unittest2", "argparse", "mock==1.0.1"])
elif sys.version_info[:2] == (2, 7) or sys.version_info[:2] <= (3, 2):
    TEST_DEPS.extend(["ipaddress", "mock"])
```

That second `elif` is also true for Python 2.6 because `or sys.version_info[:2] <= (3, 2)` is always true.

Python 2.6 then has both `mock==1.0.1` and `mock` in `TEST_DEPS`.


giampaolo:
setup.py just changed and this is no longer necessary. Thanks though.

PR #1239, opened 2018-03-09
Adding 'slab' info to virtual_memory

hush-hush:
On linux 'Slab' is available from `/proc/meminfo`.


giampaolo:
There are 48 memory stats in /proc/meminfo:
```
$ cat /proc/meminfo | wc -l
48
```
Why is slab memory important?

hush-hush:
Hi @giampaolo,

`slab` memory allocation reflects kernel cache utilization (usefull to detect kernel memory leaks for example). That metric is very valuable to kernel developers, not so much to users indeed.

Also filesystem caching is reflected in the slab since `inodes` and `dentries` are cached too.

PR #1241, opened 2018-03-11
Remove executable permission from tests

jdufresne:
Tests don't require executable permission. The test files themselves are not started as a program.


giampaolo:
Tests are independent scripts which can be run individually, hence why I set the X bit. I see them as the files we have in `scripts/` directory.

jdufresne:
I see. That makes sense. Thanks.

PR #1243, opened 2018-03-12
cpu_times() GetSystemTimes function double dword convert to __int64,t…

stswandering:
cpu_times() GetSystemTimes function double dword convert to __int64,then double,instead of float.

### DIFF
--- a/psutil/_psutil_windows.c
+++ b/psutil/_psutil_windows.c
@@ -51,8 +51,8 @@
 
 #define MALLOC(x) HeapAlloc(GetProcessHeap(), 0, (x))
 #define FREE(x) HeapFree(GetProcessHeap(), 0, (x))
-//#define LO_T ((float)1e-7)
-//#define HI_T (LO_T*4294967296.0)
+#define LO_T 1e-7
+#define HI_T 429.4967296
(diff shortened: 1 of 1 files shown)


giampaolo:
So, I took those "weird" numbers from cPython source code (`os.times()` function):
https://github.com/python/cpython/blob/3b20d3454e8082e07dba93617793de5dc9237828/Modules/posixmodule.c#L7734
As such, I would kinda feel safer if we leave them in place instead of introducing a new conversion logic.
What happens if simply cast the numbers to `double`?
Something like (pseudo code):

```c
        user = (float)((HI_T * sppi[i].UserTime.HighPart) +
                       (LO_T * sppi[i].UserTime.LowPart));
        ...
        Py_BuildValue("d", user);
```

stswandering:
Cool! I have tried this method,it returns the same value. So, I change my pull request.

giampaolo:
You also need to change this:
```
Py_BuildValue("(fff)", user, system, idle);
```
...to this:
```
Py_BuildValue("(ddd)", user, system, idle);
```

stswandering:
Yeah,Sure! It's my fault.

giampaolo:
LGTM. Merging. Thanks. ;)

Issue #338, opened 2014-05-23
disk_io_counters failed to find some disks on GNU/linux


giampaolo:
_From [mani...@gmail.com](https://code.google.com/u/102145627999561308363/) on November 16, 2012 18:18:38_

```
What steps will reproduce the problem?  
1. Just have some lvm partition on a virtual server (xen in my case) with no 
digit in its name, like this /proc/partitions : 

major minor  #blocks  name

 202        0   19922944 xvda
 202       16    1048576 xvdb 

What is the expected output?  
Any correct result 

What do you see instead?  
RuntimeError: couldn't find any physical disk 

What version of psutil are you using? What Python version?  
psutil 0.6.1 on python 2.6.6 

On what operating system? Is it 32bit or 64bit version?  
GNU/linux debian squeeze, 64 bits, in a xen domU 

Please provide any additional information below.  
The file /proc/diskstats correctly show some informations : 

 202       0 xvda 27846519 29828 1002976690 121876532 13449001 43255546 
453670328 1152982588 0 79363348 1275022004
 202      16 xvdb 55557 1134 453464 412056 38511 33466 575840 1699516 0 272092 2111504

It's not a real physical disk, it's a lvm volume, but the virtual server show 
it as a real disk.

In _pslinux.py , line 322, I can see that to be added as a partition, the name 
of the partition must end with a digit, but that's not always the case. This is the bug.
```

_Original issue: http://code.google.com/p/psutil/issues/detail?id=338_

giampaolo:
Reopening as per https://github.com/giampaolo/psutil/pull/1244

PR #1244, opened 2018-03-13
Avoid counting disk io twice for NVMe SSD block device

spacewander:
The first NVMe block device will be named as nvme0n1,
and its partitions are nvme0n1p1, nvme0n1p2, etc.
Therefore we could not use the last letter to identify if the name is a
partition or not.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1017,23 +1017,24 @@ def disk_io_counters():
     system as a dict of raw tuples.
     """
     # determine partitions we want to look for
+    partition_suffix = re.compile(r'^p*\d+$')
+
     def get_partitions():
         partitions = []
         with open_text("%s/partitions" % get_procfs_path()) as f:
             lines = f.readlines()[2:]
         for line in reversed(lines):
             _, _, _, name = line.split()
-            if name[-1].isdigit():
-                # we're dealing with a partition (e.g. 'sda1'); 'sda' will
-                # also be around but we want to omit it
-                partitions.append(name)
-            else:
-                if not partitions or not partitions[-1].startswith(name):
-                    # we're dealing with a disk entity for which no
-                    # partitions have been defined (e.g. 'sda' but
-                    # 'sda1' was not around), see:
-                    # https://github.com/giampaolo/psutil/issues/338
+            if partitions and partitions[-1].startswith(name):
+                suffix = partitions[-1][len(name):]
+                if not partition_suffix.match(suffix):
+                    # If a disk entity (e.g. 'sda' or 'nvme0n1') has
+                    # partition(e.g. 'sda1' or 'nvme0n1p1'),
+                    # we will deal with the partition and ignore the
+                    # disk entiy.
                     partitions.append(name)
+            else:
+                partitions.append(name)
         return partitions
 
     retdict = {}
--- a/psutil/tests/test_linux.py
+++ b/psutil/tests/test_linux.py
@@ -1014,6 +1014,39 @@ class TestSystemDisks(unittest.TestCase):
             self.assertEqual(ret.write_time, 0)
             self.assertEqual(ret.busy_time, 0)
 
+    def test_disk_io_counters_with_nvme(self):
+        def open_mock(name, *args, **kwargs):
+            if name == '/proc/partitions':
+                return io.StringIO(textwrap.dedent(u"""\
+                    major minor  #blocks  name
+
+                    259        0  250059096 nvme0n1
+                    259        1     266240 nvme0n1p1
+                    259        2      16384 nvme0n1p2
+                    """))
+            elif name == '/proc/diskstats':
+                return io.StringIO(textwrap.dedent(u"""\
+                        259       0 nvme0n1  3 6 9 12 15 18 21 24 27 30 33
+                        259       1 nvme0n1p1 1 2 3 4 5 6 7 8 9 10 11
+                        259       2 nvme0n1p2 2 4 6 8 10 12 14 16 18 20 22
+                    """))
+            else:
+                return orig_open(name, *args, **kwargs)
+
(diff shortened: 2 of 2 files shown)


spacewander:
Here is the output from current `disk_io_counters`:
```
In [2]: psutil.disk_io_counters(perdisk=True)
Out[2]: 
{'nvme0n1': sdiskio(read_count=59871, write_count=26114, read_bytes=1706764800, write_bytes=993231872, read_time=21832, write_time=29972, read_merged_count=43, write_merged_count=38182, busy_time=19128),
 'nvme0n1p1': sdiskio(read_count=669, write_count=2, read_bytes=6141440, write_bytes=1024, read_time=20, write_time=0, read_merged_count=0, write_merged_count=0, busy_time=20),
 'nvme0n1p2': sdiskio(read_count=32, write_count=0, read_bytes=581632, write_bytes=0, read_time=4, write_time=0, read_merged_count=0, write_merged_count=0, busy_time=4),
 'nvme0n1p3': sdiskio(read_count=42, write_count=0, read_bytes=2121728, write_bytes=0, read_time=8, write_time=0, read_merged_count=0, write_merged_count=0, busy_time=8),
 'nvme0n1p4': sdiskio(read_count=42, write_count=0, read_bytes=2121728, write_bytes=0, read_time=20, write_time=0, read_merged_count=0, write_merged_count=0, busy_time=12),
 'nvme0n1p5': sdiskio(read_count=42, write_count=0, read_bytes=2121728, write_bytes=0, read_time=24, write_time=0, read_merged_count=0, write_merged_count=0, busy_time=16),
 'nvme0n1p6': sdiskio(read_count=58958, write_count=25940, read_bytes=1691456512, write_bytes=993230848, read_time=21748, write_time=29808, read_merged_count=43, write_merged_count=38182, busy_time=18948)}
```

spacewander:
Any suggestion about this pull request?

giampaolo:
Sorry, been busy. I briefly took a look at this and it seems more complex. Looking for a number is not enough, see:
https://github.com/giampaolo/psutil/issues/338

```
major minor  #blocks  name

 202        0   19922944 xvda
 202       16    1048576 xvdb 
``` 
In the case above `xvdb` should be taken into account. As such it appears a logic based on the partition 
 `name` alone is not enough. We probably want to play with `minor` field instead. The info that I've found online about /proc/partitions is pretty scarce but I suspect we may want to collect all names with a `minor` number > 0. I will have to dig more in order to confirm this though.

spacewander:
@giampaolo 
Thanks for your quick reply!

I dived into the Linux's source code tonight, and found the `/proc/partitions` is displayed by `show_partition`:
https://elixir.bootlin.com/linux/v3.10.108/source/block/genhd.c#L845
Each disk/partition is printed by `disk_name`:
https://elixir.bootlin.com/linux/v3.10.108/source/block/partition-generic.c#L34
I have also checked the source from a higher version, the code is almost unchanged.

There are three combinations of the name part in `/proc/partitions`:
1. disk_name, if partition_no is None
2. disk_name + 'p' + partition_no, if disk_name ends with a digit
3. disk_name + partition_no

I have updated my code as a result of this new discovery.

psutil/_pslinux.py
```diff
@@ -1017,23 +1017,24 @@ def disk_io_counters():
...
-                    # https://github.com/giampaolo/psutil/issues/338
+            if partitions and partitions[-1].startswith(name):
+                suffix = partitions[-1][len(name):]
+                if not partition_suffix.match(suffix):
+                    # If a disk entity (e.g. 'sda' or 'nvme0n1') has
+                    # partition(e.g. 'sda1' or 'nvme0n1p1'),
+                    # we will deal with the partition and ignore the
+                    # disk entiy.
                     partitions.append(name)
+            else:
```
kportertx: I believe this incorrectly assumes the last disk it a partition.

psutil/_pslinux.py
```diff
@@ -1017,23 +1017,24 @@ def disk_io_counters():
     system as a dict of raw tuples.
     """
     # determine partitions we want to look for
+    partition_suffix = re.compile(r'p*\d+')
```
kportertx: Wouldn't this still match nvme0n1? It is looking for 0 or more 'p's followed by one or more digits.

giampaolo:
Note: https://github.com/giampaolo/psutil/issues/1305#issuecomment-406567114 states that we can have read/write loads also for disks with no partition, which complicates things even more.

psutil/_pslinux.py
spacewander: @kportertx 
`get_partitions` returns both partitions and disk without any partitions. If the last disk is not a partition, it will be a disk without partition.

psutil/_pslinux.py
spacewander: @kportertx 
The `partition_suffix` is used to match partition suffix, not the whole partition name. Certainly, it will be better to write `r'^p*\d+$'` to avoid confusing. I will change it in the new commit.

giampaolo:
Update: it turns out `iostat` skips some disks/partitions (and psutil doesn't). That's what we should do. I'm trying to skim through iostat source code in order to understand where the "skip" of certain occurs:
https://github.com/sysstat/sysstat/blob/master/iostat.c
Any help is welcome. ;)

giampaolo:
I filed a bug on `iostat` project asking for help:
https://github.com/sysstat/sysstat/issues/185

giampaolo:
@wiggin15 thanks a lot man, that's exactly what I was looking for

wiggin15:
@giampaolo  FYI I started working on a new (bigger) patch here:
https://github.com/Infinidat/psutil/commit/512395df8f46b8b88fae8737a2e66b4b3018d511
just so we won't do the work twice

Issue #1194, opened 2017-12-13
psutil fails to return the nice value for some processes on Solaris 10


gsauthof:
How to reproduce - chose xntp on Solaris 10:

    >>> p = psutil.Process(960)
    >>> p.name()
    'xntpd'
    >>> p.nice()
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/path/to/venv/lib/python3.5/site-packages/psutil/__init__.py", line 755, in nice
	return self._proc.nice_get()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 346, in wrapper
	return fun(self, *args, **kwargs)
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 451, in nice_get
	return cext_posix.getpriority(self.pid)
    OSError: [Errno 22] Invalid argument

The system's ps is able to return the nice value:

    $ ps -e -o pid,uid,gid,ruid,nice | grep 960
       960     0     0     0 RT

A simple C test program that just calls `getpriority()` on that PID also fails with EINVAL.

Perhaps `getpriority()` always fails for realtime processes, on Solaris 10. Is kind of plausible - since priorities doesn't really make sense for realtime. The man page doesn't include details on that topic, though.

This breaks `psutil.process_iter()` unless one uses a non-default attribute set that doesn't include `'nice'`. Example how to make it fail with a minimal attribute set:

    >>> for p in psutil.process_iter(): print(p.as_dict(attrs=['nice']))
    {'nice': 0}
    {'nice': -20}
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/path/to/venv/lib/python3.5/site-packages/psutil/__init__.py", line 574, in as_dict
	ret = meth()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/__init__.py", line 755, in nice
	return self._proc.nice_get()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 346, in wrapper
	return fun(self, *args, **kwargs)
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 451, in nice_get
	return cext_posix.getpriority(self.pid)
    OSError: [Errno 22] Invalid argument

The psutil Solaris-specific code [contains a note][1] regarding `getpriority()` vs. `psinfo` that mentions issue #1082.

There are several ways how to fix this:

[...]

Issue #1120, opened 2017-09-02
PEP-0527 - no longer provide .exe files 


giampaolo:
For Windows we currently upload and provide `.exe` and `.wheel` files on each release. PEP-527 deprecates `.exe` files (aka: `python setup.py bdist_wininst`) and AFAIU the new PYPI infrastructure will even prevent uploading / hosting them:
https://www.python.org/dev/peps/pep-0527/
This is good news as we'll have to upload 10 files (+1, the tar.gz) per release instead of 20.

giampaolo:
Fixed in 5.3.1.

PR #1248, opened 2018-03-15
Solaris 10 Fixes

### LINKED ISSUE #1193: psutil only returns uids of user-owned process on Solaris 10
How to reproduce:

As normal user, chose a PID of another user:

    >>> p = psutil.Process(12345)
    >>> p.uids()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 346, in wrapper
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 477, in uids
        real, effective, saved, _, _, _ = self._proc_cred()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_common.py", line 336, in wrapper
        return fun(self)
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 400, in _proc_cred
        return cext.proc_cred(self.pid, self._procfs_path)
    PermissionError: [Errno 13] Permission denied: '/proc/12345/cred'
     
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_common.py", line 336, in wrapper
        return fun(self)
      File "/path/to/venv/lib/python3.5/site-packages/psutil/__init__.py", line 768, in uids
        return self._proc.uids()
      File "/path/to/venv/lib/python3.5/site-packages/psutil/_pssunos.py", line 362, in wrapper
        raise AccessDenied(self.pid, self._name)
    psutil.AccessDenied: psutil.AccessDenied (pid=12345)

The thing is that psutil tries to read the uids from `/proc/$pid/cred` that is always only user-readable on Solaris 10:

    $ ls /proc/12345/cred -l
    -r-------- 1 juser juser 52 2017-11-23 11:27 /proc/12345/cred

A better choice on Solaris 10 is to use `/proc/$pid/psinfo` since it [also contains the uids](https://docs.oracle.com/cd/E19253-01/816-5174/proc-4/index.html) (modulo the saved user id) - and is readable by everyone. Thus, if psutil really cares about the saved user id, it could also try `/proc/$pid/cred` first and then fall back to `/proc/$pid/psinfo`.

For our example:

    $ ls /proc/12345/psinfo -l
    -r--r--r-- 1 juser juser 336 2017-11-23 11:27 /proc/12345/psinfo

gsauthof:
Fixes #1193 and #1194, a double free, some undefined behavior and some warnings.

### DIFF
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -432,27 +432,10 @@ class Process(object):
 
     @wrap_exceptions
     def nice_get(self):
-        # Note #1: for some reason getpriority(3) return ESRCH (no such
-        # process) for certain low-pid processes, no matter what (even
-        # as root).
-        # The process actually exists though, as it has a name,
-        # creation time, etc.
-        # The best thing we can do here appears to be raising AD.
-        # Note: tested on Solaris 11; on Open Solaris 5 everything is
-        # fine.
-        #
-        # Note #2: we also can get niceness from /proc/pid/psinfo
-        # but it's wrong, see:
-        # https://github.com/giampaolo/psutil/issues/1082
-        try:
-            return cext_posix.getpriority(self.pid)
-        except EnvironmentError as err:
-            # 48 is 'operation not supported' but errno does not expose
-            # it. It occurs for low system pids.
-            if err.errno in (errno.ENOENT, errno.ESRCH, 48):
-                if pid_exists(self.pid):
-                    raise AccessDenied(self.pid, self._name)
-            raise
+        # Note #1: getpriority(3) doesn't work for realtime processes.
+        # Psinfo is what ps uses, see:
+        # https://github.com/giampaolo/psutil/issues/1194
+        return self._proc_basic_info()[proc_info_map['nice']]
 
     @wrap_exceptions
     def nice_set(self, value):
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -79,7 +79,11 @@ proc_info_map = dict(
     nice=4,
     num_threads=5,
     status=6,
-    ttynr=7)
+    ttynr=7,
+    uid=8,
+    euid=9,
+    gid=10,
+    egid=11)
 
 
 # =====================================================================
@@ -394,7 +398,10 @@ class Process(object):
 
     @memoize_when_activated
     def _proc_cred(self):
-        return cext.proc_cred(self.pid, self._procfs_path)
+        @wrap_exceptions
+        def proc_cred(self):
+            return cext.proc_cred(self.pid, self._procfs_path)
+        return proc_cred(self)
 
     @wrap_exceptions
     def name(self):
@@ -454,12 +461,22 @@ class Process(object):
 
     @wrap_exceptions
     def uids(self):
-        real, effective, saved, _, _, _ = self._proc_cred()
+        try:
+            real, effective, saved, _, _, _ = self._proc_cred()
+        except AccessDenied:
+            real = self._proc_basic_info()[proc_info_map['uid']]
+            effective = self._proc_basic_info()[proc_info_map['euid']]
+            saved = None
         return _common.puids(real, effective, saved)
 
     @wrap_exceptions
     def gids(self):
-        _, _, _, real, effective, saved = self._proc_cred()
+        try:
+            _, _, _, real, effective, saved = self._proc_cred()
+        except AccessDenied:
+            real = self._proc_basic_info()[proc_info_map['gid']]
+            effective = self._proc_basic_info()[proc_info_map['egid']]
+            saved = None
         return _common.puids(real, effective, saved)
 
     @wrap_exceptions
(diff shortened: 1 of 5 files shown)


psutil/_pssunos.py
```diff
@@ -432,27 +439,10 @@ def num_threads(self):
...
-            # 48 is 'operation not supported' but errno does not expose
-            # it. It occurs for low system pids.
-            if err.errno in (errno.ENOENT, errno.ESRCH, 48):
-                if pid_exists(self.pid):
-                    raise AccessDenied(self.pid, self._name)
-            raise
+        # Note #1: getpriority(3) doesn't work for realtime processes.
+        # Psinfo is what ps uses, see:
+        # https://github.com/giampaolo/psutil/issues/1194
+        return self._proc_basic_info()[proc_info_map['nice']]
```
giampaolo: Nice that we got rid of that mess.

psutil/_psutil_common.c
```diff
@@ -40,7 +40,7 @@ PyUnicode_DecodeFSDefaultAndSize(char *s, Py_ssize_t size) {
  * If msg != "" the exception message will change in accordance.
  */
 PyObject *
-NoSuchProcess(char *msg) {
+NoSuchProcess(const char *msg) {
```
giampaolo: why is this necessary?

psutil/_psutil_common.h
```diff
@@ -4,6 +4,9 @@
  * found in the LICENSE file.
  */
 
+#ifndef PSUTIL_PSUTIL_COMMON_H
+#define PSUTIL_PSUTIL_COMMON_H
```
giampaolo: why is this necessary?

psutil/_pssunos.py
```diff
@@ -394,7 +398,10 @@ def _proc_basic_info(self):
 
     @memoize_when_activated
     def _proc_cred(self):
-        return cext.proc_cred(self.pid, self._procfs_path)
+        @wrap_exceptions
```
giampaolo: Why not decorate the method instead?

psutil/_pssunos.py
```diff
@@ -471,12 +461,22 @@ def ppid(self):
...
 
     @wrap_exceptions
     def gids(self):
-        _, _, _, real, effective, saved = self._proc_cred()
+        try:
+            _, _, _, real, effective, saved = self._proc_cred()
+        except AccessDenied:
+            real      = self._proc_basic_info()[proc_info_map['gid']]
+            effective = self._proc_basic_info()[proc_info_map['egid']]
+            saved     = None
```
giampaolo: PEP8 (please remove extra spaces)

psutil/_pssunos.py
```diff
@@ -471,12 +461,22 @@ def ppid(self):
 
     @wrap_exceptions
     def uids(self):
-        real, effective, saved, _, _, _ = self._proc_cred()
+        try:
+            real, effective, saved, _, _, _ = self._proc_cred()
+        except AccessDenied:
+            real      = self._proc_basic_info()[proc_info_map['uid']]
+            effective = self._proc_basic_info()[proc_info_map['euid']]
+            saved     = None
```
giampaolo: PEP8 (please remove extra spaces)

giampaolo (changes requested):
Nice work. Could you please update HISTORY file describing the fixes/improvements which occurred in this PR?

psutil/_psutil_common.c
gsauthof: `const char *` means pointer to const characters. The const modifier signals the caller that NoSuchProcess will not change the passed message. Otherwise, it's not clear and one has to inspect the function body. In case someone accidentally writes to msg using that const pointer it's a compile error. This change also remove compiler warnings since NoSuchProcess is called with string literals that effectively have the type `const char *`. Thus, this additional type safety protects against accidental writing into a const character array through a non-const pointer which would invoke undefined behavior (e.g. crash if the string literals are placed in a read-only data section).

psutil/_psutil_common.h
gsauthof: This guards the header from being included multiple times into one translation unit (either directly/indirectly). In C, unless the header does something special, you always define such guards (or use the non-standardized once pragma).

psutil/_pssunos.py
gsauthof: I tried stacking the `@wrap_exceptions` on top of `@memoize_when_activated` of `_proc_cred` but it didn't work. Something about a missing attribute, IIRC.

psutil/_pssunos.py
gsauthof: Ok, will update (rebase) the commit.

psutil/_pssunos.py
gsauthof: Ok, will fix this.

gsauthof:
So I've updated the pull request and implemented the requested changes (HISTORY and PEP 8).

Btw, looking at the `HISTORY.rst` file I noticed:

    - 1120_: [SunOS] cmdline() could be truncated at the 15th character when
      reading it from /proc. An extra effort is made by reading it from process
      address space first.

But #1120 doesn't really match the description. Probably a mixup with #694, right?

giampaolo:
> But #1120 doesn't really match the description. Probably a mixup with #694, right?

correct; thanks for the PR.

PR #1250, opened 2018-03-17
Add the cpu number on tuple of threads 

alexfurtunato:
Sometimes its necessary to get the number of cpu that threads are running. So, i propose this merge to add this enhancement on Process Class


giampaolo:
Uhm... I'm skeptical about how useful this is. For the record, we added something similar, on a per-process basis, for Linux, FreeBSD, SunOS: `Process.cpu_num()`.

giampaolo:
Closing out as rejected.

Issue #1224, opened 2018-02-20
wait() function could throw TimeoutExpired exception when it shouldn't


xiaolingbao:
This issue is about function wait() https://github.com/giampaolo/psutil/blob/dd19ba7b5bbf5ea67a4d386baa3fabd13b118533/psutil/_pswindows.py#L791.

My test scenario is listed below, although it doesn't repro 100%, it happens quite often. I didn't measure how frequently it happens though.

1) Have a process object, proc, call proc.kill().
2) Call proc.wait(30)
3) proc.wait() throws TimeoutExpired almost instantly.

Speculation:
1) cext.proc_wait() returns quickly because process exited (killed) and returns a value not equal to WAIT_TIMEOUT.
2) Somehow, the process still exists, and thus pid_exists() returns True. (Zombie status? Other processes, threads still have references to the process handle?)
3) Since timeout value is not None, TimeoutExpired throws.

This behavior is different from the previous version. In the previous version, wait() will not throw TimeoutExpired, but just return the value returned from proc_wait().

giampaolo:
Mmmm, can you paste the traceback?

xiaolingbao:
This is our code snippet that calls psutil:
--------------------------------
          #  timeout = 30
 377   logging.info('Terminating process id: %r', process_id)
 378   try:
 379     proc = psutil.Process(process_id)
 380     proc.terminate()
 381     proc.wait(timeout)
 382     logging.info('Process %d was terminated successfully', process_id)
 383     return True
 384   except psutil.TimeoutExpired as e:
 385     logging.exception(e)
 386     logging.error('Failed to terminate process %d in %d seconds',
 387                   process_id, timeout)
 388     return False
 389   except psutil.AccessDenied:
 390     logging.error(
 391         'Unable to terminate process %d due to insufficient privileges',
 392         process_id)
 393     return False
 394   except psutil.NoSuchProcess:
 395     logging.info('Process %d was not running', process_id)
 396     return True
--------------------------------

This is the log got:
--------------------------------
[INFO 14:31:30.256 5056:516 test_lib.py:377]Terminating process id: 5472
[ERROR 14:31:30.367 5056:516 test_lib.py:385]psutil.TimeoutExpired timeout after 30 seconds (pid=5472)
Traceback (most recent call last):
  File "c:\TestAutomation\lib\test_lib.py", line 381, in _TerminateProcess
    proc.wait(timeout)
[...]

PR #1253, opened 2018-03-24
1224 fix windows Process.wait()

giampaolo:
See #1224


xiaolingbao:
I'm using python 2.7.14. Thank you.

wiggin15:
Just my one cent - `WAIT_ABANDONED` cannot return for `hProcess`. This value can return only when waiting for mutexes.

PR #1260, opened 2018-03-29
Fixes swap_memory not returning bytes on FreeBSD

### LINKED ISSUE #1255: swap_memory is not in bytes on FreeBSD
Running `psutil.swap_memory` on FreeBSD 10.3 I gives me the following output:

    sswap(total=2097088, used=17347, free=2079741, percent=0.8, sin=53145, sout=76176)

Running `swapinfo` at the same time returns with the following output:

```bash
Device          1K-blocks     Used    Avail Capacity
/dev/gpt/swap0    4194304    34308  4159996     1%
/dev/gpt/swap1    4194304    35080  4159224     1%
Total             8388608    69388  8319220     1%
```

In other words, psutil thinks I got `2'097'088` bytes of swap, even though I actually have `8'388'608'000` bytes. Since the answer is off by a factor of almost exactly 4000, I assume that psutil doesn't really return bytes here.

Am I missing something? These values should be in bytes, right?

href:
This fixes #1255

### DIFF
--- a/psutil/arch/freebsd/specific.c
+++ b/psutil/arch/freebsd/specific.c
@@ -516,10 +516,12 @@ psutil_swap_mem(PyObject *self, PyObject *args) {
     if (sysctlbyname("vm.stats.vm.v_vnodeout", &nodeout, &size, NULL, 0) == -1)
         goto error;
 
-    return Py_BuildValue("(iiiII)",
-                         kvmsw[0].ksw_total,                     // total
-                         kvmsw[0].ksw_used,                      // used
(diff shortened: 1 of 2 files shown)


giampaolo:
Thanks. Did you verify it matches swapinfo? It would be nice to add a unit test for it checking swapinfo output in test_bsd.py.

href:
I did, and there is a test in test_bsd.py ;)

giampaolo:
Great stuff! Merging.

href:
Cheers! As an aside: I love how straight forward it was to get the tests for this project running 👍

PR #1266, opened 2018-04-12
Improve sensors_battery for linux platforms

yehan2002:
add fallback for power_now
add new paths for power cables

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1243,7 +1243,9 @@ def sensors_battery():
     energy_full = multi_cat(
         root + "/energy_full",
         root + "/charge_full")
-    if energy_now is None or power_now is None:
+    if power_now is None: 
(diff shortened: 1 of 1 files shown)


psutil/_pslinux.py
```diff
@@ -1243,7 +1243,9 @@ def multi_cat(*paths):
     energy_full = multi_cat(
         root + "/energy_full",
         root + "/charge_full")
-    if energy_now is None or power_now is None:
+    if power_now is None: 
+        power_now = 0 # provide a fallback since some systems don't provide this.
```
giampaolo: This doesn't look right. If `power_now` can't be determined we also cannot determine `secsleft` so it's probably better to just return None

yehan2002: but in case `power_now` is not available this still gives the battery percentage and whether or not the charger is connected

Issue #1258, opened 2018-03-28
Python interpreter crash on Windows 10 while printing processes username with another user logged in.


KelSolaar:
Hi,

We are experiencing hard Python interpreter crash on Windows 10 and psutil 5.4.3 when printing processes username with another user logged. Basically I'm logged into the workstation and another user is also logged in on his account.

Here is the reproducible code:

```python
import psutil


def list_processes_username():
    for pid in psutil.pids():
        if not psutil.pid_exists(pid):
            continue

        try:
            print(psutil.Process(pid).username())
        except psutil.AccessDenied:
            pass


print(list_processes_username())
print(list_processes_username())
```

Crash Dialog:

![image](https://user-images.githubusercontent.com/99779/38060761-a164e716-3348-11e8-9002-1d2df921c0e0.png)

Current Windows version:

![image](https://user-images.githubusercontent.com/99779/38060754-909b0690-3348-11e8-9752-c52885bb7908.png)

KelSolaar:
By printing the PIDs and checking in the Task Manager I was able to confirm that it is when trying to access a process from the other user.

giampaolo:
Shit... 
What Python and Windows version is this?

KelSolaar:
Python 2.7, Windows version is in the last screenshot of OP.

giampaolo:
Python 2.7 32 or 64 bit?

KelSolaar:
64bit! Sorry for not being specific.

giampaolo:
No worries. As expected, I cannot reproduce it. I tried with this with no luck:

```python
import psutil

while 1:
    for p in psutil.process_iter():
        try:
            print(p.username())
        except psutil.AccessDenied:
            pass
```
Can you identify the PID causing the crash, run this script and paste the output?
```python
from pprint import pprint as pp
import psutil
attrs = set(['num_ctx_switches', 'pid', 'memory_full_info', 'connections', 'cmdline', 'create_time', 'ionice', 'memory_maps', 'cpu_percent', 'ppid', 'cwd', 'nice', 'status', 'cpu_times', 'io_counters', 'memory_info', 'threads', 'open_files', 'num_threads', 'exe', 'name', 'cpu_affinity', 'memory_percent', 'environ'])
PID = 1283
pp(psutil.Process(PID).as_dict(attrs=attrs))
```

KelSolaar:
Yeah, I'll try that at work, we might be able to provide a stack trace. I'll keep you posted.

[...]

PR #1289, opened 2018-06-07
Fix Windows crash on proc_username()

jmigot-tehtris:
There is a crash possibility in proc_username() function on Windows platform, if  any of the Windows API calls fail, because the to "goto error" which executes this:
    Py_XDECREF(py_domain);
    Py_XDECREF(py_username);
    Py_XDECREF(py_tuple);

This will lead to deallocation of uninitialized random memory, which will segfault and crash the entire Python interperter.


giampaolo:
Mmm... possibly this could be related to https://github.com/giampaolo/psutil/issues/1258.
Patch LGTM. Merging.

giampaolo:
New release is out.

PR #1292, opened 2018-06-10
Add seconds to thread run times on MacOS.

### LINKED ISSUE #1258: Python interpreter crash on Windows 10 while printing processes username with another user logged in.

### LINKED ISSUE #1278: MacOS thread times are incorrect since they only consider microseconds
On MacOS, psutil uses only the microseconds field here https://github.com/giampaolo/psutil/blob/master/psutil/_psutil_osx.c#L1072. 

The struct also has a seconds field, which should be added to the return value.

nikhilm:
Fixes #1278

### DIFF
--- a/psutil/_psutil_osx.c
+++ b/psutil/_psutil_osx.c
@@ -1069,8 +1069,8 @@ psutil_proc_threads(PyObject *self, PyObject *args) {
         py_tuple = Py_BuildValue(
             "Iff",
             j + 1,
-            (float)basic_info_th->user_time.microseconds / 1000000.0,
-            (float)basic_info_th->system_time.microseconds / 1000000.0
+            basic_info_th->user_time.seconds + (float)basic_info_th->user_time.microseconds / 1000000.0,
+            basic_info_th->system_time.seconds + (float)basic_info_th->system_time.microseconds / 1000000.0
         );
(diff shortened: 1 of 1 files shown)


giampaolo:
Hello, sorry for the delay. This looks good. #1258 appears to be unrelated though. What was the issue you were trying to address? Also, could you please respect the 80 chars per line style? Like this:

```
            basic_info_th->user_time.seconds + \
                (float)basic_info_th->user_time.microseconds / 1000000.0,
            basic_info_th->system_time.seconds + \
                (float)basic_info_th->system_time.microseconds / 1000000.0
```

nikhilm:
Sorry I meant #1278

PR #1300, opened 2018-07-04
Improved macOS temperature detection

JMY1000:
**Basically just a dummy PR at this point**

### Current blocking issue
[This line](https://github.com/JMY1000/psutil/blob/4305c07cf4299fd0f6728cbf38a2370803541b55/psutil/_psutil_osx.c#L1956) throws `EXC_BAD_ACCESS`.

### Goals
- Better performance
- More extensible
- Easier to configure
- Improved detection
- More sensors

### Checklist 
- [ ] Minimize I/O calls
    - [ ] Call `SMCClose()` properly
    - [ ] Maintain sensor dict
- [ ] Detection
    - [ ] On-the-fly
    - [ ] Polled values
- [ ] Grouped sensors
    - [ ] Underscore replacement
- [ ] More sensors
    - [ ] Temperature sensors
    - [ ] Fan sensors
    - [ ] Voltage sensors
    - [ ] Other sensors
- [ ] Style
    - [ ] Enforce line length
    - [ ] Enforce indentation length


JMY1000:
@giampaolo This is the branch I've been working in, changes from master merged and all. I've been testing using the following test file in CLion (using CMake as the build system.) 

[CMakeLists.txt](https://github.com/giampaolo/psutil/files/2172175/CMakeLists.txt)
[test.c](https://github.com/giampaolo/psutil/files/2172177/test.txt)

[_psutil_osx.c:2251](https://github.com/JMY1000/psutil/blob/da78f27701ecc74f3953d03c04862c8c931b06e9/psutil/_psutil_osx.c#L2251) has the error "Can't resolve variable PSUTIL_VERSION" when building with CMake. Commenting the line out makes it happy, but makes the builtin tests unhappy.

Besides that, the status is the same as before: After several calls to `SMCGetTemperature()`, `SMCOpen()` (called inside of `SMCGetTemperature()`) errors out with the following error:

    test_sensors(19918,0x7fff9d644340) malloc: *** error for object 0x7fbd53c00348: incorrect checksum for freed object - object was probably modified after being freed.
    *** set a breakpoint in malloc_error_break to debug

It doesn't seem to consistently fail on a single value or at a singular time. I don't know what's causing this, but it doesn't immediately appear to be something in what I wrote. The memory address shifts around, but not every run. Any ideas what might be happening?

giampaolo:
> _psutil_osx.c:2251 has the error "Can't resolve variable PSUTIL_VERSION" when building with CMake. Commenting the line out makes it happy, but makes the builtin tests unhappy.

Don't know about CMake. I use standard `make` to compile psutil on all UNIX platforms. You may want to take a look at https://github.com/giampaolo/psutil/blob/master/DEVGUIDE.rst first.

As for your error, I'm not sure why it happens or what it means, and I can't test your code because unfortunately my virtualized OSX box does not provide any sensor. Does it occur also with `master` branch? The only thing I can do is support you via code reviews. Again, my original suggestion was to:

1) rewrite the current key/value mapping in C, refactor C code so that `SMCOpen()` is called only once, submit a PR
2) add additional temperature sensors, submit a PR
3) add voltage sensors, submit a PR

This is, in my opinion, the correct way to proceed if you want to actually contribute and see your PRs merged.

JMY1000:
@giampaolo 

> Don't know about CMake. I use standard make to compile psutil on all UNIX platforms. You may want to take a look at https://github.com/giampaolo/psutil/blob/master/DEVGUIDE.rst first.

I've looked at that guide. I've been using CMake because that's what CLion supports, and I enjoy CLion's development and (more importantly) *debugging* environment. If you have an alternative method, please feel free to share it.

> Does it occur also with master branch?
Not in a plain loop, but with my code, yes. I'm not sure exactly what's going on, but it seems to be an  issue with [this line.](https://github.com/JMY1000/psutil/blob/da78f27701ecc74f3953d03c04862c8c931b06e9/psutil/_psutil_osx.c#L151) Still working on a solution.

> This is, in my opinion, the correct way to proceed if you want to actually contribute and see your PRs merged.
That's already what I'm doing.

giampaolo:
On Tue, 10 Jul 2018 at 03:11, JMY1000 <notifications@github.com> wrote:

> @giampaolo <https://github.com/giampaolo>
>
> Don't know about CMake. I use standard make to compile psutil on all UNIX
> platforms. You may want to take a look at
> https://github.com/giampaolo/psutil/blob/master/DEVGUIDE.rst first.
>
> I've looked at that guide. I've been using CMake because that's what CLion
> supports, and I enjoy CLion's development and (more importantly)
> *debugging* environment. If you have an alternative method, please feel
> free to share it.
>

The alternative method is make. That is what psutil project uses since day
one, sorry.


Does it occur also with master branch?
> Not in a plain loop, but with my code, yes.
>
>
What do you mean? Does it occur or not? My question is to understand
whether your branch  introduced the bug or whether the bug was already
present in the master branch. Have you tested master branch or only your
branch?

>
This is, in my opinion, the correct way to proceed if you want to actually
> contribute and see your PRs merged.
> That's already what I'm doing.
>
>
This PR is clearly not doing that.

> --
Giampaolo - http://grodola.blogspot.com

JMY1000:
> The alternative method is make. That is what psutil project uses since day
one, sorry.

What debugger are you using for the C code?

> What do you mean? Does it occur or not? My question is to understand
whether your branch  introduced the bug or whether the bug was already
present in the master branch. Have you tested master branch or only your branch.

My branch introduced the bug, and I've patched it with 94b5d4276866fced31bd346c8300e965f5505fe7 (although I'll need to introduce a less clumsy fix in the future.) I was incorrect in my previous assumption that just because the error was occurring in code that I wrote it wasn't caused by some of my code.

> This PR is clearly not doing that.

I'm adding a single feature at a time with the plan being to submit PRs individually for each. Like I said, it's basically a dummy PR at this point; the checklist is more general than I plan to actually have be finished. I've split things out into different branches in my fork. The current code I've been focusing on deals mostly refactoring the existing code into C–not immediately to improve performance or add more sensors, but to get ready for that. If there's a problem with how I'm doing that, I'm happy to change.

JMY1000:
@giampaolo Separate thing: Do we want to assume that temperatures aren't being disabled and enabled so we can keep more things constant (e.g. only generate the dictionary with all the sensors once and simply update the values after that?)

giampaolo:
> What debugger are you using for the C code?

I don't use any debugger, nor I know how to use CMake (or anything else) to do it. If you think this is valuable for others feel free to make a PR for it. Let's just not mix apples and oranges in here please.

> I'm adding a single feature at a time with the plan being to submit PRs individually for each. [...] The current code I've been focusing on deals mostly refactoring the existing code into C–not immediately to improve performance or add more sensors, but to get ready for that. 

Fair enough. 

>  Separate thing: Do we want to assume that temperatures aren't being disabled and enabled so we can keep more things constant (e.g. only generate the dictionary with all the sensors once and simply update the values after that?)

I haven't looked at your PR in detail but it appears you're doing some kind of detection first. Don't do that. You should just have a list of constant strings in C, iterate over all of them, call `SMCReadKey()` and "skip" when that returns a non-meaningful value (`< 0` or whatever). That's it. This is what current master branch does and it does it fine: on a real macOS this will return some results, on my virtualized OSX box it just returns an empty dict without crashing.

JMY1000:
> If there is a way to determine the number of CPUs/GPUs/etc. at runtime then fine

There is. On top of that, since they're ordered sequentially, just iterating up from 0 and then stopping once it fails works too.

giampaolo:
This never got past the dummy stage and the files it touches (smc.c, _psutil_posix.c) are long gone. Closing. macOS temperature support is still wanted, but it'd have to start from scratch at this point.

JMY1000:
Fair enough, I honestly completely forgot about this haha.

Issue #1305, opened 2018-07-19
Misreporting disk io (values inflated)


kportertx:
Note: This could be an issue with glances.

Using glances to observe a system (ubuntu18.04 under test, I noticed that the disk throughput on 8 NVME devices were 8x larger than expected when compared with values from iostat. We then noticed that another instance running ubuntu16.04 was reporting the correct values. Glances versions differed on the two instances (2.11.1 vs 2.3) as well as psutil version (5.4.6 vs 3.4.2). Downgrading glances didn't fix the problem, but downgrading psutil did.

FYI, this only affected NVME disks, the regular drives reported the expected values. Also these are GCE instances.

I'm not sure what API glances is using, the following seems like a reasonable guess:

With psutil 3.4.2:

```
psutil.disk_io_counters(perdisk=True)
Out[5]: 
{'loop0': sdiskio(read_count=4971, write_count=0, read_bytes=6156288, write_bytes=0, read_time=168, write_time=0),
 'loop1': sdiskio(read_count=60, write_count=0, read_bytes=392192, write_bytes=0, read_time=0, write_time=0),
 'nvme0n1': sdiskio(read_count=12330268, write_count=71920, read_bytes=125630955520, write_bytes=75413585920, read_time=1328276, write_time=31400),
 'nvme0n2': sdiskio(read_count=12337189, write_count=71920, read_bytes=125658554368, write_bytes=75413585920, read_time=1338748, write_time=32188),
 'nvme0n3': sdiskio(read_count=12333970, write_count=71921, read_bytes=125648957440, write_bytes=75414634496, read_time=1340996, write_time=32440),
 'nvme0n4': sdiskio(read_count=12337859, write_count=71921, read_bytes=125662195712, write_bytes=75414634496, read_time=1349044, write_time=32272),
 'nvme0n5': sdiskio(read_count=12335489, write_count=71921, read_bytes=125655179264, write_bytes=75414634496, read_time=1355292, write_time=31796),
 'nvme0n6': sdiskio(read_count=12332568, write_count=71921, read_bytes=125640523776, write_bytes=75414634496, read_time=1354424, write_time=29684),
 'sda1': sdiskio(read_count=9915, write_count=38226, read_bytes=392524288, write_bytes=1677148160, read_time=11464, write_time=101900),
 'sda14': sdiskio(read_count=184, write_count=0, read_bytes=746496, write_bytes=0, read_time=84, write_time=0),
[...]

PR #1312, opened 2018-07-28
Differentiate between partitions and devices using sysfs

wiggin15:
disk_io_counters skips devices that have partitions, and to differentiate partitions from devices we can't use the names - we need to look in sysfs.
Partitions and devices have entries in /sys/class/block and only devices have entries in /sys/block.
This should fix issues #1244 and #1305.
Also fixed looking up the sector size of partitions.


wiggin15:
Closed in favor of #1313.

PR #1320, opened 2018-08-10
[aix] improve compilation on AIX: better support for gcc/g++, fix cpu metrics

truthbk:
I needed to build this with GCC/g++ via the AIX linux toolkit, the guards added here allowed me to do so.

I also found what appeared to be inconsistent CPU utilization values, they were being reported with ticks unit, as opposed to seconds. This PR fixes that as well.

### DIFF
--- a/psutil/_psutil_posix.c
+++ b/psutil/_psutil_posix.c
@@ -655,6 +655,10 @@ struct module_state {
 #define GETSTATE(m) (&_state)
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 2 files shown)


psutil/_psutil_posix.c
```diff
@@ -655,6 +655,10 @@ struct module_state {
 #define GETSTATE(m) (&_state)
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
giampaolo: why is this necessary?

PR #1322, opened 2018-08-11
gracefully handle unicode surrogate bogosity in pmap.py

Kelledin:
Lately, while running 5.4.6 unit-tests on Python 3.6, I discovered the test_pmap case failing due to scripts/pmap.py barfing on some fugly Unicode sequences.  Apparently, ext4 on Linux can have some oddball Unicode embedded in filenames that Python 3.6.3 rejects with complaints about surrogate characters.  Specifically, the offending paths appear to be transient testfn tmpfiles that happen to be mapped within the interpreter when test_pmap.py acts on it, with pathnames like 'build/$testfnf\udcc0\udc80dxyit7fq.so'.

I'm not sure which other testcase creates these files, but if I run the affected test_pmap case by itself, it will happily succeed.  It only seems to fail when I run the test suite in its entirety.  For background, the testbed is an arm64 system, running a custom Linux distro with a mostly-vanilla 4.16.15 kernel.

For now the most robust way I can see to handle this is catch the exception and fall back on running the offending pathname through repr().  It appears to produce sane pmap.py output and pass tests on Python 2.7.15 and 3.6.3.

### DIFF
--- a/scripts/pmap.py
+++ b/scripts/pmap.py
@@ -45,11 +45,24 @@ def main():
     total_rss = 0
     for m in p.memory_maps(grouped=False):
         total_rss += m.rss
-        print(templ % (
-            m.addr.split('-')[0].zfill(16),
-            str(m.rss / 1024) + 'K',
-            m.perms,
-            m.path))
+        try:
+            print(templ % (
+                m.addr.split('-')[0].zfill(16),
+                str(m.rss / 1024) + 'K',
+                m.perms,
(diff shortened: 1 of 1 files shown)


giampaolo:
That is probably due to default shell encoding which is != UTF8 (I guess you can figure it out with `echo $LANG` or something). Anyway, I don't think it's worth the trouble. pmap.py is just an example script after all...

Issue #1321, opened 2018-08-11
Try use /sys/block/*/stat for disk_io_counters if /process/diskstats doesn't exist


LEAFERx:
For some VPS such as OPENVZ based, there is no /proc/diskstats, then psutil.disk_io_counters will fail.
But io Info's can still be collected by looking through /sys/block/*/stat. iostat has done that. Would you nicely add this feature for us who don't have /proc/diskstats? Really appreciate it :)

giampaolo:
Why don't you make a PR instead? This is open source, you know... ;-)

LEAFERx:
Yeah I know but I'm lack of coding experience especially those with Linux system stuff. Sure I will try, but I'm still hoping you could make elegant solution for us.

PR #1324, opened 2018-08-13
Try to use /sys/block/*/stat for disk_io_counters if /process/diskstats doesn't exist

LEAFERx:
fix of #1321


giampaolo:
Thanks for writing this. I'm gonna merge it now and modify it a bit later as I prefer to split logic in two different blocks.

PR #1332, opened 2018-08-27
make psutil_debug() aware of PSUTIL_DEBUG

### LINKED ISSUE #1331: psutil_debug() function completely ignores PSUTIL_DEBUG variable
Despite the comment saying: "No-op if PSUTIL_DEBUG env var is not set."

yanok:
Fixes #1331


giampaolo:
Does this mean that right now debug messages are always printed on screen? Really?

yanok:
yep. There are not too many of them. In fact, I've only ever seen debugging messages about `task_for_pid()` call.

PR #1335, opened 2018-09-14
Fix random 0xC0000001 errors when querying for Connections

sylvainduchesne:
It seems as though this particular windows api can return an undocumented error code

also, cleaned up comments regarding releasing GIL

Error rate did drop drastically after initial fix, but it seems as though some unknown event on the OS causes seemingly random errors to popup. 

see 
https://social.msdn.microsoft.com/Forums/sqlserver/en-US/e8eaa270-8662-4202-bac4-2b4930b5bde9/getextendedtcptable-sometimes-returns-0xc0000001?forum=windowsgeneraldevelopmentissues

https://github.com/xilun/cbwin/issues/17
https://github.com/xilun/cbwin/blob/master/outbash/tcp_help.cpp#L67

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -548,7 +548,7 @@ N: Alex Manuskin
 W: https://github.com/amanusk
 I: 1284
 
-N: sylvainduchesne
+N: Sylvain Duchesne
 W: https://github.com/sylvainduchesne
 I: 1294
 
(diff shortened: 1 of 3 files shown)


giampaolo:
This looks reasonable. I would just add a comment on why `0xC0000001` is necessary. A link to this PR would also suffice. 

(also please update HISTORY and CREDITS files)

PR #1341, opened 2018-09-20
Refactored ps() function in test_posix

embray:
This is based off of some changes I originally introduced for my Cygwin branch (#998).

The idea is to make the `ps()` function used in `test_posix` a little more abstract, so that it provides just a single Python API for whatever use cases are necessary, and builds the correct platform-specific `ps` call from there.  This is opposed to `ps(...)` taking actual command-line flags for `ps` and manipulating them depending on the platform.

This will be especially useful on Cygwin which does have a `ps` command, but it is not POSIX-compatible.  But I thought it would be clearer, before adding the Cygwin-specific bits, to first propose this change on its own.

### DIFF
--- a/psutil/tests/test_linux.py
+++ b/psutil/tests/test_linux.py
@@ -1978,14 +1978,6 @@ class TestProcessAgainstStatus(unittest.TestCase):
 @unittest.skipIf(not LINUX, "LINUX only")
 class TestUtils(unittest.TestCase):
 
-    def test_open_text(self):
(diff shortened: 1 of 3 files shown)


giampaolo:
Good change. Thanks. Merging.

embray:
Uhh, I'm confused.  I think I might have made a mistake because I didn't intend to include that commit in this PR, I don't think.  Anyways that's fine I will just rebase in #1342.  

Thanks for accepting this change.  Indeed, I think it will make running these particular tests on other platforms easier.

Issue #1310, opened 2018-07-26
No Sensor Information Available using a Raspberry Pi


PolynomialDivision:
Hi,
if I call

    psutil.sensors_temperatures()

on a raspberry pi 3, the result is always

    {}

Is there no information about CPU temperature?

giampaolo:
Probably not due to `/sys/class/hwmon/hwmon*` directory structure not existing on raspberrypy (try to check)

vzorglub:
The directory structure '/sys/class/hwmon/hwmon*` does not exits on the raspberry pi.
The processor temperature can be obtained with:

`cat /sys/class/thermal/thermal_zone0/temp` in mC

I don't have a Pi2 or Pi3 to test multicore

giampaolo:
Re-opening. The goal should be to inspect `/sys/class/thermal/` tree in case `/sys/class/hwmon` does not exist.

amanusk:
@giampaolo I tried looking into this a few days ago.
There was a discussion about this in `lm-sensors`: [#30](https://github.com/lm-sensors/lm-sensors/issues/30), that can be used as a reference. 
Also:
* I could not find a way to obtain `high` and `critical` temperatures from `/sys/class/thermal/`
* Although raspberry-pi has 4 cores reported by `psutil.cpu_count()`, there is only a file for `sys/class/thermal/thermal_zone0/`

Other than that, I could give this a try :)

giampaolo:
Please do. If there's no way to obtain `high` and `critical` just set them to `None`.

PR #1345, opened 2018-09-27
Add parsing for /sys/class/thermal

amanusk:
This pull request addresses #1310. 

Basically, in case no files were found that match `/sys/class/hwmon`, which should have more information, look at `/sys/class/thermal`

When running `python scripts/temperatures.py`, (after "hiding" `/sys/class/hwmon` ) it returns:
On Thinkpad T430 + Ubuntu:
```
x86_pkg_temp
    x86_pkg_temp         53.0 °C (high = None °C, critical = None °C)

acpitz
    acpitz               46.0 °C (high = 103.0 °C, critical = 103.0 °C)

```

On raspberry pi 3 + Raspbian: 
```
cpu-thermal
    cpu-thermal          46.16 °C (high = None °C, critical = None °C)
```

Need to add unit tests for this

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1230,7 +1230,7 @@ def sensors_temperatures():
         ret[unit_name].append((label, current, high, critical))
 
     # Indication that no sensors were detected in /sys/class/hwmon/
-    if len(basenames) == 0:
+    if not basenames:
         basenames = glob.glob('/sys/class/thermal/thermal_zone*')
         basenames = sorted(set(basenames))
 
@@ -1255,10 +1255,10 @@ def sensors_temperatures():
                 trip_type = cat(path, fallback='', binary=False)
                 if trip_type == 'critical':
                     critical = cat(os.path.join(base, trip_point + "_temp"),
-                                   Fallback=None)
+                                   fallback=None)
                 elif trip_type == 'high':
                     high = cat(os.path.join(base, trip_point + "_temp"),
-                               Fallback=None)
+                               fallback=None)
 
                 if high is not None:
                     try:
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1229,6 +1229,49 @@ def sensors_temperatures():
 
         ret[unit_name].append((label, current, high, critical))
 
+    # Indication that no sensors were detected in /sys/class/hwmon/
+    if len(basenames) == 0:
+        basenames = glob.glob('/sys/class/thermal/thermal_zone*')
+        basenames = sorted(set(basenames))
+
+        for base in basenames:
+            try:
+                path = os.path.join(base, 'temp')
+                current = float(cat(path)) / 1000.0
+                path = os.path.join(base, 'type')
+                unit_name = cat(path, binary=False)
+            except (IOError, OSError, ValueError) as err:
+                warnings.warn("ignoring %r for file %r" % (err, path),
+                              RuntimeWarning)
+                continue
+
+            trip_paths = glob.glob(base + '/trip_point*')
+            trip_points = [os.path.basename(p) for p in trip_paths]
+            trip_points = sorted(set(["_".join(x.split('_')[0:3])
+                                      for x in trip_points]))
+            critical = None
+            high = None
+            for trip_point in trip_points:
+                path = os.path.join(base, trip_point + "_type")
+                trip_type = cat(path, fallback='', binary=False)
+                if trip_type == 'critical':
+                    critical = cat(os.path.join(base, trip_point + "_temp"))
+                elif trip_type == 'high':
+                    high = cat(os.path.join(base, trip_point + "_temp"))
+
+                if high is not None:
+                    try:
+                        high = float(high) / 1000.0
+                    except ValueError:
+                        high = None
+                if critical is not None:
+                    try:
+                        critical = float(critical) / 1000.0
+                    except ValueError:
+                        critical = None
+
+            ret[unit_name].append(('', current, high, critical))
+
     return ret
 
 
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1246,18 +1246,19 @@ def sensors_temperatures():
                 continue
 
             trip_paths = glob.glob(base + '/trip_point*')
-            trip_points = [os.path.basename(p) for p in trip_paths]
-            trip_points = sorted(set(["_".join(x.split('_')[0:3])
-                                      for x in trip_points]))
+            trip_points = set(['_'.join(
+                os.path.basename(p).split('_')[0:3]) for p in trip_paths])
             critical = None
             high = None
             for trip_point in trip_points:
                 path = os.path.join(base, trip_point + "_type")
                 trip_type = cat(path, fallback='', binary=False)
                 if trip_type == 'critical':
-                    critical = cat(os.path.join(base, trip_point + "_temp"))
+                    critical = cat(os.path.join(base, trip_point + "_temp"),
+                                   Fallback=None)
                 elif trip_type == 'high':
-                    high = cat(os.path.join(base, trip_point + "_temp"))
+                    high = cat(os.path.join(base, trip_point + "_temp"),
+                               Fallback=None)
 
                 if high is not None:
                     try:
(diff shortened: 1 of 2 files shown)


psutil/_pslinux.py
```diff
@@ -1229,6 +1229,49 @@ def sensors_temperatures():
 
         ret[unit_name].append((label, current, high, critical))
 
+    # Indication that no sensors were detected in /sys/class/hwmon/
+    if len(basenames) == 0:
```
giampaolo: You can simply spell it as `if not basenames:` instead

psutil/_pslinux.py
```diff
@@ -1229,6 +1229,49 @@ def sensors_temperatures():
...
+            trip_points = [os.path.basename(p) for p in trip_paths]
+            trip_points = sorted(set(["_".join(x.split('_')[0:3])
+                                      for x in trip_points]))
+            critical = None
+            high = None
+            for trip_point in trip_points:
+                path = os.path.join(base, trip_point + "_type")
+                trip_type = cat(path, fallback='', binary=False)
+                if trip_type == 'critical':
+                    critical = cat(os.path.join(base, trip_point + "_temp"))
```
giampaolo: please add fallback arg as in `cat(..., fallback=None)`

psutil/_pslinux.py
```diff
@@ -1229,6 +1229,49 @@ def sensors_temperatures():
...
+                elif trip_type == 'high':
+                    high = cat(os.path.join(base, trip_point + "_temp"))
```
giampaolo: please add fallback arg as in `cat(..., fallback=None)`

psutil/_pslinux.py
```diff
@@ -1229,6 +1229,49 @@ def sensors_temperatures():
...
+                unit_name = cat(path, binary=False)
+            except (IOError, OSError, ValueError) as err:
+                warnings.warn("ignoring %r for file %r" % (err, path),
+                              RuntimeWarning)
+                continue
+
+            trip_paths = glob.glob(base + '/trip_point*')
+            trip_points = [os.path.basename(p) for p in trip_paths]
+            trip_points = sorted(set(["_".join(x.split('_')[0:3])
+                                      for x in trip_points]))
```
giampaolo: Mmmm... I have mixed feelings about this. You make 2 for loops here and another 1 later for a total of 3. I wonder if there's a way to reduce the number of loops.

giampaolo (changes requested):
Nice work. I added some inline comments.

psutil/_pslinux.py
amanusk: `basenames = sorted(set([x.split('_')[0] for x in basenames]))` returns an empty list even if not files match, so `None` will not enter the `if`

psutil/_pslinux.py
```diff
@@ -1229,6 +1229,50 @@ def sensors_temperatures():
...
+                current = float(cat(path)) / 1000.0
+                path = os.path.join(base, 'type')
+                unit_name = cat(path, binary=False)
+            except (IOError, OSError, ValueError) as err:
+                warnings.warn("ignoring %r for file %r" % (err, path),
+                              RuntimeWarning)
+                continue
+
+            trip_paths = glob.glob(base + '/trip_point*')
+            trip_points = set(['_'.join(
```
amanusk: Squashed 2 for loops into one

psutil/_pslinux.py
```diff
@@ -1229,6 +1229,50 @@ def sensors_temperatures():
...
+            high = None
+            for trip_point in trip_points:
+                path = os.path.join(base, trip_point + "_type")
+                trip_type = cat(path, fallback='', binary=False)
+                if trip_type == 'critical':
+                    critical = cat(os.path.join(base, trip_point + "_temp"),
+                                   Fallback=None)
+                elif trip_type == 'high':
+                    high = cat(os.path.join(base, trip_point + "_temp"),
+                               Fallback=None)
```
amanusk: Added Fallback

amanusk (commented):
@giampaolo, applied your requested changes.

I would like to ask for your help in writing a test for this case. 
My understanding from this:
```
with mock.patch(patch_point, side_effect=open_mock):
            with mock.patch('glob.glob',
                            return_value=['/sys/class/hwmon/hwmon0/temp1']):
```
Is that `glob.glob` is replaces, and always returns `/sys/class/hwmon/hwmon0/temp1`.
In order to test this case, we need it to return `[ ]` the first time, and something like `['/sys/class/thermal/thermal_zone0']` the second time.  
I unfortunately don't know how to do this.

Or is there a completely different approach?

psutil/_pslinux.py
amanusk: Added

psutil/_pslinux.py
amanusk: Added

giampaolo:
Nice work! 

> I would like to ask for your help in writing a test for this case.

Sure. Not tested, just pseudo code, but it should give you an idea:

```python
def glob_mock(path):
    if path == ...:
        return ...
    else:
        return ...

with mock.patch('glob.glob', create=True, side_effect=glob_mock):
    ...
```

amanusk:
> Sure. Not tested, just pseudo code, but it should give you an idea
Thanks. It works. Added the test.

psutil/tests/test_linux.py
```diff
@@ -1466,6 +1466,8 @@ def test_emulate_eio_error(self):
         def open_mock(name, *args, **kwargs):
             if name.endswith("_input"):
                 raise OSError(errno.EIO, "")
+            if name.endswith("temp"):
```
giampaolo: `elif`

giampaolo (commented):
Thanks. I think it's better to split the two tests though that exercise the 2 `glob.glob` branches separately. Can you do that?

amanusk:
Ok, split the tests for /sys/class/hwmon and /sys/class/thermal

giampaolo (approved):
Excellent job. Thanks. Merging.

amanusk:
Thanks!
I would like to try and implement `sensors_temperatures()` for FreeBSD. Would this be interesting? Should I open a new issue for this?

amanusk:
I think it should be. For example:
`sysctl -n dev.cpu.0.temperature` returns the cpu temperature, and `sysctl -n dev.cpu.0.coretemp.tjmax` should be the `critical` temperature. 
I assume these can be called from `C` code, similarly to how `sensors_battery()` is implemented.

Cicicok:
There is a bug if you trying to get the CPU temperature and you are also using an external 1-Wire temperature sensor.

I'm running Raspberry zero with Raspbian Buster (Linux version 4.19.97+ (dom@buildbot) (gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0-88-g8460611))

The command `ll /sys/class/hwmon/hwmon*` shows 2 temperature sensors one is the external 1-Wire and the internal, which we want to read:
```
/sys/class/hwmon/hwmon0 -> ../../devices/w1_bus_master1/28-0316a5e4f7ff/hwmon/hwmon0
/sys/class/hwmon/hwmon1 -> ../../devices/platform/soc/soc:firmware/raspberrypi-hwmon/hwmon/hwmon1
```

1-Wire `ll /sys/class/hwmon/hwmon0/`:
```
device -> ../../../28-0316a5e4f7ff
name
power
subsystem -> ../../../../../class/hwmon
temp1_input
uevent
```
Internal `ll /sys/class/hwmon/hwmon1/`:
```
device -> ../../../raspberrypi-hwmon
in0_lcrit_alarm
name
power
subsystem -> ../../../../../../../class/hwmon
uevent
```

As you can see the external 1-Wire includes `temp1_input` but the internal one don't and that's the problem. As far as I understand the code looks at first into `/sys/class/hwmon/`
https://github.com/giampaolo/psutil/blob/ea4887e758ee2845f8e238655aaadb56f504e128/psutil/_pslinux.py#L1204-L1213
and after that if no temperature data has been found looks at `/sys/class/thermal/`:
https://github.com/giampaolo/psutil/blob/ea4887e758ee2845f8e238655aaadb56f504e128/psutil/_pslinux.py#L1247-L1249
But because the external 1-Wire provides `temp1_input` the code never reaches the part where is read `/sys/class/thermal/` including cpu temperature.

giampaolo:
With covid and everything I don't have time. Please make a PR.

Issue #1346, opened 2018-10-08
sunos10 - net_connections - regression 


alxchk:
```
-bash-3.2# python
Python 2.7.15 (default, Oct  8 2018, 14:49:06) 
[GCC 3.4.3 (csl-sol210-3_4-branch+sol_rpath)] on sunos5
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil 
>>> psutil.net_connections()
[]
>>> psutil.__version__
'5.4.7'
>>> 
```

```
-bash-3.2# python
Python 2.7.15 (default, Oct  8 2018, 14:55:52) 
[GCC 3.4.3 (csl-sol210-3_4-branch+sol_rpath)] on sunos5
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> len(psutil.net_connections())
112
>>> psutil.__version__
'5.4.3'
>>> 
```

giampaolo:
This was introduced by @gsauthof  in https://github.com/giampaolo/psutil/pull/1248. #1347 fixes it.

PR #1347, opened 2018-10-09
Fix https://github.com/giampaolo/psutil/issues/1346

alxchk:


giampaolo:
Thanks. Can you also update HISTORY and CREDITS files?

giampaolo:
Done.

PR #1350, opened 2018-10-20
FreeBSD adding temperature sensors (WIP)

amanusk:
The pull requests adds `sensors_temperatures()` capability for FreeBSD. This requires `coretemp` kernel module. 

This is a "WIP" pull request to make sure this is on the right track.

@giampaolo, some questions: 

* In this implementation, the `C` code only returns a single tuple with `current` and `max`, while creating the dictionary is handled in `python`. This makes it easier to iterate over the number of cpus, although this can also be done in `C`. I believe the number of errors can be smaller this way, and the number of `sysctlbyname` calls does not change. 

* Should `psutil_sensors_temperatures` be named something else?

* Is there a good open source tool to test pep7 compliance? 

Current implementation returns the following when running `scripts/temperatures.py`
```
coretemp
    Core 0               46 °C (high = 105 °C, critical = 105 °C)
    Core 1               46 °C (high = 105 °C, critical = 105 °C)
    Core 2               48 °C (high = 105 °C, critical = 105 °C)
    Core 3               48 °C (high = 105 °C, critical = 105 °C)
    Core 4               46 °C (high = 105 °C, critical = 105 °C)
    Core 5               46 °C (high = 105 °C, critical = 105 °C)
    Core 6               43 °C (high = 105 °C, critical = 105 °C)
    Core 7               43 °C (high = 105 °C, critical = 105 °C)
```

### DIFF
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -11,6 +11,7 @@ import os
 import xml.etree.ElementTree as ET
 from collections import namedtuple
 from socket import AF_INET
+from collections import defaultdict
 
 from . import _common
 from . import _psposix
@@ -433,18 +434,19 @@ if FREEBSD:
         return _common.sbattery(percent, secsleft, power_plugged)
 
     def sensors_temperatures():
-        """Return systemp temperatures"""
-        ret = dict()
-        ret["coretemp"] = list()
+        "Return CPU cores temperatures if available, else an empty dict."
+        ret = defaultdict(list)
         num_cpus = cpu_count_logical()
-        try:
-            for cpu in range(num_cpus):
-                current, tjmax = cext.sensors_temperatures(cpu)
-                name = "Core {}".format(cpu)
+        for cpu in range(num_cpus):
+            try:
+                current, high = cext.sensors_cpu_temperature(cpu)
+                if high <= 0:
+                    high = None
+                name = "Core %s" % cpu
                 ret["coretemp"].append(
-                    _common.shwtemp(name, current, tjmax, tjmax))
-        except NotImplementedError:
-            return None
+                    _common.shwtemp(name, current, high, high))
+            except NotImplementedError:
+                pass
 
         return ret
 
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -432,6 +432,22 @@ if FREEBSD:
             secsleft = minsleft * 60
         return _common.sbattery(percent, secsleft, power_plugged)
 
+    def sensors_temperatures():
+        """Return systemp temperatures"""
+        ret = dict()
+        ret["coretemp"] = list()
+        num_cpus = cpu_count_logical()
+        try:
+            for cpu in range(num_cpus):
+                current, tjmax = cext.sensors_temperatures(cpu)
+                name = "Core {}".format(cpu)
+                ret["coretemp"].append(
+                    _common.shwtemp(name, current, tjmax, tjmax))
+        except NotImplementedError:
+            return None
+
+        return ret
+
 
 # =====================================================================
 #  --- other system functions
--- a/psutil/_psutil_bsd.c
+++ b/psutil/_psutil_bsd.c
@@ -981,8 +981,8 @@ PsutilMethods[] = {
 #if defined(PSUTIL_FREEBSD)
     {"sensors_battery", psutil_sensors_battery, METH_VARARGS,
      "Return battery information."},
-    {"sensors_temperatures", psutil_sensors_temperatures, METH_VARARGS,
-     "Return temperatures information."},
+    {"sensors_cpu_temperature", psutil_sensors_cpu_temperature, METH_VARARGS,
+     "Return temperature information for a given CPU core number."},
 #endif
 
     // --- others
--- a/psutil/_psutil_bsd.c
+++ b/psutil/_psutil_bsd.c
@@ -981,6 +981,8 @@ PsutilMethods[] = {
 #if defined(PSUTIL_FREEBSD)
     {"sensors_battery", psutil_sensors_battery, METH_VARARGS,
      "Return battery information."},
+    {"sensors_temperatures", psutil_sensors_temperatures, METH_VARARGS,
+     "Return temperatures information."},
 #endif
 
     // --- others
--- a/psutil/arch/freebsd/specific.c
+++ b/psutil/arch/freebsd/specific.c
@@ -31,7 +31,7 @@
 #define PSUTIL_TV2DOUBLE(t)    ((t).tv_sec + (t).tv_usec / 1000000.0)
 #define PSUTIL_BT2MSEC(bt) (bt.sec * 1000 + (((uint64_t) 1000000000 * (uint32_t) \
         (bt.frac >> 32) ) >> 32 ) / 1000000)
-#define SYSCTLTEMP(t) ((int)((t + 270)) - 3000) / 10
+#define DECIKELVIN_2_CELCIUS(t) (t - 2731) / 10
 #ifndef _PATH_DEVNULL
 #define _PATH_DEVNULL "/dev/null"
 #endif
@@ -1014,10 +1014,10 @@ error:
 
 
 /*
- * Return temperature information.
+ * Return temperature information for a given CPU core number.
  */
 PyObject *
-psutil_sensors_temperatures(PyObject *self, PyObject *args) {
+psutil_sensors_cpu_temperature(PyObject *self, PyObject *args) {
     int current;
     int tjmax;
     int core;
@@ -1029,13 +1029,13 @@ psutil_sensors_temperatures(PyObject *self, PyObject *args) {
     sprintf(sensor, "dev.cpu.%d.temperature", core);
     if (sysctlbyname(sensor, &current, &size, NULL, 0))
         goto error;
-    current = SYSCTLTEMP(current);
+    current = DECIKELVIN_2_CELCIUS(current);
 
+    // Return -273 in case of faliure.
     sprintf(sensor, "dev.cpu.%d.coretemp.tjmax", core);
     if (sysctlbyname(sensor, &tjmax, &size, NULL, 0))
-        goto error;
-    tjmax = SYSCTLTEMP(tjmax);
-
+        tjmax = 0;
+    tjmax = DECIKELVIN_2_CELCIUS(tjmax);
 
     return Py_BuildValue("ii", current, tjmax);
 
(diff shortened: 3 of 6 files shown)


psutil/arch/freebsd/specific.c
```diff
@@ -1010,3 +1011,38 @@ psutil_sensors_battery(PyObject *self, PyObject *args) {
...
+    if (! PyArg_ParseTuple(args, "i", &core))
+        return NULL;
+    sprintf(sensor, "dev.cpu.%d.temperature", core);
+    if (sysctlbyname(sensor, &current, &size, NULL, 0))
+        goto error;
+    current = SYSCTLTEMP(current);
+
+    sprintf(sensor, "dev.cpu.%d.coretemp.tjmax", core);
+    if (sysctlbyname(sensor, &tjmax, &size, NULL, 0))
+        goto error;
```
giampaolo: Considering we can have a valid "current" temp I think it's better to ignore this error and set "max" temp to `0`. Then from Python you can check if it's `0` and turn it into `None`.

psutil/_psbsd.py
```diff
@@ -432,6 +432,22 @@ def sensors_battery():
...
         return _common.sbattery(percent, secsleft, power_plugged)
 
+    def sensors_temperatures():
+        """Return systemp temperatures"""
+        ret = dict()
+        ret["coretemp"] = list()
+        num_cpus = cpu_count_logical()
+        try:
+            for cpu in range(num_cpus):
+                current, tjmax = cext.sensors_temperatures(cpu)
```
giampaolo: I think it's better to catch `NotImplementedError` here instead of against the whole `for` loop. That way if there are temperatures for certain CPUs only (instead of all of them) we can at least return a partial result.

Also `tmax` looks like a better name than `tjmax`.

psutil/arch/freebsd/specific.c
```diff
@@ -1010,3 +1011,38 @@ psutil_sensors_battery(PyObject *self, PyObject *args) {
         PyErr_SetFromErrno(PyExc_OSError);
     return NULL;
 }
+
+
+/*
+ * Return temperature information.
+ */
+PyObject *
+psutil_sensors_temperatures(PyObject *self, PyObject *args) {
```
giampaolo: I would rename this to `psutil_sensors_cpu_temperature`

psutil/_psbsd.py
```diff
@@ -432,6 +432,22 @@ def sensors_battery():
             secsleft = minsleft * 60
         return _common.sbattery(percent, secsleft, power_plugged)
 
+    def sensors_temperatures():
+        """Return systemp temperatures"""
```
giampaolo: I would rephrase this as: "Return CPU cores temperatures if available, else an empty dict."

psutil/_psbsd.py
```diff
@@ -432,6 +432,22 @@ def sensors_battery():
...
+        ret = dict()
+        ret["coretemp"] = list()
```
giampaolo: This is not correct because if there are no temperatures you will return a `{"coretemp": []}` dict which will evaluate to `True`. Instead define ret as:

`ret = collections.defaultdict(list)`

...and when you add elements to it do:

`ret["coretemp"].append(...)`

Finally, at the end of the function transform it to a plain dict with:

`return dict(ret)`

psutil/arch/freebsd/specific.c
```diff
@@ -1010,3 +1011,38 @@ psutil_sensors_battery(PyObject *self, PyObject *args) {
         PyErr_SetFromErrno(PyExc_OSError);
     return NULL;
 }
+
+
+/*
+ * Return temperature information.
```
giampaolo: "Return temperature information for a given CPU core number."

psutil/_psutil_bsd.c
```diff
@@ -981,6 +981,8 @@ PsutilMethods[] = {
 #if defined(PSUTIL_FREEBSD)
     {"sensors_battery", psutil_sensors_battery, METH_VARARGS,
      "Return battery information."},
+    {"sensors_temperatures", psutil_sensors_temperatures, METH_VARARGS,
+     "Return temperatures information."},
```
giampaolo: "Return temperature information for a given CPU core number."

psutil/_psbsd.py
```diff
@@ -432,6 +432,22 @@ def sensors_battery():
...
+        num_cpus = cpu_count_logical()
+        try:
+            for cpu in range(num_cpus):
+                current, tjmax = cext.sensors_temperatures(cpu)
+                name = "Core {}".format(cpu)
```
giampaolo: Please use `name = "Core %s" % cpu` instead in order to be consistent with the rest of code base.

psutil/arch/freebsd/specific.c
```diff
@@ -31,6 +31,7 @@
 #define PSUTIL_TV2DOUBLE(t)    ((t).tv_sec + (t).tv_usec / 1000000.0)
 #define PSUTIL_BT2MSEC(bt) (bt.sec * 1000 + (((uint64_t) 1000000000 * (uint32_t) \
         (bt.frac >> 32) ) >> 32 ) / 1000000)
+#define SYSCTLTEMP(t) ((int)((t + 270)) - 3000) / 10
```
giampaolo: What's the logic here? Where did you take this? Are we returning Celsius? If there's some source code of reference it would be worth mentioning it here as a comment.

giampaolo (changes requested):
Thanks Alex! This really is awesome work! 
I added some comments inline. Other than that it would be nice to have some tests. In details what I would do is:

- in `test_bsd.py` run `sysctl` cmdline tool, get the numbers of `"dev.cpu.%d.temperature"` entries and assert they are equal to `len(psutil.sensors_temperatures())`
- for every `psutil.sensors_temperatures()` check "current" against "dev.cpu.%d.temperature" with a `10` units tolerance or something and "max" against `"dev.cpu.%d.coretemp.tjmax"` (no tolerance)

> This requires coretemp kernel module.

What does this mean? Is it something you have to install? If so it may be worth mentioning it into doc.

> Is there a good open source tool to test pep7 compliance?

Not that I'm aware of but it's worth having such a tool integrated in our workflow. For now I created https://github.com/giampaolo/psutil/issues/1351 to keep track of it.

amanusk:
@giampaolo, Thanks for the review. I wanted to make sure this is feasible before continuing work on this. (At least no licensing issues :smile: ) 

> Other than that it would be nice to have some tests

Sure!

> What does this mean? Is it something you have to install? If so it may be worth mentioning it into doc.

I am not much of an expert in FreeBSD myself, I got the idea for this PR from here: https://www.cyberciti.biz/faq/freebsd-determine-processor-cpu-temperature-command/ 


The kernel module needs to be enabled, but there is no need for additional installations. Definitely worth mentioning in the Doc.

psutil/arch/freebsd/specific.c
amanusk: Yeah.. this bit kinda looks like a hack.. Explanation: 

Calling `sysctlbyname` returns an integer. Usually something like this: `3121`. 

Always ending with `1`, and starts with `3` for the temperatures I have observed. 

By comparing this with `sysctl` calls, I was able to determine that the actual temperature is the middle 2 digits (in this case `12`) with and offset of `27`, so in this case: `39`. 

I wasn't able go find official documentation of this, so if you or a FreeBSD expert have a better idea of how to get actual values, that would be great. 

Link to `coretemp` implementation: 
https://github.com/freebsd/freebsd/blob/master/sys/dev/coretemp/coretemp.c

giampaolo: I see. It looks like coretemp.c implements the conversion logic like this: 
https://github.com/freebsd/freebsd/blob/b3f6f192fe65a06ad95d7d6eef0c58a789e7326f/sys/dev/coretemp/coretemp.c#L372-L375
Ideally I think it would be better to do the same but practically I'm not sure if the same logic can be applied to your code.
I'm CC-ing @glebius, @sunpoet, @kostikbel just in case they know more about this.

glebius: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3b-part-2-manual.pdf

Paragraph 14.8 there. Let me read myself :) and come with review .

giampaolo: Thanks @glebius.
Here's how netdata implements the conversion logic:
https://github.com/netdata/netdata/blob/545d07eb78222a80124e9187e88921013ee403ea/collectors/freebsd.plugin/freebsd_sysctl.c#L520

glebius: Okay, I reviewed and I don't understand why do you want to do conversion in the python module? The sysctl already returns Celsius. Looks like on your box it returns some strange data. What version of FreeBSD do you use @amanusk and what is the hardware?

amanusk: @glebius, Thanks for you help with this.
`freebsd-version -k` returns `11.2-RELEASE`

System is Thinkpad T430, with i7-3740QM 

Running `sysctl dev.cpu.0.temperature` in Shell returns: `dev.cpu.0.temperature: 42.0C`  which is indeed in Celsius. The bit where I need to convert is line 1030 in the discussed file (sorry for not knowing how to link line). It returns a 4 digit integer as explained in this conversion.

psutil/arch/freebsd/specific.c
```diff
@@ -1010,3 +1011,38 @@ psutil_sensors_battery(PyObject *self, PyObject *args) {
...
+    int current;
+    int tjmax;
+    int core;
+    char sensor[26];
+    size_t size = sizeof(current);
+
+    if (! PyArg_ParseTuple(args, "i", &core))
+        return NULL;
+    sprintf(sensor, "dev.cpu.%d.temperature", core);
+    if (sysctlbyname(sensor, &current, &size, NULL, 0))
```
amanusk: @glebius: This call returns a 4 digit integer, as explained.

glebius: I reproduced that. Looks strange indeed. Investigating.

glebius: I see. The sysctl has special format. It reports in deciKelvin. Here is example parser:

https://github.com/freebsd/freebsd/blob/master/sbin/sysctl/sysctl.c#L790

Just copy math from there. Your math is almost correct.

amanusk:
@giampaolo, sorry for the long delay. 
I have implemented the changes and the tests.

giampaolo:
Thanks a lot Alex! Merging.

Issue #1352, opened 2018-10-20
Get CPU frequency on FreeBSD 


giampaolo:
It appears it's possible:
- https://github.com/netdata/netdata/blob/545d07eb78222a80124e9187e88921013ee403ea/collectors/freebsd.plugin/freebsd_sysctl.c#L538
- https://lists.freebsd.org/pipermail/freebsd-stable/2015-March/082042.html

On my virtualized FreeBSD 11 box though it appears the frequency metrics are not available:

```
vagrant@freebsd11~$  sysctl dev.cpu
dev.cpu.1.cx_method: C1/hlt
dev.cpu.1.cx_usage_counters: 9540
dev.cpu.1.cx_usage: 100.00% last 33074us
dev.cpu.1.cx_lowest: C2
dev.cpu.1.cx_supported: C1/1/0
dev.cpu.1.%parent: acpi0
dev.cpu.1.%pnpinfo: _HID=none _UID=0
dev.cpu.1.%location: handle=\_PR_.CPU1
dev.cpu.1.%driver: cpu
dev.cpu.1.%desc: ACPI CPU
dev.cpu.0.cx_method: C1/hlt
dev.cpu.0.cx_usage_counters: 17748
dev.cpu.0.cx_usage: 100.00% last 17867us
dev.cpu.0.cx_lowest: C2
dev.cpu.0.cx_supported: C1/1/0
dev.cpu.0.%parent: acpi0
dev.cpu.0.%pnpinfo: _HID=none _UID=0
dev.cpu.0.%location: handle=\_PR_.CPU0
dev.cpu.0.%driver: cpu
dev.cpu.0.%desc: ACPI CPU
dev.cpu.%parent: 
```

Not sure if I have to enable them similarly to CPU temperatures (https://www.cyberciti.biz/faq/freebsd-determine-processor-cpu-temperature-command/). @glebius, @sunpoet, @kostikbel  any idea? I'm also CC-ing @amanusk in case you want to give this a try.

amanusk:
I can give this a try. From the code on `netdata` and this answer: https://stackoverflow.com/questions/9941664/how-to-get-real-cpu-frequency-while-system-running, it appears that only the information for cpu 0 is available, and it must by the same frequency for all CPUs.

This can also be enabled, similar to temperatures.

giampaolo:
Nice. How do you enable it?

amanusk:
[...]

PR #1369, opened 2018-11-30
Add CPU frequency support for FreeBSD

amanusk:
This adds support for CPU frequency on FreeBSD. (Addresses #1352)
An example `sysctl` call looks like this:

```
dev.cpu.0.freq_levels: 2701/45000 2700/45000 2600/42708 2500/40461 2400/38257 2300/36525 2200/34401 2100/32327 1900/28698 1800/26736 1700/24815 1600/23334 1500/21489 1400/19683 1300/18298 1200/16566
dev.cpu.0.freq: 1200
```

Frequency reading is only available for core 0, as explained here: https://stackoverflow.com/questions/9941664/how-to-get-real-cpu-frequency-while-system-running

The available frequency levels are in the format `(frequency/power usage)` (as per https://www.unix.com/man-page/FreeBSD/4/cpufreq/). 
To my knowledge, in this example the top frequency level is `2701`, and the one after it is `2700`. The `1` indicates that _Turbo_ frequency is enabled, but I do not know of a way to get the actual frequency (obviously this should more than 1 MHz :) . On this system the _Turbo_ frequency is 3.3GHz when running on all cores)

Example command line:
```
[~/psutil]> python3 -c "import psutil; print(psutil.cpu_freq())"
scpufreq(current=2701, min='1200', max='2701')
```

(I will add tests after initial review)

### DIFF
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -456,25 +456,29 @@ if FREEBSD:
         return ret
 
     def cpu_freq():
+        """
+        Return frequency metrics for CPUs. Currently, only CPU 0 is supported
+        by FreeBSD, all other cores match the frequency of CPU 0.
+        """
         ret = []
         num_cpus = cpu_count_logical()
         for cpu in range(num_cpus):
             try:
-                current, available = cext.cpu_frequency(cpu)
-                low = None
-                high = None
-                if available:
-                    try:
-                        low = int(available.split(" ")[-1].split("/")[0])
-                    except(IndexError, ValueError):
-                        pass
-                    try:
-                        high = int(available.split(" ")[0].split("/")[0])
-                    except(IndexError, ValueError):
-                        pass
-                ret.append(_common.scpufreq(current, low, high))
+                current, available_freq = cext.cpu_frequency(cpu)
             except NotImplementedError:
-                pass
+                continue
+            min_freq = None
+            max_freq = None
+            if available_freq:
+                try:
+                    min_freq = int(available_freq.split(" ")[-1].split("/")[0])
+                except(IndexError, ValueError):
+                    pass
+                try:
+                    max_freq = int(available_freq.split(" ")[0].split("/")[0])
+                except(IndexError, ValueError):
+                    pass
+            ret.append(_common.scpufreq(current, min_freq, max_freq))
         return ret
 
 
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -455,6 +455,28 @@ if FREEBSD:
 
         return ret
 
+    def cpu_freq():
+        ret = []
+        num_cpus = cpu_count_logical()
+        for cpu in range(num_cpus):
+            try:
+                current, available = cext.cpu_frequency(cpu)
+                low = None
+                high = None
+                if available:
+                    try:
+                        low = available.split(" ")[-1].split("/")[0]
+                    except IndexError:
+                        pass
+                    try:
+                        high = available.split(" ")[0].split("/")[0]
+                    except IndexError:
+                        pass
+                ret.append(_common.scpufreq(current, low, high))
+            except NotImplementedError:
+                pass
+        return ret
+
 
 # =====================================================================
 #  --- other system functions
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -465,12 +465,12 @@ if FREEBSD:
                 high = None
                 if available:
                     try:
-                        low = available.split(" ")[-1].split("/")[0]
-                    except IndexError:
+                        low = int(available.split(" ")[-1].split("/")[0])
+                    except(IndexError, ValueError):
                         pass
                     try:
-                        high = available.split(" ")[0].split("/")[0]
-                    except IndexError:
+                        high = int(available.split(" ")[0].split("/")[0])
+                    except(IndexError, ValueError):
                         pass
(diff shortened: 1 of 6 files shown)


psutil/_psbsd.py
```diff
@@ -455,6 +455,28 @@ def sensors_temperatures():
 
         return ret
 
+    def cpu_freq():
+        ret = []
+        num_cpus = cpu_count_logical()
+        for cpu in range(num_cpus):
+            try:
+                current, available = cext.cpu_frequency(cpu)
```
giampaolo: This try/except block is too broad. It should only wrap the `cext.cpu_frequency(cpu)` line. 
Also consider renaming `available` to `minmax` or something.

psutil/_psbsd.py
```diff
@@ -455,6 +455,28 @@ def sensors_temperatures():
...
+                low = None
+                high = None
+                if available:
+                    try:
+                        low = int(available.split(" ")[-1].split("/")[0])
```
giampaolo: Mmm. How does the string (`available`) look like in here?

psutil/_psutil_bsd.c
```diff
@@ -983,6 +983,8 @@ PsutilMethods[] = {
      "Return battery information."},
     {"sensors_cpu_temperature", psutil_sensors_cpu_temperature, METH_VARARGS,
      "Return temperature information for a given CPU core number."},
+    {"cpu_frequency", psutil_cpu_freq, METH_VARARGS,
+     "Return cpu 0 frequency"},
```
giampaolo: Shouldn't this be `"Return frequency of a given CPU"`?

psutil/arch/freebsd/specific.c
```diff
@@ -1046,3 +1046,38 @@ psutil_sensors_cpu_temperature(PyObject *self, PyObject *args) {
         PyErr_SetFromErrno(PyExc_OSError);
     return NULL;
 }
+
+
+/*
+ * Return cpu 0 frequency information
```
giampaolo: Same here ("Return frequency of a given CPU number")

psutil/arch/freebsd/specific.c
```diff
@@ -1046,3 +1046,38 @@ psutil_sensors_cpu_temperature(PyObject *self, PyObject *args) {
...
+    size_t size = sizeof(current);
+
+    if (! PyArg_ParseTuple(args, "i", &core))
+        return NULL;
+    sprintf(sensor, "dev.cpu.%d.freq", core);
+    if (sysctlbyname(sensor, &current, &size, NULL, 0))
+        goto error;
+
+    size = sizeof(available);
+    // In case of faliure, an empty string is returned
```
giampaolo: typo (faliure)

psutil/arch/freebsd/specific.c
```diff
@@ -1046,3 +1046,38 @@ psutil_sensors_cpu_temperature(PyObject *self, PyObject *args) {
...
+ * Return cpu 0 frequency information
+ */
+PyObject *
+psutil_cpu_freq(PyObject *self, PyObject *args) {
+    int current;
+    int max;
+    int core;
+    char sensor[26];
+    char available[1000];
+    char* answer;
```
giampaolo: It appears this is not used.

psutil/arch/freebsd/specific.c
```diff
@@ -1046,3 +1046,38 @@ psutil_sensors_cpu_temperature(PyObject *self, PyObject *args) {
...
+/*
+ * Return cpu 0 frequency information
+ */
+PyObject *
+psutil_cpu_freq(PyObject *self, PyObject *args) {
+    int current;
+    int max;
+    int core;
+    char sensor[26];
+    char available[1000];
```
giampaolo: Maybe it makes more sense to rename this to `minmax` or something (assuming that's what the string represents, which I don't know).

psutil/_psbsd.py
amanusk: The `available` line looks like this:
`2701/45000 2700/45000 2600/42708 2500/40461 2400/38257 2300/36525 2200/34401 2100/32327 1900/28698 1800/26736 1700/24815 1600/23334 1500/21489 1400/19683 1300/18298 1200/16566`

It represents all the available levels of frequency the CPU can be in, not just min/max, that's why I thought it would be a better name

psutil/arch/freebsd/specific.c
amanusk: Fixed

psutil/arch/freebsd/specific.c
amanusk: :+1:

psutil/arch/freebsd/specific.c
amanusk: The string represents all available CPU frequency levels

psutil/_psutil_bsd.c
amanusk: :+1:

giampaolo:
Thanks Alex.

davidbrochart:
I have a machine where:
- `/sys/devices/system/cpu/cpufreq/` exists but is empty.
- `/sys/devices/system/cpu/cpu0/' exists but doesn't have `cpufreq` in it.

Hence we will not fall back to:
https://github.com/giampaolo/psutil/blob/c0aba35a78649c453f0c89ab163a58a8efb4639e/psutil/_pslinux.py#L705-L716

PR #1371, opened 2018-12-04
Add Windows support to /scripts/top.py

jdgregson:
While I realize that the files in /scripts are examples and not critical parts of the project, I feel that it's important for examples of a cross-platform library to themselves be cross-platform. This PR addresses the following Windows compatibility issues with /scripts/top.py:

- os.getloadavg() is not available on Windows, causing the script to throw an unhandled exception.
- Nice values are provided as strings like 'None' and 'Priority.NORMAL_PRIORITY_CLASS'.
- Control+C is implemented differently on Windows which prevents the user form terminating the script.
- As a bonus, 'q' to quit has been implemented on Windows and Linux.

### DIFF
--- a/scripts/top.py
+++ b/scripts/top.py
@@ -99,6 +99,18 @@ def bytes2human(n):
     return "%sB" % n
 
 
+def niceness2priority(niceness):
+    if not niceness:
+        niceness == 'None'
+    elif str(niceness) == 'Priority.HIGH_PRIORITY_CLASS':
+        niceness = 'High'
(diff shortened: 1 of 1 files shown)


giampaolo:
Hello. scripts/top.py is not supposed to work on Windows because `curses` module is not available (for me at least). Are you able to import curses module on Windows?

giampaolo:
(looking back at this 4 years later)

I'm going to reject this because it adds too much complexity for little value (IMO). Scripts in `scripts/*.py` are intended to show how some APIs can be used and not much more than that. I prefer to keep them simple at the cost of sacrificing Windows support.

Issue #1379, opened 2018-12-11
[Windows] Cannot resume process suspended by another program


ogurets:
If an app like Process Explorer suspends the process, psutil cannot resume it because it operates on threads.
I've rewritten psutil_proc_suspend_or_resume routine using ntdll functions as it is the only method I found for my WinXP OS. And I guess the guys from SysInternals are doing it the same way.

The patch is here - https://github.com/ogurets/psutil/commit/9e8123167e6479fef4442bbfd3e50a290260552b
Anyone willing to test it out on Windows 7, 8 or 10?

giampaolo:
Mmm... interesting. Some questions / considerations:

1) What do you mean by "psutil cannot resume processes"? What happens when you do it (error?)? 
2) `NtSuspendProcess` and `NtResumeProcess` are not documented hence unofficial - not sure how big of a problem this actually is.
3) It seems there are at least 3 methods to achieve process suspend/resume. See [here](https://stackoverflow.com/questions/11010165/how-to-suspend-resume-a-process-in-windows) and [here](https://j00ru.vexillium.org/2009/08/suspending-processes-in-windows/). These 2 links point out the risks of using the current method based on thread suspension/resumption (which I wasn't aware of). As such it could be desirable to change it. Point is figuring out what's the better method to use amongst those 3
4) how are you using psutil on Windows XP? What version is it? Last thing I knew about Windows XP support in psutil was that the binary wheels didn't work (because are compiled on Windows 7) but compiling from sources did. I recently committed some code (PR ##1376) which should definitively break win XP support but perhaps I may consider adding a workaround if psutil still (kind of) works on Win XP.

ogurets:
1. The process stays suspended without any errors. Looks like the process has it's own "pause" flag, which overrides the threads' flags. I.e. if the process is "suspended", but threads are "running", nothing is actually running.
Suppose we have a busy process with 4 threads, consuming 100% of a 4-core CPU. It's started by conventional means and is running freely. Then:
   - Process suspended with `Process Explorer` (CPU usage 100%->0%), resumed with `psutil` - nothing happens (CPU usage 0%).
   - Suspended with `psutil` (CPU usage 100%->0%), resumed with `psutil` (usage 0%->100%).
   - Not sure what happens if suspended with `psutil` and resumed with `Process Explorer`. I vaguely remember that it still works.

2. Oh yes, thought about that too.
[...]

PR #1382, opened 2018-12-14
#1379: added ntdll method for suspending/resuming processes

ogurets:
Slightly modified patch for #1379, which will use threading method if `NtSuspendProcess` or `NtResumeProcess` are unavailable.

PR #1383, opened 2018-12-15
#1379: added ntdll method for suspending/resuming processes

ogurets:
Slightly modified patch for #1379, which will use threading method if NtSuspendProcess or NtResumeProcess are unavailable.


giampaolo:
Closing this one as a duplicate of #1382 as I don't think keeping the alternate method to support Windows XP is worth it. Old code relying on `CreateToolhelp32Snapshot` should just go.

PR #1397, opened 2019-01-26
Add parents method

gigi206:
Add the parents method :
```
>>> import psutil
>>> from pprint import pprint
>>> p=psutil.Process()
>>> pprint(p.parents())
[psutil.Process(pid=3943, name='gnome-terminal-server', started='16:20:34'),
 psutil.Process(pid=2172, name='systemd', started='16:17:26'),
 psutil.Process(pid=1, name='systemd', started='16:17:07')]
```

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -557,6 +557,18 @@ class Process(object):
             except NoSuchProcess:
                 pass
 
+    def parents(self):
+        """Return the parents process as a list of Process
+        instances, pre-emptively checking whether PID has been reused.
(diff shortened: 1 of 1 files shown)


giampaolo:
Hi and sorry for replying so late. This is a nice feature which I am tempted to include. It lacks tests though and I also have to think about other details (e.g. `as_dict()` should probably not return this and some tests should skip this).

giampaolo:
I opened #1433 which includes test and doc updates.

Issue #1384, opened 2018-12-19
Windows >= 8.1 : get commandline through NtQueryInformationProcess + ProcessCommandLineInformation


EccoTheFlintstone:
Starting from windows 8.1 and protected processes, it' impossible to get `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ` on certain processes (protected processes for instance)
psutil retrieves the command line of processes by reading the PEB of those processes (thus needing `PROCESS_VM_READ`)

Starting from windows 8.1, it's possible to get the commandline of a process using  `NtQueryInformationProcess` with `ProcessCommandLineInformation` class

To use this, only `PROCESS_QUERY_LIMITED_INFORMATION` rights are needed on the process

discussion about this here : 
https://wj32.org/processhacker/forums/viewtopic.php?t=2760

implementation by process hacker here:
https://github.com/processhacker/processhacker/blob/master/phlib/native.c#L721

giampaolo:
Interesting. I suppose that means less AccessDenied exceptions. Thanks for digging into this. I'm not sure when I'll have time to look into this (definitively not this year) so if you're interested in working on this be my guest. =)

giampaolo:
Closing out as fixed.

PR #1398, opened 2019-01-26
Starting from windows 8.1, get commandline content using NtQueryInformationProcess (see #1384)

EccoTheFlintstone:
- had to fix enable_se_debug that seemed not to work (access to protected processes info does not work with old method, works with enabling se_debug on the whole process token (and not only 1 thread))

- starting from windows 8.1, get commandline value using NtQueryInformationProcess + ProcessCommandLineInformation (see #1384), this way we can get commandline for protected processes (smss.exe, msmpeng.exe, ...) instead of getting AccessDenied
Do this only if PEB read fails (by default, still use PEB read to be able to get actual process arguments, see https://blog.xpnsec.com/how-to-argue-like-cobalt-strike/)

### DIFF
--- a/psutil/arch/windows/process_info.c
+++ b/psutil/arch/windows/process_info.c
@@ -864,23 +864,25 @@ int psutil_get_cmdline_data(long pid, WCHAR **pdata, SIZE_T *psize) {
     PUNICODE_STRING tmp = NULL;
     DWORD string_size;
     _NtQueryInformationProcess NtQueryInformationProcess = NULL;
+    int ret = -1;
 
     NtQueryInformationProcess = psutil_NtQueryInformationProcess();
     if (NtQueryInformationProcess == NULL) {
         PyErr_SetFromWindowsErr(0);
-        return -1;
+        goto error;
     }
 
     cmdline_buffer = calloc(ret_length, 1);
     if (cmdline_buffer == NULL) {
         PyErr_NoMemory();
-        return -1;
+        goto error;
     }
 
     hProcess = psutil_handle_from_pid(pid, PROCESS_QUERY_LIMITED_INFORMATION);
     if (hProcess == NULL) {
         // psutil_handle_from_pid sets errorcode/exception, don't need to do it it
-        return -1;
+        PyErr_SetFromWindowsErr(0);
+        goto error;
     }
     status = NtQueryInformationProcess(
         hProcess,
@@ -893,26 +895,30 @@ int psutil_get_cmdline_data(long pid, WCHAR **pdata, SIZE_T *psize) {
         // set error before closing handle to keep original error
         // CloseHandle might fail and set a new errno/GetLastError
         PyErr_SetFromWindowsErr(0);
-        CloseHandle(hProcess);
-        return -1;
+        goto error;
     }
-    CloseHandle(hProcess);
+
     tmp = (PUNICODE_STRING)cmdline_buffer;
     string_size = wcslen(tmp->Buffer) + 1;
     cmdline_buffer_wchar = (WCHAR *)calloc(string_size, sizeof(WCHAR));
 
     if (cmdline_buffer_wchar == NULL) {
-        free(cmdline_buffer);
         PyErr_NoMemory();
-        return -1;
+        goto error;
     }
 
     wcscpy_s(cmdline_buffer_wchar, string_size, tmp->Buffer);
     *pdata = cmdline_buffer_wchar;
     *psize = string_size * sizeof(WCHAR);
-    free(cmdline_buffer);
+    ret = 0;
 
-    return 0;
+error:
+    if (cmdline_buffer != NULL)
+        free(cmdline_buffer);
+    if (hProcess != NULL)
+        CloseHandle(hProcess);
+
+    return ret;
 }
 
 /*
--- a/psutil/arch/windows/process_info.c
+++ b/psutil/arch/windows/process_info.c
@@ -906,7 +906,7 @@ psutil_get_cmdline(long pid) {
                     string_size = wcslen(tmp->Buffer) + 1;
                     cmdline_buffer_wchar = (WCHAR *)calloc(string_size, sizeof(WCHAR));
                     if (cmdline_buffer_wchar != NULL) {
-                        wcscpy_s(cmdline_buffer_wchar, wcslen(string_size, tmp->Buffer);
+                        wcscpy_s(cmdline_buffer_wchar, string_size, tmp->Buffer);
                         data = cmdline_buffer_wchar;
                         size = string_size * sizeof(WCHAR);
                     }
--- a/psutil/arch/windows/process_info.c
+++ b/psutil/arch/windows/process_info.c
@@ -948,22 +948,23 @@ psutil_get_cmdline(long pid) {
     (see here : https://blog.xpnsec.com/how-to-argue-like-cobalt-strike/)
     */
     func_ret = psutil_get_process_data(pid, KIND_CMDLINE, &data, &size);
-    if ((ret != 0) &&
-        (GetLastError() == ERROR_ACCESS_DENIED) &&
-        (windows_version >= WINDOWS_81))
-    {
-        // reset that we had an error
-        // and retry with NtQueryInformationProcess
-        // (for protected processes)
-        PyErr_Clear();
+    if (ret != 0) {
+        if ((GetLastError() == ERROR_ACCESS_DENIED) &&
+            (windows_version >= WINDOWS_81))
+        {
+            // reset that we had an error
+            // and retry with NtQueryInformationProcess
+            // (for protected processes)
+            PyErr_Clear();
 
-        func_ret = psutil_get_cmdline_data(pid, &data, &size);
-        if (func_ret != 0) {
+            func_ret = psutil_get_cmdline_data(pid, &data, &size);
+            if (func_ret != 0) {
+                goto out;
+            }
+        }
+        else {
             goto out;
         }
-    }
-    else {
-        goto out;
     }
    
     // attempt to parse the command line using Win32 API
@@ -986,7 +987,6 @@ psutil_get_cmdline(long pid) {
         PyList_SET_ITEM(py_retlist, i, py_unicode);
         py_unicode = NULL;
     }
-
     ret = py_retlist;
     py_retlist = NULL;
 
--- a/psutil/arch/windows/process_info.c
+++ b/psutil/arch/windows/process_info.c
@@ -162,6 +162,72 @@ const int STATUS_INFO_LENGTH_MISMATCH = 0xC0000004;
 const int STATUS_BUFFER_TOO_SMALL = 0xC0000023L;
 
 
+
+#define WINDOWS_UNITIALIZED 0
+#define WINDOWS_XP 51
+#define WINDOWS_VISTA 60
+#define WINDOWS_7 61
+#define WINDOWS_8 62
+#define WINDOWS_81 63
+#define WINDOWS_10 100
+
+
+int get_windows_version() {
+    OSVERSIONINFO ver_info;
+    BOOL result;
+    DWORD dwMajorVersion;
+    DWORD dwMinorVersion;
+    DWORD dwBuildNumber;
+    static int windows_version = WINDOWS_UNITIALIZED;
+    // didn't get version yet
+    if (windows_version == WINDOWS_UNITIALIZED) {
+        memset(&ver_info, 0, sizeof(ver_info));
+        ver_info.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+        result = GetVersionEx(&ver_info);
+        if (result != FALSE) {
+            dwMajorVersion = ver_info.dwMajorVersion;
+            dwMinorVersion = ver_info.dwMinorVersion;
+            dwBuildNumber = ver_info.dwBuildNumber;
+            // Windows XP, Windows server 2003
+            if (dwMajorVersion == 5 && dwMinorVersion == 1) {
+                windows_version = WINDOWS_XP;
+            }
+            // Windows Vista
+            else if (dwMajorVersion == 6 && dwMinorVersion == 0) {
+                windows_version = WINDOWS_VISTA;
+            }
+            // Windows 7, Windows Server 2008 R2
+            else if (dwMajorVersion == 6 && dwMinorVersion == 1) {
+                windows_version = WINDOWS_7;
+            }
+            // Windows 8, Windows Server 2012
+            else if (dwMajorVersion == 6 && dwMinorVersion == 2) {
+                windows_version = WINDOWS_8;
+            }
+            // Windows 8.1, Windows Server 2012 R2
+            else if (dwMajorVersion == 6 && dwMinorVersion == 3)
+            {
+                windows_version = WINDOWS_81;
+            }
+            // Windows 10, Windows Server 2016
+            else if (dwMajorVersion == 10) {
+                windows_version = WINDOWS_10;
+            }
+        }
+    }
+    return windows_version;
+}
+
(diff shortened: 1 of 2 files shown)


giampaolo:
Beautiful! I will try take a look at this soon.

psutil/arch/windows/process_info.c
```diff
@@ -804,10 +870,78 @@ psutil_get_cmdline(long pid) {
...
+    _NtQueryInformationProcess NtQueryInformationProcess = NULL;
+
+    windows_version = get_windows_version();
+    if (windows_version >= WINDOWS_81) {
+        // by defaut, still use PEB (if command line params have been patched in PEB, we will get the actual ones)
+        if (psutil_get_process_data(pid, KIND_CMDLINE, &data, &size) != 0) {
+            PyErr_Clear(); // reset that we had an error, and retry with NtQueryInformationProcess
+            // if it fails, fallback to NtQueryInformationProcess (for protected processes)
+            NtQueryInformationProcess = get_NtQueryInformationProcess();
+            if (NtQueryInformationProcess != NULL) {
```
giampaolo: In order to stay consistent with the rest of the code I suggest to check for failure instead of success. This way we will avoid long `if` blocks and indentation. Also avoid `else` at all (you save one indented block). In practice: 

```c
if (NtQueryInformationProcess == NULL) {
    PyErr_SetFromWindowsErr(0);
    goto out;
}
...
```

psutil/arch/windows/process_info.c
```diff
@@ -804,10 +870,78 @@ psutil_get_cmdline(long pid) {
...
+
+                cmdline_buffer = calloc(4096, 1);
+                if (cmdline_buffer == NULL) {
+                    PyErr_NoMemory();
+                    goto out;
+                }
+
+                hProcess = psutil_handle_from_pid(pid, PROCESS_QUERY_LIMITED_INFORMATION);
+                if (hProcess != NULL) {
```
giampaolo: Same here (check for failure, not success)

psutil/arch/windows/process_info.c
```diff
@@ -804,10 +870,78 @@ psutil_get_cmdline(long pid) {
...
+                    status = NtQueryInformationProcess(
+                        hProcess,
+                        60, // ProcessCommandLineInformation
+                        cmdline_buffer,
+                        ret_length,
+                        &ret_length
+                    );
+                    CloseHandle(hProcess);
+                    if (NT_SUCCESS(status)) {
```
giampaolo: and here

psutil/arch/windows/security.c
```diff
@@ -160,19 +159,17 @@ psutil_set_privilege(HANDLE hToken, LPCTSTR Privilege, BOOL bEnablePrivilege) {
 int
 psutil_set_se_debug() {
     HANDLE hToken;
-    if (! OpenThreadToken(GetCurrentThread(),
+	if (!OpenProcessToken(GetCurrentProcess(),
```
giampaolo: looks like this was accidental/unnecessary

psutil/arch/windows/security.c
```diff
@@ -160,19 +159,17 @@ psutil_set_privilege(HANDLE hToken, LPCTSTR Privilege, BOOL bEnablePrivilege) {
...
-                          FALSE,
                           &hToken)
        ) {
         if (GetLastError() == ERROR_NO_TOKEN) {
             if (!ImpersonateSelf(SecurityImpersonation)) {
                 CloseHandle(hToken);
                 return 0;
             }
-            if (!OpenThreadToken(GetCurrentThread(),
+			if (!OpenProcessToken(GetCurrentProcess(),
```
giampaolo: looks like this was accidental/unnecessary

psutil/arch/windows/security.c
```diff
@@ -198,17 +195,15 @@ psutil_set_se_debug() {
 int
 psutil_unset_se_debug() {
     HANDLE hToken;
-    if (! OpenThreadToken(GetCurrentThread(),
+	if (!OpenProcessToken(GetCurrentProcess(),
```
giampaolo: looks like this was accidental/unnecessary

psutil/arch/windows/security.c
```diff
@@ -198,17 +195,15 @@ psutil_set_se_debug() {
...
                           TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY,
-                          FALSE,
                           &hToken)
        ) {
         if (GetLastError() == ERROR_NO_TOKEN) {
             if (! ImpersonateSelf(SecurityImpersonation))
                 return 0;
-            if (!OpenThreadToken(GetCurrentThread(),
+			if (!OpenProcessToken(GetCurrentProcess(),
```
giampaolo: looks like this was accidental/unnecessary

giampaolo (changes requested):

psutil/arch/windows/process_info.c
```diff
@@ -804,10 +870,72 @@ psutil_get_cmdline(long pid) {
...
+                60, // ProcessCommandLineInformation
+                cmdline_buffer,
+                ret_length,
+                &ret_length
+            );
+            CloseHandle(hProcess);
+            if (!NT_SUCCESS(status)) {
+                PyErr_SetFromWindowsErr(0);
+                goto out;
+            }
```
giampaolo: OK this is a bit subtle. =) `CloseHandle` may also fail and set errno, so in order to not overwrite the original error code you want to call it *after* you do `PyErr_SetFromWindowsErr(0);`. So in summary I think what you want to do this:

```c
            if (!NT_SUCCESS(status)) {
                PyErr_SetFromWindowsErr(0);
                CloseHandle(hProcess);
                goto out;
            }

            CloseHandle(hProcess);
  ```

File: psutil/arch/windows/process_info.c [hunk 9]
```diff
@@ -162,6 +162,72 @@ const int STATUS_INFO_LENGTH_MISMATCH = 0xC0000004;
...
+            // Windows 8.1, Windows Server 2012 R2
+            else if (dwMajorVersion == 6 && dwMinorVersion == 3)
+            {
+                windows_version = WINDOWS_81;
+            }
+            // Windows 10, Windows Server 2016
+            else if (dwMajorVersion == 10) {
+                windows_version = WINDOWS_10;
+            }
```
giampaolo: Nitpick: how about using a `switch/case/default` block instead? Not terribly important... up to you.

psutil/arch/windows/security.c
```diff
@@ -160,19 +159,17 @@ psutil_set_privilege(HANDLE hToken, LPCTSTR Privilege, BOOL bEnablePrivilege) {
 int
 psutil_set_se_debug() {
     HANDLE hToken;
-    if (! OpenThreadToken(GetCurrentThread(),
+    if (!OpenProcessToken(GetCurrentProcess(),
                           TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY,
-                          FALSE,
```
giampaolo: mmm this looks wrong; aren't you now passing 3 args instead of 4?

psutil/arch/windows/process_info.c
```diff
@@ -804,10 +870,72 @@ psutil_get_cmdline(long pid) {
...
+    char * cmdline_buffer = NULL;
+    WCHAR * cmdline_buffer_wchar = NULL;
+    PUNICODE_STRING tmp = NULL;
+    DWORD string_size;
+    _NtQueryInformationProcess NtQueryInformationProcess = NULL;
+
+    windows_version = get_windows_version();
+    if (windows_version >= WINDOWS_81) {
+        // by defaut, still use PEB (if command line params have been patched in PEB, we will get the actual ones)
+        if (psutil_get_process_data(pid, KIND_CMDLINE, &data, &size) != 0) {
```
giampaolo: Mmmm... I'm a bit confused about this. It appears regardless of the Windows version you first attempt to use the PEB method first, and fallback on using `NtQueryInformationProcess` in case of failure on Win >= 8.1. Wouldn't it make sense to always use `NtQueryInformationProcess` on Win >= 8.1? I a bit against trying 2 different methods and suppressing error in case the first one fails. Unless there's a good reason I think it's better to always use 1 method only.

psutil/arch/windows/process_info.c
```diff
@@ -162,6 +162,72 @@ const int STATUS_INFO_LENGTH_MISMATCH = 0xC0000004;
...
+}
+
+_NtQueryInformationProcess get_NtQueryInformationProcess() {
+    static _NtQueryInformationProcess NtQueryInformationProcess = NULL;
+    if (NtQueryInformationProcess == NULL) {
+        NtQueryInformationProcess = (_NtQueryInformationProcess)GetProcAddress(
+            GetModuleHandleA("ntdll.dll"), "NtQueryInformationProcess");
+    }
+    return NtQueryInformationProcess;
+}
```
giampaolo: I like putting this in a utility function because it can be reused. Would you mind using this utility function also in other parts of the code? Also, for consistency I think it's better to rename this function to `psutil_NtQueryInformationProcess`.

psutil/arch/windows/process_info.c
```diff
@@ -162,6 +162,72 @@ const int STATUS_INFO_LENGTH_MISMATCH = 0xC0000004;
 const int STATUS_BUFFER_TOO_SMALL = 0xC0000023L;
 
 
+
+#define WINDOWS_UNITIALIZED 0
```
giampaolo: Typo? `WINDOWS_UNITIALIZED` -> `WINDOWS_UNINITIALIZED`

giampaolo (changes requested):

psutil/arch/windows/process_info.c
```diff
@@ -162,6 +162,72 @@ const int STATUS_INFO_LENGTH_MISMATCH = 0xC0000004;
...
+
+int get_windows_version() {
+    OSVERSIONINFO ver_info;
+    BOOL result;
+    DWORD dwMajorVersion;
+    DWORD dwMinorVersion;
+    DWORD dwBuildNumber;
+    static int windows_version = WINDOWS_UNITIALIZED;
+    // didn't get version yet
+    if (windows_version == WINDOWS_UNITIALIZED) {
```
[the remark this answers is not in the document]
EccoTheFlintstone: `windows_version` is a static var, this check is taken only the first time the function is called

psutil/arch/windows/security.c
EccoTheFlintstone: I'm using GetProcessToken instead of GetThreadToken and it needs 1 less argument
In the case where the actual privileged call is performed from another thread than the one getting th SE_DEBUG privilege, it might be best to have the whole process enable the privilege

psutil/arch/windows/process_info.c
EccoTheFlintstone: Reading the PEB to get the command line parameters still seem to be the best method if somebody has tampered with the parameters after creating the process.
For instance, create a process as suspended, patch the command line in its PEB and unfreeze it.
The process will use the "new" parameters whereas the system (with `NtQueryInformationProcess`) will give you the "old" ones (see here : https://blog.xpnsec.com/how-to-argue-like-cobalt-strike/)
By default, reading from PEB is still allowed if one has sufficient privileges except for protected processes where one must use `NtQueryInformationProcess`.
So the PEB method will work for almost all processes on windows 8.1+ except the few (4-5) protected processes lying around (where we can fallback on NtQIP), and it will accurately report the actual command line parameters used

psutil/arch/windows/process_info.c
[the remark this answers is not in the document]
giampaolo: Gotcha. I'd add a comment then ('cause I will forget for sure =)).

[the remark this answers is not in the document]
EccoTheFlintstone: done :)

psutil/arch/windows/security.c
giampaolo: I see (didn't notice you changed syscall). I checked ProcessHacker source code and it also uses `OpenProcessToken` so I guess it makes sense to do the same. Good one.

psutil/arch/windows/process_info.c
giampaolo: OK, I see. In this case I would recommend the following:
* use `psutil_handle_from_pid(pid, PROCESS_QUERY_INFORMATION | PROCESS_VM_READ)`
* if it succeeds use the PEB method (and only that)
* if it doesn't (for any reason), use `NtQueryInformationProcess` if Win >= 8.1, else fail
In both cases you should simply close the returned handle: the only reason to get the handle is to see if you have permissions to do so. 
Makes sense?

Also, please add a full explanation of why one method is preferred over the other one (you can include the information you wrote in the comment above).

giampaolo (changes requested):

psutil/arch/windows/process_info.c
EccoTheFlintstone: Actually, I think using `psutil_handle_from_pid(pid, PROCESS_QUERY_INFORMATION | PROCESS_VM_READ)` is redundant as reading from PEB will also open process with those rights, so for every process, we would open/close it twice

EccoTheFlintstone: As it is, the code works. What I could do is add an explanation as why the PEB method is used first for windows >= 8.1 as you propose, but go with this code anyway, no?

giampaolo: Yes but as-is it tries the other method for any error. I would like to limit the check to ERROR_ACCESS_DENIED. If it fails for any other reason I think it's saner to just raise exception.

giampaolo: So something like this (not tested):

```C
    int ret;

    ret = psutil_get_process_data(pid, KIND_CWD, &data, &size);
    if ((ret != 0) && 
            (GetLastError() == ERROR_ACCESS_DENIED) && 
            (windows_version >= WINDOWS_81)) 
    {
        // try NtQueryInformationProcess()
        PyErr_Clear();
        ...
    }
    else {
        goto out;
    }
```

giampaolo: ...also, I suggest to put the whole `NtQueryInformationProcess()` in a separate function, similarly to what it has been done for `psutil_get_process_data`

giampaolo: hint: function name could be `psutil_get_cmdline_data`

psutil/arch/windows/process_info.c
```diff
@@ -791,10 +855,70 @@ static int psutil_get_process_data(long pid,
...
+        CloseHandle(hProcess);
+        return -1;
+    }
+    CloseHandle(hProcess);
+    tmp = (PUNICODE_STRING)cmdline_buffer;
+    string_size = wcslen(tmp->Buffer) + 1;
+    cmdline_buffer_wchar = (WCHAR *)calloc(string_size, sizeof(WCHAR));
+
+    if (cmdline_buffer_wchar == NULL) {
+        free(cmdline_buffer);
```
giampaolo: ...and here (use goto instead)

giampaolo (changes requested):
hopefully the last change

giampaolo:
Merged. Thanks for the hard work. ;)

EccoTheFlintstone:
Hi,

thanks for the merge

There's a slight problem with your latest commit in psutil_get_cmdline_data:

return 0;


error:

if (cmdline_buffer != NULL)

free(cmdline_buffer);

if (hProcess != NULL)

CloseHandle(hProcess);
return -1;

the free() and CloseHandle() are always necessary
Before, ret was set by default to -1 (error) and was set to 0 (success) only at the last moment to distinguish between failure and success
All the cleanup was performed at the end for clarity

as it is now, hProcess is not closed and cmdline_buffer is not freed (but it should be)

-- 
 Securely sent with Tutanota. Get your own encrypted, ad-free mailbox: 
 https://tutanota.com


Feb 3, 2019, 7:57 AM by notifications@github.com:

Issue #1394, opened 2019-01-24
WindowsError: [Error 0] The operation completed successfully for exe() after upgrading to 5.5.0


jablonskim:
Tested with Python 2.7.13 and 3.7.1, Windows 10 1809 (10.0.17763.0).
psutil version: 5.5.0

It fails on 'Secure System' process, only when **Python was run as an Administrator**:

```
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> [p for p in psutil.process_iter()]
[psutil.Process(pid=0, name='System Idle Process', started='2018-11-06 04:18:12'), psutil.Process(pid=4, name='System', started='2018-11-06 04:18:12'), Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\psutil\__init__.py", line 395, in __str__
    info["name"] = self.name()
  File "C:\Python27\lib\site-packages\psutil\__init__.py", line 609, in name
    name = self._proc.name()
  File "C:\Python27\lib\site-packages\psutil\_pswindows.py", line 635, in wrapper
    return fun(self, *args, **kwargs)
  File "C:\Python27\lib\site-packages\psutil\_pswindows.py", line 687, in name
    return py2_strencode(os.path.basename(self.exe()))
  File "C:\Python27\lib\site-packages\psutil\_pswindows.py", line 635, in wrapper
    return fun(self, *args, **kwargs)
  File "C:\Python27\lib\site-packages\psutil\_pswindows.py", line 701, in exe
    return py2_strencode(convert_dos_path(cext.proc_exe(self.pid)))
WindowsError: [Error 0] The operation completed successfully
```

giampaolo:
Mmm not good, and I can't reproduce it. Do you have VS? Could you re-compile psutil from sources and see where exactly the error originates from?

jablonskim:
Yes, sure. I will try to provide more details till end of the day or next Monday.

About reproducing it: just running it from Administrator account is not enough. Try starting cmd or powershell using right click and 'Run as an Administrator' and then run Python from it. Without running as an Administrator it runs perfectly fine.

jablonskim:
[...]

PR #1413, opened 2019-02-17
#1394 / win / exe: use QueryFullProcessImageNameW to get the exe()

giampaolo:
See #1394.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,5 +1,16 @@
 *Bug tracker at https://github.com/giampaolo/psutil/issues*
 
+5.5.2
+=====
+
+XXXX-XX-XX
+
+**Bug fixes**
+
+- 1394_: [Windows] Process name() and exe() may erronously return "Registry".
+  QueryFullProcessImageNameW is now used instead of GetProcessImageFileNameW
+  in order to prevent that.
+
 5.5.1
 =====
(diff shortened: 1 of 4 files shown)


giampaolo:
As said in https://github.com/giampaolo/psutil/issues/1394#issuecomment-464476034 this has the drawback of breaking backward compatibility on Win XP. I'm not sure how much time I'm willing to invest into this but roughly I see one solution:

1) use C pre-processor in order to identify the Windows version and choose the right code path:

```c
#if (_WIN32_WINNT >= 0x0600) // Windows Vista and above
   // new code using QueryFullProcessImageNameW
#else // Windows XP
    // old code using GetProcessImageFileNameW
#endif

2) implement a similar routine in the python code so that if WINVER < VISTA we use `QueryDosDevice` to convert the returned path in the right format

giampaolo:
@ogurets if I land a patch including Win XP support could you please try (I have no Win XP box to test against)?

ogurets:
@giampaolo Sure! Cannot promise I'll be swift though, I'm pretty piled up right now :(

giampaolo:
No problem. In general terms this is low-priority anyway (I was tempted to remove Win XP support, but sort of felt a bit cautious... =)).

PR #1418, opened 2019-02-20
Windows / speeup: load libraries only on startup only

giampaolo:


giampaolo:
I restarted from scratch with this and created https://github.com/giampaolo/psutil/pull/1422. Closing this out.

PR #1436, opened 2019-02-26
OSX memory_maps() - fix segfault

### LINKED ISSUE #1291: OSX: memory_maps() segfaults or raise EINVAL
I get Exception: OSError(22, 'Invalid argument') on OSX 10.13.4 with Python 2.7

Adding this to the loop makes it go away:

```
        if proc.pid == os.getpid(): continue   
```

giampaolo:
This is an attempt to fix #1291.


wiggin15:
I think this PR is good even if it doesn't fix the problem in #1291. Maybe we can merge it but keep the original ticket open.

PR #1444, opened 2019-02-28
IBM i support (take 2)

ThePrez:


ThePrez:
@giampaolo, you can see in these commits how we can easily get some info via SQL on IBM i. However, it introduces a dependency on the `ibm_db` module. I don't think it's the right answer to have psutil explicitly require it, but are you ok with an init-time check and an exception saying something like `you must have the ibm_db package installed`?

giampaolo:
You basically duplicated all the logic in 2 places. Don't do that. Your other PR is the right way to go. See the comment I added in there.

PR #1446, opened 2019-02-28
Restore Win-7 support on GIT master (5.5.1 was OK)

giampaolo:
Given the recent concerns re. older Windows versions I installed Win 7. It turns out we broke Win 7 support but only on GIT master (psutil 5.5.1 works fine, which is good). This PR:
- cmdline: avoid using NtQueryInformationProcess on win < 8.1 (raises OSError)
- stop using GetNativeSystemInfo, use GetSystemInfo instead (broke cpu_cunt())
- allocate memory after getting the right size as suggested by @EccoTheFlintstone here: https://github.com/giampaolo/psutil/commit/59e3c5e2aa889d443f5f0e44beb52f654fc6e23e#commitcomment-32526230 (fix segfault)

### DIFF
--- a/DEVGUIDE.rst
+++ b/DEVGUIDE.rst
@@ -82,7 +82,7 @@ On Windows:
 
 .. code-block:: bat
 
-    set TSCRIPT=foo.py && make test
+    make test foo.py
 
 Adding a new feature
 ====================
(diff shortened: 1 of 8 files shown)


giampaolo:
@EccoTheFlintstone could you review this?

EccoTheFlintstone:
looks OK to me

I tested GIT master (5.6.0), before this PR,  on a  win 7 x64 and I don't have problems with `cpu_count() `:


```
>>> psutil.__version__
'5.6.0'
>>> psutil.cpu_count()
2
>>> import win32api
>>> win32api.GetNativeSystemInfo()[5]
2
```

giampaolo:
Thanks.

Issue #875, opened 2016-08-25
Process.environ() on Windows may raise WindowsError


giampaolo:
https://ci.appveyor.com/project/giampaolo/psutil/build/806/job/fpxxbai3sg291i75

```
======================================================================
ERROR: test_proc_environ (test_process.TestNonUnicode)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\psutil\psutil\tests\test_process.py", line 2054, in test_proc_environ
    encode_path(p.environ()['FUNNY_ARG']), funny_path)
  File "c:\projects\psutil\psutil\__init__.py", line 772, in environ
    return self._proc.environ()
  File "c:\projects\psutil\psutil\_pswindows.py", line 536, in wrapper
    return fun(self, *args, **kwargs)
  File "c:\projects\psutil\psutil\_pswindows.py", line 597, in environ
    return parse_environ_block(cext.proc_environ(self.pid))
WindowsError: [Error 299] Only part of a ReadProcessMemory or WriteProcessMemory request was comp
```

giampaolo:
@fbenkstein do you mind taking a look at this or do you have any suggestion?
The error occurs here:
https://github.com/giampaolo/psutil/blob/7c36a6341f5b13c4d8bd5afcb5aeca0f7278ed1d/psutil/arch/windows/process_info.c#L663

fbenkstein:
@giampaolo how often does this occur?  I assume it isn't reproducible reliably.  Unfortunately I don't have a Windows dev environment at the moment so I can't debug this myself.  Maybe it's possible that we're trying to read the process memory when it isn't fully created yet or otherwise the test machine was so busy that the test process exited while we were trying to read its memory.

giampaolo:
I can reproduce it if I run that test in a loop. After a while (say 20-30 times) I bump into the exception. I suppose in that case we could realloc the buffer with a bigger size and retry. I will try that.

fbenkstein:
I don't think it's the buffer size.  If it's too small you'll simply read less.  IIRC increasing the buffer will probably increase the likelihood of the error since you might try to read memory beyond what is actually valid in the process.  If you want to handle the error, it's probably better to check if the process is still alive, though I think this should be left up to the caller of environ().

giampaolo:
In that case (unit test) the process is alive for sure.

giampaolo:
[...]

PR #1449, opened 2019-03-06
Retry WindowsError ERROR_PARTIAL_COPY

gozdal:
Workaround for https://github.com/giampaolo/psutil/issues/875


giampaolo:
Mmm no, this should be handled in C, and without using sleep().

PR #1462, opened 2019-03-13
Make tests invariant to LANG setting

bdrung:
When `LANG` is set to a non English locale (e.g. `de_DE.UTF-8`), `free` and `vmstat` will output the text translated and cause the test cases to fail.

Therefore set `LANG=C.UTF-8` when calling `free` or `vmstat` to always get the English output.


giampaolo:
Nice on, thanks.

Issue #1460, opened 2019-03-13
[Linux] Only first battery reported with psutil.sensors_battery()


VorpalBlade:
**Linux**
* Ubuntu 16.04 x86_64
* psutil version 5.6.1

**Bug description**

Only linux the first batter is reported. Some laptops (such as Lenovo Thinkpad T480) has two batteries, one built in, and one removable. The information about the removable battery is not reported
Example:
```
In [7]: psutil.sensors_battery()                                                                                                                                                                                                     
Out[7]: sbattery(percent=92.84843559528647, secsleft=<BatteryTime.POWER_TIME_UNLIMITED: -2>, power_plugged=True)
```

But two batteries (and AC) are available in the sys filesystem:
```
$ ls /sys/class/power_supply/
AC  BAT0  BAT1
```

giampaolo:
Yeah, that's by design (at the moment). We may want to introduce a `all=False` parameter or something. When True it should return a list of namedtuples.

amanusk:
@giampaolo I can try and take a look at this

giampaolo:
Go for it Alex.

mohkale:
I'm using psutil to create components for my tmux status-line and it seems to be picking up the baterry level for my bluetooth keyboard by default :joy:, any idea if #1466 might be merged soon?

PR #1466, opened 2019-03-21
Support multiple batteries 

amanusk:
This is a WIP PR.
The purpose is to discuss possible solutions to issue #1460 

The change shouldn't be too difficult (at least in Linux it isn't), but might require some API adjustments.
Here I added a `per_battery` to make minimal changes to the API.
I added a 'label' to the tuple to the battery tuple. (This will require adjustment for the code of other platforms but '' can be used by default)

I wanted to first discuss the way to go before moving forward with this. 

Example of adjusted battery script printout:
```
Label:     BAT0
charge:     79.48%
status:     charging
plugged in: yes
Label:     BAT1
charge:     65.09%
status:     charging
plugged in: yes
```

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -2370,7 +2370,7 @@ if hasattr(_psplatform, "sensors_fans"):
 # Linux, Windows, FreeBSD, macOS
 if hasattr(_psplatform, "sensors_battery"):
 
-    def sensors_battery():
+    def sensors_battery(per_battery=False):
         """Return battery information. If no battery is installed
         returns None.
 
@@ -2380,7 +2380,7 @@ if hasattr(_psplatform, "sensors_battery"):
                      POWER_TIME_UNLIMITED or POWER_TIME_UNLIMITED.
          - power_plugged: True if the AC power cable is connected.
         """
-        return _psplatform.sensors_battery()
+        return _psplatform.sensors_battery(per_battery)
 
     __all__.append("sensors_battery")
 
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -197,7 +197,8 @@ scpufreq = namedtuple('scpufreq', ['current', 'min', 'max'])
 shwtemp = namedtuple(
     'shwtemp', ['label', 'current', 'high', 'critical'])
 # psutil.sensors_battery()
-sbattery = namedtuple('sbattery', ['percent', 'secsleft', 'power_plugged'])
+sbattery = namedtuple(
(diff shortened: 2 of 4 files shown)


amanusk:
@giampaolo please advise which direction (if any) do you want for this:
Returning a tuple with label as suggested in this PR, or a dictionary or something else.

giampaolo:
@amanusk hey Alex, sorry for being late. I think it's not worth it to add a `label` to the namedtuple. 
Also (unrelated with this change) I noticed this:

```python
    bats = [x for x in os.listdir(POWER_SUPPLY_PATH) if x.startswith('BAT')]
```

Looks like this deserves to be `sorted()` because by default we should return the first battery ("BAT0").
Before proceeding with this I wanted to investigate if it's possible to do the same on other platforms.

amanusk:
> I think it's not worth it to add a label to the namedtuple

So the alternative would be to return a dictionary or list instead? 
 
> I wanted to investigate if it's possible to do the same on other platforms.

I can try testing this on FreeBSD, not sure if MacOS has laptops with 2 batteries

P.S
I did notice the test for `psutil.sensors_battery()` fails on my system if it is currently using `BAT1` because it always compares the result to `BAT0` 

P.P.S
Seems like the T*80 Thinkpad series were the last generation of Thinkpads with dual batteries :(

giampaolo:
> >  I think it's not worth it to add a label to the namedtuple

> So the alternative would be to return a dictionary or list instead?

I would just return a list of ntuples. "BAT{N}" is not really a label so it's not worth returning it. This may change if a label/name can be extracted on other systems but I doubt that (and (extra) I don't think it's worth it to return detailed battery info such as model, etc. - the main goal of this function is to detect battery charge). 

> I did notice the test for psutil.sensors_battery() fails on my system if it is currently using BAT1 because it always compares the result to BAT0

Not sure I understand this one. If it's a separate issue perhaps it'd be better to fix it in another PR.

amanusk:
> Not sure I understand this one. If it's a separate issue perhaps it'd be better to fix it in another PR.

It is related. Here is an example:
```
FAIL: psutil.tests.test_linux.TestSensorsBattery.test_percent
  File "/home/amanusk/Code/psutil/psutil/tests/test_linux.py", line 1388, in test_percent
    self.assertAlmostEqual(acpi_value, psutil_value, delta=1)
AssertionError: 92 != 100.0 within 1 delta
```

The laptop is currently discharging BAT1, while BAT0 stays at 100%

Issue #1307, opened 2018-07-23
PROCFS_PATH for disk_partitions not functioning as expected.


srivignessh:
The disk_partitions function in _pslinux.py uses the C executable _psutil_linux.so.

The PROCFS_PATH parameter is unused in C executable and hence I still get the default disk_partitions for the path '/proc' instead of '/host/proc'.

```
from . import _psutil_linux as cext
def disk_partitions(all=False):
    """Return mounted disk partitions as a list of namedtuples."""
    .....
    .....
    retlist = []
    partitions = cext.disk_partitions()
    for partition in partitions:
        device, mountpoint, fstype, opts = partition
        .....
        ntuple = _common.sdiskpart(device, mountpoint, fstype, opts)
        retlist.append(ntuple)
    return retlist 
```

srivignessh:
The '/etc/mtab' is a symbolic link to '/proc/mounts' so I get receive the metrics from '/proc' instead of modified PROCFS_PATH='/host/proc'.

giampaolo:
Fixed.

srivignessh:
Thanks

Create-python:
/etc/matb 里面的字段都是什么含义？

Create-python:
The fields in /etc/matb is what means？

PR #1470, opened 2019-03-28
Fix corner case when /etc/mtab doesn't exist and procfs=/proc

clamoriniere:
In some Linux configurations the `/etc/mtab` does not exist but the
procfs_path is equal to `/proc`. 
With the fix done for the issue #1307, the described configuration didn't work.

This Commit introduce an additional check that verifies if the
`/etc/mtab` file exists before using it, else it defaults to `<procfs_path>/self/mounts`


clamoriniere:
@giampaolo thank you for your fast action. Any plan to create a new release in the coming days?

giampaolo:
No plans yet.

Issue #1472, opened 2019-04-02
[Linux] Differences between psutil.cpu_freq(True) and psutil.cpu_count()


amanusk:
**Rasbperry pi 3**
* Raspbian 
* latest master branch b23fb29

**Bug description**
Wonder about behaviour:

`psutil.cpu_freq(percpu=True)`, on a Raspberry pi 3 now returns:
`[scpufreq(current=600.0, min=600.0, max=1200.0)]` i.e. only 1 CPU

When both `psutil.cpu_count(logical=False)` and `psutil.cpu_count(logical=True)` return `4`: The expected behaviour. 

Looking at the code, it seems this is because on the Pi, 
there is only `/sys/devices/system/cpu/cpufreq/policy0` while there are 
```
/sys/devices/system/cpu/cpu0
/sys/devices/system/cpu/cpu1
/sys/devices/system/cpu/cpu2
/sys/devices/system/cpu/cpu3
```

And the former is checked first.
Should this be compared to `psutil.cpu_count(logical=False)`, and the more accurate result taken?

Should I open a PR for this?

giampaolo:
Mh... annoying. Looking back at how this code is implemented, it looks like we may want to try something different. Instead of `glob` we could use the CPU numbers directly. Something like not tested:

```python
for x in range(psutil.cpu_count(logical=True)):
    if os.path.exists("/sys/devices/system/cpu/cpufreq/policy%s" % x):
        ...
    elif os.path.exists("/sys/devices/system/cpu/cpu%s/cpufreq" % x):
        ...
    else:
        break
```

It seems to me this will be more robust.

giampaolo:
...or possibly this:
```python

def cpu_freq():
    def get_path(num):
        for p in ("/sys/devices/system/cpu/cpufreq/policy%s" % num,
                  "/sys/devices/system/cpu/cpu%s/cpufreq" % num):
            if os.path.exists(p):
                return p

    for x in range(psutil.cpu_count(logical=True)):
        path = get_path(x)
        if not path:
            break
        ...
```

PR #1481, opened 2019-04-05
Cpu freq fix

amanusk:
This PR addresses issue #1472.

`cpu_freq()` is updated to to check whether `/sys/devices/system/cpu/cpufreq/policy*` or `/sys/devices/system/cpu/cpu*/cpufreq` exists.

Tests updated to test for both cases and for multiple CPUs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ syntax: glob
 *.rej
 *.so
 *.swp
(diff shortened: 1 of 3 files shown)


giampaolo:
Great work as usual Alex. Thanks.

Issue #1402, opened 2019-02-03
leaking internal exeptions


arekm:
The problem is that with psutil (5.4.2) I'm getting some, internal, I guess, exceptions which shouldn't be visible externally, right? For example I'm getting psutil._exceptions.NoSuchProcess while documented is psutil.NoSuchProcess

````
Traceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/psutil/_pslinux.py", line 1390, in wrapper
    return fun(self, *args, **kwargs)
  File "/usr/lib64/python3.6/site-packages/psutil/_pslinux.py", line 1965, in uids
    data = self._read_status_file()
  File "/usr/lib64/python3.6/site-packages/psutil/_common.py", line 337, in wrapper
    return fun(self)
  File "/usr/lib64/python3.6/site-packages/psutil/_pslinux.py", line 1445, in _read_status_file
    with open_binary("%s/%s/status" % (self._procfs_path, self.pid)) as f:
  File "/usr/lib64/python3.6/site-packages/psutil/_pslinux.py", line 187, in open_binary
    return open(fname, "rb", **kwargs)
FileNotFoundError: [Errno 2] No such file or directory: '/proc/27862/status'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "myscript.py", line 72, in <module>
    process_pureftpd(pureftpd_pid)
  File "myscript.py", line 55, in process_pureftpd
    r_uid, e_uid, s_uid = p.uids()
  File "/usr/lib64/python3.6/site-packages/psutil/_common.py", line 337, in wrapper
    return fun(self)
  File "/usr/lib64/python3.6/site-packages/psutil/__init__.py", line 722, in uids
    return self._proc.uids()
  File "/usr/lib64/python3.6/site-packages/psutil/_pslinux.py", line 1401, in wrapper
    raise NoSuchProcess(self.pid, self._name)
psutil._exceptions.NoSuchProcess: psutil.NoSuchProcess process no longer exists (pid=27862)
````

giampaolo:
Yes, the `__repr__` is incorrect, but catching `psutil.NoSuchProcess` works, so it's a cosmetic fix.

giampaolo:
Change was introduced in https://github.com/giampaolo/psutil/commit/fe68b30dacec3255b023fcefac5c9095d96a692f.
[...]

PR #1342, opened 2018-09-21
Some refactoring of string encoding/decoding utilities

embray:
This is a bit of refactoring pulled out from my Cygwin branch #998.

I think these new string utilities functions are very clear and logical in their respective purposes.  I also took a close look at the current state of the _psutil_windows.c module, and I believe it is quite consistent about returning unicode strings, so the several calls to `unicode2str` should be safe.

I also moved `open_text` and `open_binary` from the linux module to the common module.  They're still only used in the linux module, but my experience from #998 shows that there should be use for them elsewhere as well.

PR #1483, opened 2019-04-08
Initial Prototype for Cygwin support module (take 2)

embray:
This is an easier to digest alternative to #998 .

Rather than try to implement the kitchen sink, this provides a barely-more-than bare-minimum implementation for a Cygwin support module.  Most top-level functions and `Process` methods either work, at least with the default arguments, or raise a `NotImplementedError` for now.

You can get a sense for what works and what doesn't work by looking at what tests are skipped.

The major bit of refactoring needed for this to work is the introduction of a `psutil._pslinux_common` module.  It implements a `Process` base class and several functions that work on Linux and Cygwin, but mostly avoids very specialized Linux-specific functionality.  In principle this could also be useful for implementing support on other platforms (if they exist?) that aim for partial Linux-compatibility without being fully Linux.

I think some version of this would be good to start with and is much better than nothing (it still does quite a bit!) and I will reintroduce additional features a bit at a time, making any additional prerequisite refactoring in the process.

TODO:
* [ ] Add CI on AppVeyor
* [ ] Add some mention of prototype support for Cygwin; for now I don't want to document in detail what does or does not work on Cygwin, but just mention that prototype support exists and is incomplete.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -16,6 +16,7 @@ sensors) in Python. Supported platforms:
  - NetBSD
  - Sun Solaris
  - AIX
+ - Cygwin (experimental)
 
 Works with Python versions from 2.6 to 3.4+.
 """
@@ -77,6 +78,7 @@ from ._common import NIC_DUPLEX_UNKNOWN
 
 from ._common import AIX
 from ._common import BSD
+from ._common import CYGWIN
 from ._common import FREEBSD  # NOQA
 from ._common import LINUX
 from ._common import MACOS
@@ -173,6 +175,10 @@ elif AIX:
     # via sys.modules.
     PROCFS_PATH = "/proc"
 
+elif CYGWIN:
+    PROCFS_PATH = "/proc"
+    from . import _pscygwin as _psplatform
+
 else:  # pragma: no cover
     raise NotImplementedError('platform %s is not supported' % sys.platform)
 
@@ -201,7 +207,7 @@ __all__ = [
     "POWER_TIME_UNKNOWN", "POWER_TIME_UNLIMITED",
 
     "BSD", "FREEBSD", "LINUX", "NETBSD", "OPENBSD", "MACOS", "OSX", "POSIX",
-    "SUNOS", "WINDOWS", "AIX",
+    "SUNOS", "WINDOWS", "AIX", "CYGWIN",
 
     # classes
     "Process", "Popen",
@@ -357,6 +363,12 @@ _psplatform.TimeoutExpired = TimeoutExpired
 if POSIX:
     from . import _psposix
     _psposix.TimeoutExpired = TimeoutExpired
+if CYGWIN:
+    from . import _pslinux
+    _pslinux.NoSuchProcess = NoSuchProcess
+    _pslinux.ZombieProcess = ZombieProcess
+    _pslinux.AccessDenied = AccessDenied
+    _pslinux.TimeoutExpired = TimeoutExpired
 
 
 # =====================================================================
@@ -657,11 +669,16 @@ class Process(object):
         if self.pid == lowest_pid:
             return None
         ppid = self.ppid()
+        if CYGWIN:
+            # See note for Process.create_time in _pscygwin.py
+            rounding_error = 1
+        else:
+            rounding_error = 0
         if ppid is not None:
             ctime = self.create_time()
             try:
                 parent = Process(ppid)
-                if parent.create_time() <= ctime:
+                if parent.create_time() <= (ctime + rounding_error):
                     return parent
                 # ...else ppid has been reused by another process
             except NoSuchProcess:
@@ -1490,7 +1507,15 @@ def pids():
     """Return a list of current running PIDs."""
     global _LOWEST_PID
     ret = sorted(_psplatform.pids())
-    _LOWEST_PID = ret[0]
+
+    if CYGWIN:
+        # Note: The _LOWEST_PID concept is not very meaningful since every
+        # "top-level" process has a non-1 PID; they do have PPID of 1 but that
+        # does not correspend to an actual process that shows up in the pids()
+        # list
+        _LOWEST_PID = 1
+    else:
+        _LOWEST_PID = ret[0]
     return ret
 
 
@@ -2462,7 +2487,7 @@ def test():  # pragma: no cover
         else:
             cputime = ''
 
-        user = p.info['username']
+        user = p.info['username'] or ''
         if not user and POSIX:
             try:
                 user = p.uids()[0]
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -36,6 +36,7 @@ from socket import SOCK_DGRAM
 from socket import SOCK_STREAM
 
 import psutil
+from psutil import CYGWIN
 from psutil import MACOS
 from psutil import POSIX
 from psutil import SUNOS
@@ -942,6 +943,11 @@ def tcp_socketpair(family, addr=("", 0)):
             raise
 
 
+# NOTE: Cygwin has a known bug
+# (https://cygwin.com/ml/cygwin/2017-01/msg00111.html) with creating a pair
+# of UNIX sockets within the same process and having them communicate with
+# each other.  Therefore any test which uses this should be skipped on Cygwin
+# until there is a better solution.
 def unix_socketpair(name):
     """Build a pair of UNIX sockets connected to each other through
     the same UNIX file name.
@@ -976,7 +982,7 @@ def create_sockets():
         if supports_ipv6():
             socks.append(bind_socket(socket.AF_INET6, socket.SOCK_STREAM))
             socks.append(bind_socket(socket.AF_INET6, socket.SOCK_DGRAM))
-        if POSIX and HAS_CONNECTIONS_UNIX:
+        if POSIX and HAS_CONNECTIONS_UNIX and not CYGWIN:
             fname1 = unix_socket_path().__enter__()
             fname2 = unix_socket_path().__enter__()
             s1, s2 = unix_socketpair(fname1)
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ syntax: glob
 *.so
 *.swp
 .failed-tests.txt
+*.dll
 .cache/
 .idea/
 .tox/
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -44,7 +44,7 @@ PY3 = sys.version_info[0] == 3
 __all__ = [
     # constants
     'FREEBSD', 'BSD', 'LINUX', 'NETBSD', 'OPENBSD', 'MACOS', 'OSX', 'POSIX',
-    'SUNOS', 'WINDOWS',
+    'SUNOS', 'WINDOWS', 'CYGWIN',
     'ENCODING', 'ENCODING_ERRS', 'AF_INET6',
     # connection constants
     'CONN_CLOSE', 'CONN_CLOSE_WAIT', 'CONN_CLOSING', 'CONN_ESTABLISHED',
@@ -86,6 +86,7 @@ NETBSD = sys.platform.startswith("netbsd")
 BSD = FREEBSD or OPENBSD or NETBSD
 SUNOS = sys.platform.startswith(("sunos", "solaris"))
 AIX = sys.platform.startswith("aix")
+CYGWIN = sys.platform.startswith("cygwin")
 
 
 # ===================================================================
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -38,6 +38,7 @@ else:
     enum = None
 
 # can't take it from _common.py as this script is imported by setup.py
+PY2 = sys.version_info[0] == 2
 PY3 = sys.version_info[0] == 3
 
 __all__ = [
@@ -61,6 +62,7 @@ __all__ = [
     'pthread', 'puids', 'sconn', 'scpustats', 'sdiskio', 'sdiskpart',
     'sdiskusage', 'snetio', 'snicaddr', 'snicstats', 'sswap', 'suser',
     # utility functions
+    'str2bytes', 'bytes2str', 'unicode2str',
     'conn_tmap', 'deprecated_method', 'isfile_strict', 'memoize',
     'parse_environ_block', 'path_exists_strict', 'usage_percent',
     'supports_ipv6', 'sockfam_to_enum', 'socktype_to_enum', "wrap_numbers",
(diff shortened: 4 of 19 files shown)


psutil/tests/__init__.py
```diff
@@ -904,6 +905,29 @@ def bind_socket(family=AF_INET, type=SOCK_STREAM, addr=None):
         raise
 
 
+if CYGWIN:
```
embray: I didn't mean to commit this bit yet.  Going to remove it and rebase.

giampaolo:
Good job @embray. I added some initial comments. On top of that, I still don't like the separate _linux_common.py file. Please keep the main logic in _pslinux.py and do this instead:
* from _pscygwin.py `import _pslinux`
* in `_pslinux.py` exclude **only** the Linux-specific code which causes error on import by using:

```python
# _pslinux.py

if not CYGWIN:
    ...
```
- in `_pscygwin.py` do this depending on whether the implementation is the same or not. Example:

```python
# _pscygwin.py

pids = _pslinux.pids

def virtual_memory():
    ...
```
- in the `Process` class just override the methods that differ:

```python
# _pscygwin.py

class Process(_pslinux.Process):
    
    @memoize_when_activated
    def _parse_stat_file(self):
         ...
```

Also, generally speaking instead of merging a partial implementation I'm aiming at implementing everything, unless it's really difficult or not possible (in which case we'll document it). For the functionality which need to use MSDN APIs in C we can figure out how to refactor _psutil_windows.c. I have no idea how things are on that front yet but it seems we have two choices:
- use #ifdef CYGWIN to include / exclude things which do not compile
- extract only the things we need and put them in `psutil/arch/windows/cygwin_common.c` file

...but let's get to that later.

Also I noticed something interesting. `ps aux` on cygwin shows both a PID and a WINPID column. I don't know how it does that but it makes me think that it would be nice to have a brand new, cygwin-only `Process.winpid` method (this can also be done later after this is merged).

I'd have further comments, like trying to reuse tests from `test_linux.py` but those can be adjusted later (and I can help you with that).

Extra: I notices something weird. While playing with this on Windows 10 + VirtualBox the VM suddenly crashed a couple of times. Scary... =)

embray:
> On top of that, I still don't like the separate _linux_common.py file. Please keep the main logic in _pslinux.py and do this instead

I strongly disagree.  You're welcome to try to doing it that way yourself but what you're going to end up with is code riddled with `if not CYGWIN:`, as well as any time new code is added blithely to `_pslinux.py` without checking whether it works on Cygwin it is likely to break. I say this based on significant experience with dealing with code just like this.  Even with CI it's not always possible to catch cases where this could break.

In fact, redoing it that way would require so much additional work that I probably won't bother; maybe I'll just fork instead.

embray:
> Also, generally speaking instead of merging a partial implementation I'm aiming at implementing everything, unless it's really difficult or not possible (in which case we'll document it).

That's going to require significant refactoring, which I do intend to do, but it's going to have to be done separately.  The complexity will grow significantly à la #998, it will become difficult to keep up with changes in master, and just generally become an unmaintainable mess again.  Better to start with something small, note in the documentation (as I have started to do) that it is partial and/or experimental, and then add on functionality as able.  All the core functionality is already there.

embray:
*sigh* @giampaolo I owe you an apology.  I went ahead and tried it your way just to see how bad it actually was, and this is all it took to at least make `_pslinux.py` *importable* on Cygwin.  And all that matters as far as reusing bits of it, I think, is that it can be imported.

I tried to keep the changes from even being Cygwin-specific and this is all it took:

```diff
diff --git a/psutil/_pslinux.py b/psutil/_pslinux.py
index a1eaa3cd..f303ed53 100644
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -23,7 +23,12 @@ from collections import namedtuple
 from . import _common
 from . import _psposix
 from . import _pslinux_common
-from . import _psutil_linux as cext
+
+try:
+    from . import _psutil_linux as cext
+except ImportError:
+    cext = object()
+
 from . import _psutil_posix as cext_posix
 from ._common import isfile_strict
 from ._common import memoize
@@ -102,11 +107,16 @@ LITTLE_ENDIAN = sys.byteorder == 'little'
 # * https://lkml.org/lkml/2015/8/17/234
 DISK_SECTOR_SIZE = 512

+if hasattr(socket, 'AF_PACKET'):
+    AF_PACKET = socket.AF_PACKET
+else:
+    AF_PACKET = -1
+
 if enum is None:
-    AF_LINK = socket.AF_PACKET
+    AF_LINK = AF_PACKET
 else:
     AddressFamily = enum.IntEnum('AddressFamily',
-                                 {'AF_LINK': int(socket.AF_PACKET)})
+                                 {'AF_LINK': int(AF_PACKET)})
     AF_LINK = AddressFamily.AF_LINK

 # ioprio_* constants http://linux.die.net/man/2/ioprio_get
```

On principle I still don't like it, per my comment that Cygwin as not Linux.  But we can try things your way for now, given that at least practically speaking it took very little to at least make the `_pslinux` module importable.

embray:
How's this as a middle ground?  I did away with the `_pslinux_common` module and instead have `_pscygwin` import `_pslinux` as you prefer.

However, it does not make `_pscygwin.Process` a subclass of `_pslinux.Process`.  So doing would have a few problems: It gives the false impression that Cygwin is somehow a superset of Linux in terms of functionality, which is false.  If anything, by that logic, it should be the other way around, with Linux inheriting from Cygwin, though that would come with its own problems.  This is why I wanted a module of common functionality between them--they are in a sense divergent common ancestors of, say, an older version of Linux.  As a technical problem, having `_pscygwin.Process` inherit `_pslinux.Process` would also mean it would inherit methods which just don't make sense on Cygwin (such as `_read_smaps_file`).

Instead, `_pscgwin.Process` stands on its own, but borrows judiciously from `_pslinux.Process` where it makes sense to.  This appears to work quite well, modulo fine details.

psutil/__init__.py
```diff
@@ -357,6 +363,12 @@ def __init__(self, seconds, pid=None, name=None):
 if POSIX:
     from . import _psposix
     _psposix.TimeoutExpired = TimeoutExpired
+if CYGWIN:
+    from . import _pslinux
+    _pslinux.NoSuchProcess = NoSuchProcess
+    _pslinux.ZombieProcess = ZombieProcess
+    _pslinux.AccessDenied = AccessDenied
+    _pslinux.TimeoutExpired = TimeoutExpired
```
embray: I don't like this.  Didn't there used to be a `psutil._exceptions` module?  Why did that go away?  Having something like that would seem a bit less crazy like this...

giampaolo:
Look @embray, I'm not willing to debate over design decisions anymore. I've basically been saying the same thing over the course of 2 years already:

* https://github.com/giampaolo/psutil/pull/998#issuecomment-288773839
* https://github.com/giampaolo/psutil/pull/998#issuecomment-468743187
* https://github.com/giampaolo/psutil/pull/998#issuecomment-468776098
* https://github.com/giampaolo/psutil/pull/1342#issuecomment-468682739
* https://github.com/giampaolo/psutil/pull/1483#issuecomment-480964630

There must be some communication issue at play here because the same goes for the str/unicode/bytes utilities which I [rejected](https://github.com/giampaolo/psutil/pull/1342#issuecomment-443299003) in the other PR some time ago already, but that you decided to include into this one anyway (they could have made some sense if `_pswindows.py` was involved in this PR, but it's not).

Also, comments like these ones:

> I don't know how many other ways to explain that Cygwin is not Linux

> on principle I still don't like it 

> this is why I wanted a module of common functionality between them [again]

 > better to start with something small, note in the documentation (as I have started to do) that it is partial and/or experimental, and then add on functionality as able. All the core functionality is already there. [I already said I'm not willing to deliver a partial implementation]

> we can try things your way for now 

>  In principle this could also be useful for implementing support on other platforms (if they exist?)  [they probably don't + we can't make assumptions for something which doesn't even exist]

> [referred to code unrelated with this PR] I don't like this. Didn't there used to be a psutil._exceptions module? Why did that go away? Having something like that would seem a bit less crazy than stuff like this... [come on - are you serious?]

...are not constructive and keep the both of us in this never ending loop in which we just disagree with each other and nothing happens. 

I believe we can easily move this forward. This is even a relatively easy task since most of the stuff is in psutil already, the differences re. the parsing of /proc on Linux vs. Cygwin are [minimal](https://github.com/giampaolo/psutil/pull/998#issuecomment-468885319), and for the rest we'll just reuse `_psutil_windows.c` implementation. In summary, it's just a matter of moving that stuff around and refactor it, and I have a pretty clear idea on how to do it. But I think it's fair that **you** get rewarded for the fine work and passion you've put into this so far, not me, and that's why I'm trying to steer direction instead of doing it myself. But in order to do that I'm not willing  to overturn a 10-years-old [Linux implementation](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py) just to make room for Cygwin the way you intended to (which is what this PR does, it even mixes Linux and Cygwin logic, which is a firm no-no). I fully understood your refactoring counter-proposal and I simply do not agree with it (sorry). Instead of bashing me and convince me to do otherwise just give me the benefit of the doubt (I do this for free). I hope this clarifies things for the better. Let me know if you're still willing to move this forward. I'll be happy if you will but *please* stop complaining.

embray:
I'm a little confused by your last comment. I think you might have missed something since I initially replied: You're quoting partly from a comment I made out of frustration and later deleted, because I agree it wasn't productive.

I even went ahead and reworked things I think closer to your way and I'm not sure if you actually looked at the code because I did almost exactly as you asked.

> I fully understood your refactoring counter-proposal and I simply do not agree with it

Like I said, I'm not sure what you mean here. Are you referring to the `_pslinux_common` module? Because I removed that as you asked.

Relatedly, you keep saying you disagree without giving any sort of justification. I've at least explained my reasoning and I'm not satisfied often that you have.

> mixes Linux and Cygwin

I don't understand the second link. The logic there is nothing specifically to do with Cygwin, and is just reworked a bit to be a bit clearer and easier to extend if needed.

> [referred to code unrelated with this PR]

I'm not sure what you mean by this. The comment you're mentioning here is exactly related to this PR, it's on code I added.

As for the bytes / str stuff as I explained last time we discussed it I wasn't satisfied with your reasoning behind rejecting the PR as-is because I foresaw future usage that you didn't. That's fair enough though, so I also explained that I would make a new PR *explicitly* incorporating it to make clearer the use I had in mind. So while you're still free to disagree with it, when you write

> you decided to include into this one anyway

You make it sound like I'm trying to make an end run around this, when in fact I'm doing exactly what I (think) I previously communicated was my plan in the first place.

> I believe we can easily move this forward.

Maybe but having done the actual work I don't think it's as easy as you think. You're welcome to either do it yourself or show some willingness to compromise as I have.

embray:
BTW I'm sorry again for being cranky about this. I think it's mostly not your fault. But I do wish you'd explain your rationale behind things better, as then it'd be easier to find understanding and compromises. Instead I just get absolutisms like

> [I already said I'm not willing to deliver a partial implementation]

embray:
Actually that last comment leaves me especially confused because there are many features in this package that are platform-dependent and some that are available on some platforms if not others, so I really don't see what the problem is, especially if it's documented as provisional and or an initial version.

Furthermore, I had just assumed you would be happier with a smaller, more minimal PR that would be easier to review. Again, I communicated clearly that that was my plan and you approved it. If you didn't like that idea I wish you had said so. 

Of course, I'm happy to build further on top of this one, though I figured it would be best to merge the initial work first (hence the care I took in stubbing out those tests that don't work yet; all other tests pass on my system). This way it can start being used and tested immediately in the development version while continuing the work of adding additional functionality.  Adding more will take some time in refactoring the Windows code. I'm particularly wary of using Windows API at all because Cygwin is not Windows. Regardless it's possible in some cases as a workaround to missing Cygwin features. Just not always without danger. It will take some additional care and I fear tbst stalling for too long will again prevent getting anything at all done.

giampaolo:
> You're quoting partly from a comment I made out of frustration and later deleted (nevermind, apparently not), because I agree it wasn't productive.
> I even went ahead and reworked things I think closer to your way and I'm not sure if you actually looked at the code because I did almost exactly as you asked.

My last comment was about the (not so nice) message you deleted, in which you kept insisting to use `_linux_common.py`, despite I previously said I disagree with it. I saw (later, since github apparently didn't email me about it) that you force-pushed a change which got rid of it and which deleted history. Because of that it took me a bit to figure out where `_linux_common.py` ended up. I noticed you removed it but I decided to make my point clear anyway, because you'll admit you've been quite insistent about this, and I was hoping we could stop the chit-chat and finally get this running.

> I don't understand the second link. The logic there is nothing specifically to do with Cygwin

You're right, my bad.

> I'm not sure what you mean by this. The comment you're mentioning here is exactly related to this PR

The reason [this](https://github.com/giampaolo/psutil/pull/1483#discussion_r273588487) is the way it is is because of #1402, which is unrelated with Cygwin.

> Maybe but having done the actual work I don't think it's as easy as you think. You're welcome to either do it yourself or show some willingness to compromise as I have.

I'm happy to compromise when I agree with your proposal and think it's reasonable, not because I necessarily have to.

> Furthermore, I had just assumed you would be happier with a smaller, more minimal PR that would be easier to review. Again, I communicated clearly that that was my plan and you approved it. If you didn't like that idea I wish you had said so.

You misunderstood me. I proposed to proceed step by step, which is different than  committing a draft PR. Compared to #998 this PR is surely a step in the right direction, but it's still a draft, useful to discuss how to proceed (together) and add functionality as we go.

Crucial things (because interconnected with other functionality) like Process `create_time()`, `ppid()`,  `pids()`, `memory_info()`, `name()`, are either not implemented or skipped in unit-tests, meaning they're broken, and you ignored my initial inline comments to this PR. I have and would have more of those as we proceed, because this is how the code review workflow is supposed to work, and for good reasons, not because I don't like you or something. 

There is no point in committing this as-is and it is inappropriate of you to tell me what should be part of a new release so insistently. My aim at delivering a close-to-complete implementation is because I believe it can be achieved. And even in case we decide to do otherwise (minimal functionality) as we go, it would at the very least pass the quality standards, meaning engaging in a code review first and progressively go from "draft" to "minimal" (but stable).

> Instead I just get absolutisms like [I already said I'm not willing to deliver a partial implementation]
> Why let something that works mostly be the enemy of something that works 100%? I don't think this is an unreasonable question.

The question is not unreasonable but if I already expressed my view on this twice in this very thread, why do you keep insisting?

> As for the bytes / str stuff as I explained last time we discussed it I wasn't satisfied with your reasoning behind rejecting the PR as-is because I foresaw future usage that you didn't.

I explained why I didn't foresee future usage: https://github.com/giampaolo/psutil/pull/1342#issuecomment-468682739

> Relatedly, you keep saying you disagree without giving any sort of justification.
> I've at least explained my reasoning and **I'm not satisfied often that you have**.

In the 10 years I've received contributions to this project for which I gave most of my spare time for free (including to you) it's the first time I've read something like this. I don't think you can delete these 2 PRs so if you can please do me a favor and do a "blank" push-force similarly to what you did above in order to remove your changes. I will eventually implement this myself from scratch sometime over the course of this year.

Issue #1484, opened 2019-04-09
[Windows] emulate UNIX os.getloadavg()


giampaolo:
I'm not sure I understand whether we can really emulate the UNIX counterpart, but this looks interesting:
https://bugs.python.org/issue34060
https://github.com/python/cpython/commit/e16467af0bfcc9f399df251495ff2d2ad20a1669
https://github.com/python/cpython/blob/master/Lib/test/libregrtest/win_utils.py

Part of the confusion is that I currently don't really understand how `getloadavg()` should be interpreted (need to read some literature) . My hope is to land something like:

```python
def getloadvg():get
     if POSIX:
        return os.getloadavg()
    else:
        return _windows_impl()
```       

...but returning 1 value (on Windows, average since last call) vs. 3 values (on UNIX, representing the average of last 1, 5, 15 minutes) maybe suggests these are 2 different things so it should be called differently?

ammaraskar:
Hey, it's entirely possible to do the 3 values.

The best literature to read would definitely be: https://en.wikipedia.org/wiki/Load_(computing)#Unix-style_load_calculation which explains really well how the value is calculated.

Essentially its the weighted moving average of the number of processes in the run queue of the OS over 1 minute, 5 minutes and 15 minutes.

You can find Linux's load average calculation here: https://github.com/torvalds/linux/blob/345671ea0f9258f410eb057b9ced9cefbbe5dc78/include/linux/sched/loadavg.h#L18-L38 which is what is emulated in https://github.com/python/cpython/blob/master/Lib/test/libregrtest/win_utils.py#L97

As the Linux comment eludes, the constants can be calculated by:
`1 / exp(sampling interval in seconds / window size in seconds)`

for example, in CPython we only care about the 1 minute window and we sample every 5 seconds, so the constant comes from:
`1 / exp(5 / 60) = 0.92004441462`

You can calculate out the constants for the other windows (5 minutes and 15 minutes) and track those separately as well like Linux does in order to fully emulate it: https://github.com/torvalds/linux/blob/345671ea0f9258f410eb057b9ced9cefbbe5dc78/kernel/sched/loadavg.c#L356-L358

Hope this helps and lemme know if you have any other questions.

ammaraskar:
Oh also, I should say the calculations are the easy part. Getting the actual number of processes in the run queue for windows is an absolute pain and we ended up running the command:

`typeperf \System\Processor Queue Length -si 5` 

[...]

PR #1485, opened 2019-04-11
Implement getloadavg on Windows. Fixes #604 and #1484

### LINKED ISSUE #604: Average system load metrics on Windows
Is there a way to query some kind of load average for Windows?
https://github.com/nicolargo/glances/issues/344

ammaraskar:
(ref. #604 and #1484)

I added some guards for Vista+ since I realized `PdhAddEnglishCounterW` is not available on XP.

There's a choice between calling `init_loadavg_counter` on import or on the first `getloadavg` call. I'm inclined to say the latter is better since then we don't introduce overhead when it isn't needed.

Not sure how to test properly this properly on the windows CI since there's nothing built in we can test against. There might be some external programs we can use but I'd like to get your take on it.


giampaolo (approved):
Excellent code. I'll merge when you'll give me the OK.

giampaolo:
> There's a choice between calling init_loadavg_counter on import or on the first getloadavg call. I'm inclined to say the latter is better since then we don't introduce overhead when it isn't needed.

I agree

> Not sure how to test properly this properly on the windows CI since there's nothing built in we can test against. There might be some external programs we can use but I'd like to get your take on it.

Probably there will be something in wmic but I wouldn't bother (too boring =)).

setup.py
```diff
@@ -135,12 +135,12 @@ def get_winver():
             'psutil/arch/windows/inet_ntop.c',
             'psutil/arch/windows/services.c',
             'psutil/arch/windows/global.c',
-            # 'psutil/arch/windows/connections.c',
+            'psutil/arch/windows/wmi.c',
```
giampaolo: Question for you: can this be considered WMI related code? I added this new file because some time in the future I am hoping to add sensors/temperatures metrics (which I know require WMI).

psutil/tests/__init__.py
```diff
@@ -159,6 +159,7 @@
 HAS_CONNECTIONS_UNIX = POSIX and not SUNOS
 HAS_CPU_AFFINITY = hasattr(psutil.Process, "cpu_affinity")
 HAS_CPU_FREQ = hasattr(psutil, "cpu_freq")
+HAS_GET_LOADAVG = hasattr(psutil, "getloadavg")
```
giampaolo: HAS_GETLOADAVG

docs/index.rst
```diff
@@ -238,6 +238,27 @@ CPU
...
     .. versionchanged:: 5.5.1 added FreeBSD support.
 
+.. function:: getloadavg()
+
+    Returns the average load on the system over the last 1 minute, 5 minutes
+    and 15 minutes respectively as a tuple. The load represents how many
+    processes are waiting to be run by the operating system.
+
+    On \*NIX systems this passes through the value of `os.getloadavg`_ and on
+    Windows psutils emulates the behavior by calculating the load internally.
```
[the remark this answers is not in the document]
giampaolo: (I updated my comment)

setup.py
ammaraskar: Yeah it can, performance counters are just an abstraction on top of WMI, check out this architectural diagram here: https://docs.microsoft.com/en-us/windows/desktop/perfctrs/about-performance-counters

ammaraskar:
I've gone ahead and incorporated the feedback from the code review along with updating `install.sh` to install a more recent version of python2 (the old one was incompatible with the latest openssl) in order to fix the CI.

giampaolo:
Nice, thanks Ammar.

vstinner:
Cool stuff :+1:

vstinner:
Python also got its toy: https://github.com/python/cpython/pull/8357/files

giampaolo:
Cool stuff indeed. I had been floating around this for a while back in the days (the original proposal is old). There was/is a bunch of info on internet mentioning the bits with which it's theoretically possible to do this ("System Processor Queue Length") but I couldn't find any real implementation (e.g. knowing how to calculate LOADAVG_FACTOR_1F magic numbers is non trivial and requires some actual understanding of how this works).

bmxp:
# getloadavg() from psutil 5.6.2 crashes with python 3.7.x 

## Description

Please test the following python code:

```
import psutil
a = psutil.getloadavg()
```
Wait a couple of seconds and Python will say quietly good bye

## Versions

* psutil 5.6.2
* Python 3.7.1 x32 or Python 3.7.3 x32
* Operating System Windows 10 Pro Build 1809 (Build 17763.437)

I checked this on a Windows Server 2012 R2 with either Python 3.5.2 and 3.6.6 
and the above code is working fine, also ob and older Windows 10 build 1607 with Python 3.6.6 no problems arose.

Could you please have a closer look at Python 3.7.x with psutil.getloadavg()

Artoria2e5:
A few things that can be improved, which I will PR:
* The 5 second interval is very prone to moire artifacts, see https://ripke.com/loadavg/moire. I recommend using (60/13).
* Queue length do not include current running count, only the waiting-count, unlike Unix load average. You gotta add processor util% to match.
* Depending if you want to match Linux or other Unix-like systems, you may want to add disk queue length.

Issue #1456, opened 2019-03-11
[CentOS] cpu_freq does not return proper type in some cases


cschanzlenist:
Linux 
CentOS 7.6
Name: psutil
Version: 5.6.1

Per https://psutil.readthedocs.io/en/latest/ , cpu_freq should return 0 if min and max cannot be determined.

However, the code (_pslinux.py line 722) can return None, which breaks the s-util package I was trying to use due to a type mismatch.

Either the documentation should be corrected or the code (likely the code) with:
```
722c722
<                     ret.append(_common.scpufreq(float(value), None, None))
---
>                     ret.append(_common.scpufreq(float(value), 0.0, 0.0))

```
For others to reference/find, here's the traceback that s-tui experienced:

```
$ s-tui 
Traceback (most recent call last):
  File "/home/user/venv/bin/s-tui", line 11, in <module>
    sys.exit(main())
  File "/home/user/venv/lib/python3.6/site-packages/s_tui/s_tui.py", line 927, in main
    graph_controller = GraphController(args)
  File "/home/user/venv/lib/python3.6/site-packages/s_tui/s_tui.py", line 728, in __init__
    self.view = GraphView(self)
  File "/home/user/venv/lib/python3.6/site-packages/s_tui/s_tui.py", line 233, in __init__
    urwid.WidgetPlaceholder.__init__(self, self.main_window())
  File "/home/user/venv/lib/python3.6/site-packages/s_tui/s_tui.py", line 531, in main_window
    freq_source = FreqSource(is_admin)
  File "/home/user/venv/lib/python3.6/site-packages/s_tui/Sources/FreqSource.py", line 118, in __init__
    if self.last_freq >= 0 and self.top_freq <= 0:
TypeError: '<=' not supported between instances of 'NoneType' and 'int'

```

giampaolo:
Can you make a PR?

cschanzlenist:
IF PR means pull request, sorry, no - not an experience git user and don't have a couple hours to learn right now. :-)  The one-liner fix is above.

amanusk:
@cschanzlenist, @giampaolo  if it's OK by you, I have opened a PR to fix the issue #1487

giampaolo:
Fixed by https://github.com/giampaolo/psutil/pull/1895.

PR #1487, opened 2019-04-12
Update cpu_freq to return 0 for max/min if not available

amanusk:
This PR fixes issue #1456

Return 0.0 for cpu_freq min and max values if not available.


giampaolo:
Thanks Alex, can you also update NOTES / HISTORY?
Note to self: this is correct because already documented.

PR #1493, opened 2019-04-23
Fix cpu_freq

davidbrochart:
I have a Linux machine where:
- `/sys/devices/system/cpu/cpufreq/` exists but is empty.
- `/sys/devices/system/cpu/cpu0/` exists but doesn't have `cpufreq` in it.

In this case, use `/proc/cpuinfo`.


giampaolo:
Reverted as it broke some tests.

PR #1496, opened 2019-04-25
Fix cpu freq

davidbrochart:
This should fix the tests.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -716,9 +716,9 @@ elif os.path.exists("/proc/cpuinfo"):
                         if line.lower().startswith(b'cpu mhz'):
                             key, value = line.split(b'\t:', 1)
                             ret.append(_common.scpufreq(float(value), 0., 0.))
-            except FileNotFoundError:
+            except IOError as err:
                 raise NotImplementedError(
-                    "can't open file %s" % path)
+                    "%r for file %r" % (err, path))
         return ret
 
 else:
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -668,7 +668,7 @@ def cpu_stats():
         ctx_switches, interrupts, soft_interrupts, syscalls)
 
 
-if os.path.exists("/sys/devices/system/cpu/cpufreq") or \
+if os.path.exists("/sys/devices/system/cpu/cpufreq/policy0") or \
         os.path.exists("/sys/devices/system/cpu/cpu0/cpufreq"):
     def cpu_freq():
         """Return frequency metrics for all CPUs.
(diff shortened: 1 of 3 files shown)


psutil/_pslinux.py
```diff
@@ -708,13 +708,25 @@ def cpu_freq():
...
+        if os.path.exists(path):
+            try:
+                with open_binary(path) as f:
+                    for line in f:
+                        if line.lower().startswith(b'cpu mhz'):
+                            key, value = line.split(b'\t:', 1)
+                            ret.append(_common.scpufreq(float(value), 0., 0.))
+            except IOError as err:
+                raise NotImplementedError(
+                    "%r for file %r" % (err, path))
```
giampaolo: Mmm... Not sure I understand why catching IOError is necessary here. Isn't the function declaration guarded by  `if os.path.exists("/proc/cpuinfo"):` already?

psutil/_pslinux.py
```diff
@@ -708,13 +708,25 @@ def cpu_freq():
...
         return ret
 
+else:
+    def cpu_freq():
+        """Dummy implementation when none of the above files are present.
+        """
+        return [_common.scpufreq(0.0, 0.0, 0.0)] * cpu_count_logical()
```
giampaolo: As per documentation:

>  If *min* and *max* cannot be determined they are set to ``0``.

In here we're including `current`. We shouldn't do that. Please return an empty list instead.

giampaolo (changes requested):

psutil/_pslinux.py
davidbrochart: It's because of this test: https://github.com/davidbrochart/psutil/blob/fix_cpu_freq/psutil/tests/test_linux.py#L859-L868

psutil/_pslinux.py
davidbrochart: You mean `return []` or `return [[] for _ in range(cpu_count_logical())]`?

giampaolo: `return []`

psutil/_pslinux.py
davidbrochart: And this test: https://github.com/davidbrochart/psutil/blob/56fa1df9bccfc837d9354ebbcc84306dbb7b734c/psutil/tests/test_linux.py#L699-L702

giampaolo:
I fixed this in c2b333b7dc051e9c0b62b619a35c6f4a321d0b5b. That should get you going.
I think these unit-tests should be rewritten (they don't fail on my machine but it'll likely fail on yours) but I'll get to this later as I want to make a release soon. I'm gonna close this for now. Thanks for the PR.

giampaolo:
...and... it turns out I was wrong. =)
Tests fails on Travis too after c2b333b, so this PR is necessary. Reopening.

PR #1505, opened 2019-05-07
Fix #1494: [AIX] implement Process.environ()

### LINKED ISSUE #1494: [AIX] implement Process.environ()
(ended up discovering this from https://github.com/giampaolo/psutil/issues/1276#issuecomment-483580196)

Looks like we can use `getevars` to determine process environ():
https://sites.ualberta.ca/dept/chemeng/AIX-43/share/man/info/C/a_doc_lib/libs/basetrf1/getevars.htm#ojq2p21aendr
EBADF should be used as an alias for `NoSuchProcess`. Not sure what happens in case of permission error (I would expect something like EACCES but it's not mentioned in the man page). 
I'm merely opening this ticket in case @wiggin15 or somebody else wants to take over as I have no AIX box to test against.

wiggin15:


giampaolo:
No complaints here. =)
Merging.

giampaolo:
While I'm at it: how are tests going? Do `make test` and `make test-memleaks` all pass on AIX?

wiggin15:
There are some test failures. That's my next PR which I already prepared :)

PR #1507, opened 2019-05-07
Fix some tests on AIX

wiggin15:


scripts/ps.py
```diff
@@ -85,7 +87,7 @@ def main():
         status = p.info['status'][:5] if p.info['status'] else ''
 
         line = templ % (
-            user[:10],
+            user,
```
giampaolo: This is needed in order to keep the formatting sane in case the username is too long. Other than this, LGTM.

giampaolo (changes requested):

scripts/ps.py
wiggin15: We already limit `user` to 9 characters in line 75 so this is redundant.

giampaolo: OK

PR #1532, opened 2019-06-12
flake8 --max-complexity -- Let's look at code complexity

cclauss:
Most projects can pass about [__flake8 --max-complexity=10__](http://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-complexity).  https://github.com/PyCQA/mccabe

### DIFF
--- a/.ci/travis/run.sh
+++ b/.ci/travis/run.sh
@@ -30,7 +30,7 @@ if [ "$PYVER" == "2.7" ] || [ "$PYVER" == "3.6" ]; then
     PSUTIL_TESTING=1 python -Wa psutil/tests/test_memory_leaks.py
     # run linter (on Linux only)
     if [[ "$(uname -s)" != 'Darwin' ]]; then
-        python -m flake8 --max-complexity=25
+        python -m flake8 --max-complexity=10
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


giampaolo:
Mmmm... interesting. I wasn't aware of this. I tried it and I got this:

```
psutil/_compat.py:47:1: C901 'TryExcept 47' is too complex (31)
psutil/tests/__init__.py:462:1: C901 'reap_children' is too complex (27)
scripts/procinfo.py:137:1: C901 'run' is too complex (52)
```
In 2 cases out of 3 I agree the functions are probably too complex, but am also sort of fine with that and would not want to change them. In summary, I think it would be too strict to make this part of the CI process. It may make sense to have this as a make target though (`make codequality / codecomplexity` or something). Actually it would be good to have a make target which analyzes code even more deeply, similarly to what codacy does: 
https://app.codacy.com/project/g-rodola/psutil/dashboard
...and prints warnings on screen. Are you aware if there is something like that?

cclauss:
I am not aware of other tools that create a complexity dashboard.  I dropped it down to the more standard --max-complexity=10 just to see some simpler functions that might be worth refactoring.

PR #1135, opened 2017-09-26
Add support for thread names

fthiery:
Rebase of the outdated 613-thread-names branch

PR #1559, opened 2019-07-31
thread names : linux and windows support

romainpereira:
This patch adds support for thread names on Linux and Windows, by adding an attribute **name** within the tuple returned by ```psutil.threads()```
On other operating system, the string is empty

The code is probably over-commented, we can clean this.

1 - Shall we remove the MacOSX comments, or add them as a comment of this PR?

2 - Remove the **thread-names-demo/** folder, which is only here giving some small sample program to be tested

3 - Here is another way of naming thread in Windows, to have thread names appearing in Visual Studio debugger:
https://blogs.msdn.microsoft.com/stevejs/2005/12/20/naming-threads-in-win32-and-net/
I think the debugger see the exception while reading process's task, and interpret the exception message as the thread name.

The ```SetThreadDescription()``` call is really likely to be used to name threads now.

4 - Do we really reject the use of **unicode** string? C.f [psutil/tests/test_contracts.py:194](https://github.com/giampaolo/psutil/blob/master/psutil/tests/test_contracts.py)
We probably want them on the name field, since Windows thread names are wide characters string.

---
_Please, note that a previous thread name support implementation was tried.
This patch is originally inspired by this pull-request: https://github.com/giampaolo/psutil/pull/1135
on which both **UbiCastTeam** and **fthiery** worked._


romainpereira:
Sorry for various failures on CIs, got some false positive locally, that's why.
I am currently fixing actual issues.

rpereira-dev:
@giampaolo I allow myself to ping you, since no one reacted in 2 months

Did you give a look to the patch?
Do you know what's making the CI (appveyor) failing for Python 2.7 and 3.5?

Thank you very much!

giampaolo:
Hello. No I didn’t have time to look at the patch (and I will still be busy for a while), but in principle I agree it would be good to have thread names. If in the meantime you could take a look at BSD and OSX implementation that would be great.

rpereira-dev:
I no longer have access to a OSX machine :(
But as far as I dig last month, there is no 'easy' way to access the thread names of any process on OSX (as _GetThreadDescription()_ system call on Windows, or reading in _/proc/..._ on Linux)

See my comments on  **psutil/_psutil_osx.c**

giampaolo:
Can you please remove the demos and other unnecessary stuff?

rpereira-dev:
> Can you please remove the demos and other unnecessary stuff?

Done. Not sure about what's making the CI failing though

psutil/_pslinux.py
```diff
@@ -1914,7 +1913,12 @@ def threads(self):
             values = st.split(b' ')
             utime = float(values[11]) / CLOCK_TICKS
             stime = float(values[12]) / CLOCK_TICKS
-            ntuple = _common.pthread(int(thread_id), utime, stime)
+            try:
+                with open_binary("%s/comm" % tdir) as fcomm:
```
giampaolo: you shoud use `open_text` instead

psutil/_psutil_windows.c
```diff
@@ -1143,7 +1150,33 @@ psutil_proc_threads(PyObject *self, PyObject *args) {
...
+            // https://wiki.python.org/moin/WindowsCompilers
+            if (!get_thread_desc_set) {
+                get_thread_desc_set = 1;
+                windll = GetModuleHandle("Kernel32.dll");
+                if (windll) {
+                    get_thread_desc = (get_thread_desc_t) GetProcAddress(windll, "GetThreadDescription");
+                }
+                get_thread_desc = (get_thread_desc_t) GetProcAddress(
+                    GetModuleHandle("Kernel32.dll"), "GetThreadDescription"
+                );
```
giampaolo: This stuff should be done differently. 
You should first define the prototype of `GetThreadDescription` in `psutil/arch/windows/ntext.h` then load `GetThreadDescription` and set it as a global name in `psutil/_psutil_common.c`.

PR #1574, opened 2019-08-26
Fix for Python 4

hugovk:
We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule).

There's some code which checks the Python major version is exactly 3, similar to this:

```python
if sys.version_info[0] == 3:
    # Python 3+ code
else:
    # Python 2 code
```

When run on Python 4, it will run the Python 2 code! Instead:

```python
if sys.version_info[0] >= 3:
    # Python 3+ code
else:
    # Python 2 code
```

Found using https://github.com/asottile/flake8-2020:
```console
$ pip install -U flake8-2020
...
$ flake8 --select YTT
./scripts/internal/winmake.py:34:7: YTT201: `sys.version_info[0] == 3` referenced (python4), use `>=`
./psutil/_common.py:41:7: YTT201: `sys.version_info[0] == 3` referenced (python4), use `>=`
./psutil/_compat.py:18:7: YTT201: `sys.version_info[0] == 3` referenced (python4), use `>=`
```

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -38,7 +38,7 @@ else:
     enum = None
 
 # can't take it from _common.py as this script is imported by setup.py
-PY3 = sys.version_info[0] == 3
+PY3 = sys.version_info[0] >= 3
 
 __all__ = [
     # constants
--- a/psutil/_compat.py
+++ b/psutil/_compat.py
@@ -15,7 +15,7 @@ __all__ = ["PY3", "long", "xrange", "unicode", "basestring", "u", "b",
            "FileNotFoundError", "PermissionError", "ProcessLookupError",
            "InterruptedError", "ChildProcessError", "FileExistsError"]
 
-PY3 = sys.version_info[0] == 3
+PY3 = sys.version_info[0] >= 3
 
 if PY3:
     long = int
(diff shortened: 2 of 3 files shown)


giampaolo:
We still don't know what Python is gonna be called by then, which (if I'm not mistaken) should be +3 years from now, so let's cross that bridge when we come to it. ;)

PS: my hunch is it's gonna be called 3.10, precisely to avoid this kind of breakage.

hugovk:
Yeah, we don't know which it will be and 3.10 will also have problems (eg. `sys.version[:3]` returns 3.1, not 3.10!) but probably less severe than 4.0.

And it'll be sooner than you think. 

Python 3.9 reaches beta in February 2020, and that is when the Python 3.9+1 branch will be made and testing will begin. That's only 6 months from now. 

There's a proposal in the 3.9 PEP to move to a 9 month release cadence, which could mean 3.9+1 final is out around November 2020, 1 year and 3 months from now.

If they stick with an 18-month cadence, the final will be out around August 2021, two years from now.

Do you know if psutil is vendored in any other packages?

giampaolo:
> Yeah, we don't know which it will be and 3.10 will also have problems (eg. sys.version[:3] returns 3.1, not 3.10!)

We use `sys.version_info[0]` though so 3.10 won't be a problem.

> Do you know if psutil is vendored in any other packages?

Not that I'm aware of, no.

PR #1588, opened 2019-09-18
Removed a typo/copy'n'paste mistake Linux->Windows

Barry1:
In Line 2222 the IOPRIO-constants for windows have been explaning with the words "A set of integers representing the I/O priority of a process on Linux." which seems to be wrong as the availability shows Windows.


giampaolo:
Good catch, thanks.

PR #1591, opened 2019-09-20
Install the scripts along with the package.

### LINKED ISSUE #1560: Include scripts in PyPI packages
The scripts/ directory is full of tools that are very useful—but they're not included in what gets installed via `pip`. It would be great if they were, probably with names like `psutil-top` so they don't conflict.

If you're interested I can probably do a PR.

itamarst:
Fixes #1560.


giampaolo:
Those scripts didn't receive enough love so that they can be installed. For instance, most of them don't even have a cmdline parser. I don't think it's worth it, sorry.

PR #1597, opened 2019-09-29
Don't pollute tree with temporary configure test files

nox:
The docs for CCompiler.compile say:

> If output_dir is given, object files will be put under it, while retaining their
> original path component. That is, foo/bar.c normally compiles to foo/bar.o (for a
> Unix implementation); if output_dir is build, then it would compile to build/foo/bar.o.

What they forget to say is that path components are also retained if output_dir is not
specified, it just means it will do so in the current directory. So if you compile
a temporary C file /tmp/foo.c, it will produce a ./tmp/foo.o file relative to the
current directory.

This commit fixes that issue by passing an explicit output_dir itself located in
a temporary directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,7 @@ clean:  ## Remove all build files.
 		build/ \
 		dist/ \
 		docs/_build/ \
-		htmlcov/ \
-		tmp/
+		htmlcov/
 
 _:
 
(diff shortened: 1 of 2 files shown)


nox:
(I don't have a Linux box so I didn't test that patch locally.)

PR #1599, opened 2019-10-04
psutil : Added support to get CPU affinity for multi core servers 

ossdev07:
[issue]
As per logic written in "test_cpu_affinity_all_combinations", it stores all the combinations in a list,
in our machine (having 96 cores) the process gets terminated because of out of memory issue 
and we observed that it is working fine upto 8 core machines.

[Fix]
This issue is fixed by implementing the logic in which list will not be stored and combinations will be checked in run time


giampaolo:
I added a fix for this some time ago. Closing.

PR #1624, opened 2019-11-18
psutil/_psutil_posix.c: better clear variables to ensure they are NULL

ret2libc:
Sorry, for some reasons _psutil_posix was left out from PR #1616 . I do not think this deserves a separate CVE as function `psutil_net_if_addrs` deals with  network interfaces info and a low-privilege attacker should not be able to manipulate them.


giampaolo:
Grazie Riccardo.

Issue #1635, opened 2019-12-03
[Linux] several test failures


DerDakon:
**Platform**
* Linux, Sparc T5120
* 5.6.7

**Bug description**
Several test failures, as below. Downstream issue: https://bugs.gentoo.org/701918

**Test results**
```
======================================================================
ERROR: psutil.tests.test_system.TestSystemAPIs.test_sensors_battery
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/tests/test_system.py", line 854, in test_sensors_battery
    ret = psutil.sensors_battery()
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/__init__.py", line 2397, in sensors_battery
    return _psplatform.sensors_battery()
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/_pslinux.py", line 1345, in sensors_battery
    bats = [x for x in os.listdir(POWER_SUPPLY_PATH) if x.startswith('BAT')]
OSError: [Errno 2] No such file or directory: '/sys/class/power_supply'

======================================================================
ERROR: psutil.tests.test_linux.TestSensorsBattery.test_emulate_energy_full_0
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/tests/test_linux.py", line 1523, in test_emulate_energy_full_0
    self.assertEqual(psutil.sensors_battery().percent, 0)
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/__init__.py", line 2397, in sensors_battery
    return _psplatform.sensors_battery()
  File "/var/tmp/portage/dev-python/psutil-5.6.7/work/psutil-5.6.7-python2_7/lib/psutil/_pslinux.py", line 1345, in sensors_battery
    bats = [x for x in os.listdir(POWER_SUPPLY_PATH) if x.startswith('BAT')]
OSError: [Errno 2] No such file or directory: '/sys/class/power_supply'

======================================================================
ERROR: psutil.tests.test_linux.TestSensorsBattery.test_emulate_energy_full_not_avail
----------------------------------------------------------------------
[...]

PR #1643, opened 2019-12-18
Fix some test fallout found on Linux/Sparc

DerDakon:
This fixes most of the problems reported in #1635.

### DIFF
--- a/psutil/tests/test_contracts.py
+++ b/psutil/tests/test_contracts.py
@@ -173,8 +173,10 @@ class TestAvailProcessAPIs(PsutilTestCase):
         self.assertEqual(hasattr(psutil.Process, "rlimit"), LINUX or FREEBSD)
 
     def test_io_counters(self):
+        not_on_linux = (LINUX and not os.path.exists('/proc/1/io'))
         hasit = hasattr(psutil.Process, "io_counters")
-        self.assertEqual(hasit, False if MACOS or SUNOS else True)
+        self.assertEqual(hasit,
+                         False if MACOS or SUNOS or not_on_linux else True)
 
     def test_num_fds(self):
         self.assertEqual(hasattr(psutil.Process, "num_fds"), POSIX)
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1393,6 +1393,8 @@ def sensors_battery():
                 return int(ret) if ret.isdigit() else ret
         return None
 
+    if not os.path.isdir(POWER_SUPPLY_PATH):
+        return None
     bats = [x for x in os.listdir(POWER_SUPPLY_PATH) if x.startswith('BAT') or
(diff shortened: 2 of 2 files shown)


DerDakon:
I don't understand the test error, looks unrelated to me.

DerDakon:
ping

DerDakon:
ping?

DerDakon:
Ping?

psutil/tests/test_contracts.py
```diff
@@ -173,8 +173,10 @@ def test_rlimit(self):
         self.assertEqual(hasattr(psutil.Process, "rlimit"), LINUX or FREEBSD)
 
     def test_io_counters(self):
+        not_on_linux = (LINUX and not os.path.exists('/proc/1/io'))
```
giampaolo, line 176: what's the meaning of this?

DerDakon: It's Linux, but the thing doesn't exist here. I'm open for any better name.

giampaolo: I see. It's OK, let's let it fail.

DerDakon: It will constantly fail e.g. on my sparc machine because this simply does not exist there, annoying me during the stabilizations for Gentoo because the testsuite is marked as failed.

giampaolo: These tests are supposed to be "alarms" testing API availability depending on the OS. I prefer to have them failing.

DerDakon: But the API is not generally available on Linux. That's all what I'm testing here: if these iocounters are available. Thinking about that this test could result in a false negative for the (unlikely) case that one has locked down /proc access more than usual, e.g. to not show processes of other users.

If you don't like this one just cherry-pick the first one.

giampaolo: > But the API is not generally available on Linux

It is generally available, at least on the majority of Linux i686 / x86 archs / distros I've seen. Most likely it's not available on Sparc, in which case I prefer to know with an explicit failure. 
I understand it is annoying to see a failure on your  specific arch / distro, but (in my mind) these tests checking API availability assume psutil is running on the most common arch / distro, not the "uncommon one", and are designed to (fail) be like this on purpose.

> If you don't like this one just cherry-pick the first one.

Wouldn't know how to do that.

DerDakon: This is a kernel configuration thing. We have even bug reports that [this is failing on amd64](https://bugs.gentoo.org/701962) ([and again](https://bugs.gentoo.org/710646)).

giampaolo:
The issue this was fixing (#1635) is now closed as stale, and the file it patches (test_contracts.py) doesn't exist anymore. Sorry for leaving you hanging on this one. Closing.

PR #1646, opened 2019-12-21
Fix Process on 12.0+ i386

sunpoet:
FreeBSD 12.0+ change ki_tdev from 32 bits to 64 bits.

Reference:	https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=242543


sunpoet:
Thanks. BTW, the credit should go to Antoine Brodin (https://github.com/ant1) who provides the patch.

PR #1652, opened 2020-01-02
Drop support for Windows XP and Server 2003

giampaolo:
Win XP support was probably broken anyway as I haven't been tested psutil on XP for years. This PR remove all the cruft related to it. Minimum supported Windows client now is Windows Vista.


giampaolo:
Windows Server 2003 support dropped as of ac61e73.

PR #1660, opened 2020-01-09
[Windows] rewrite of open_files() 

giampaolo:
The original code was messy and had different issues, the main one being there was zero error checking (I suppose in that case we got an empty list). Amongst other things I got rid of global vars (except one), thread events and took inspiration from ProcessHacker source code for the `NtQuerySystemInformation` and `NtQueryObject` parts which are more robust.

Note: the previous version reused the same thread to get the file handle name till the thread hanged (in which case it was terminated) emulating some sort of process pool. This new version serially spawns one thread for each handle on every loop instead. This is a bit slower, but much more sane in terms of code. An alternative to this would be using some sort of IPC method (queue, thread pool) but it's outside of my current expertise.

Finally, this is nicer to (future) Cygwin integration (previous version caused segfault).

Update: related to #340 and #596. Also CC @mrjefftang (original author).

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -20,6 +20,7 @@ XXXX-XX-XX
   due to a backward incompatible change in a C type introduced in 12.0.
 - 1656_: [Windows] Process.memory_full_info() raises AccessDenied even for the
   current user and os.getpid().
+- 1660_: [Windows] Process.open_files() complete rewrite + check of errors.
 
 5.6.7
 =====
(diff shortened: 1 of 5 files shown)


giampaolo:
Related to #340 and 596. Also CC @mrjefftang (original author).

giampaolo:
Related to #340 and #596. Also CC @mrjefftang (original author).

PR #1667, opened 2020-01-18
Add *new_only* parameter for process_iter()

giampaolo:
When true, `process_iter()` will yield new process only (meaning new PIDs appeared since last call):

```python
>>> import psutil
>>> for proc in psutil.process_iter(attrs=['pid', 'name'], new_only=True):
...     print(proc.info)
...
```

This comes from a real-world use case I bumped into: I spawn a subprocess but don't know its PID, so I look for it by checking the command line of all processes. Instead of iterating over all processes, I want to iterate over new PIDs only (faster).

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -656,3 +656,7 @@ I: 1646
 N: Javad Karabi
 W: https://github.com/karabijavad
 I: 1648
+
+N: Mike Hommey
+W: https://github.com/glandium
+I: 1665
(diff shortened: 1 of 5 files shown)


giampaolo:
On a second thought, I realized that `process_iter()` uses a global variable, so it's not thread safe.
Despite this is acceptable when using it in the standard way, this case is different: if the are 2 threads using `process_iter()`, the first one calling it (+ consuming the iterator) will "steal" the processes of the second thread. That aside, it's unclear how to solve this problem in terms of API, so I reverted this PR in c9fc4fd. Problem solved. 

We have another one though: `psutil.cpu_percent()`. Since it also uses a global var, multiple threads will "reset" the internal timings of each other and invalidate result. Not sure why I realized this only now but the doc should be more clear about this.

Anyway, I'll discuss this in another ticket.

PR #1676, opened 2020-01-31
Optimizations of psutil_net_connections under windows.

mlabonte:
- Pre-allocate table memory so we take a chance to call it only once.
- Give some buffer in case the table size is growing fastely.

When the table is really big and a lot of new connections are establishing, the loop to retry allocating the memory keeps going. Adding a small buffer helps a lot. Also, as a single call can takes seconds, it worths to pre-allocate its memory so we take a chance to call it once.

### DIFF
--- a/psutil/arch/windows/socks.c
+++ b/psutil/arch/windows/socks.c
@@ -19,6 +19,8 @@
 typedef DWORD (WINAPI * TYPE_GetExtendedTcpTable)();
 typedef DWORD (WINAPI * TYPE_GetExtendedUdpTable)();
 
+DWORD g_dwRoomyTcpTableSize = 0;
+DWORD g_dwRoomyUdpTableSize = 0;
 
 static DWORD __GetExtendedTcpTable(TYPE_GetExtendedTcpTable call,
                                    ULONG family,
@@ -32,11 +34,21 @@ static DWORD __GetExtendedTcpTable(TYPE_GetExtendedTcpTable call,
     // See https://github.com/giampaolo/psutil/pull/1335 concerning 0xC0000001 error
     // and https://github.com/giampaolo/psutil/issues/1294
     DWORD error = ERROR_INSUFFICIENT_BUFFER;
-    *size = 0;
-    *data = NULL;
-    error = call(NULL, size, FALSE, family, TCP_TABLE_OWNER_PID_ALL, 0);
+    
+    // The call to get the table can be expensive. By pre-allocating the buffer,
+    // we take a chance to call it only once.
+    *size = g_dwRoomyTcpTableSize;
+    if (*size) {
+        *data = malloc(*size);
+    }
+    else {
+        *data = NULL;
+    }
+    
+    error = call(*data, size, FALSE, family, TCP_TABLE_OWNER_PID_ALL, 0);
     while (error == ERROR_INSUFFICIENT_BUFFER || error == 0xC0000001)
     {
+        *size = *size * 3 / 2;  // Give some buffer in case the table is growing fastely.
         *data = malloc(*size);
         if (*data == NULL) {
             error = ERROR_NOT_ENOUGH_MEMORY;
@@ -56,6 +68,7 @@ static DWORD __GetExtendedTcpTable(TYPE_GetExtendedTcpTable call,
         PyErr_SetString(PyExc_RuntimeError, "GetExtendedTcpTable failed");
         return 1;
     }
+    g_dwRoomyTcpTableSize = *size * 3 / 2;
     return error;
 }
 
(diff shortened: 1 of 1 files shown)


psutil/arch/windows/socks.c
```diff
@@ -56,6 +68,7 @@ static DWORD __GetExtendedTcpTable(TYPE_GetExtendedTcpTable call,
         PyErr_SetString(PyExc_RuntimeError, "GetExtendedTcpTable failed");
```
mlabonte:
```suggestion
        if (*data) {
            free(*data);
            *data = NULL;
        }
        PyErr_SetString(PyExc_RuntimeError, "GetExtendedTcpTable failed");
```
I should include it to avoid memory leak but I want to know if you like the idea behind the two changes. I can split them in two different Pull requests if you prefer.

psutil/arch/windows/socks.c
```diff
@@ -96,6 +119,7 @@ static DWORD __GetExtendedUdpTable(TYPE_GetExtendedUdpTable call,
         PyErr_SetString(PyExc_RuntimeError, "GetExtendedUdpTable failed");
```
mlabonte: Idem
```suggestion
        if (*data) {
            free(*data);
            *data = NULL;
        }
        PyErr_SetString(PyExc_RuntimeError, "GetExtendedUdpTable failed");
```

giampaolo:
Did you benchmark this? How much faster is it?

mlabonte:
Best case (fiew connections):
$ python -m timeit "import psutil; psutil.Process().connections(kind='tcp4')"
1000 loops, best of 3: 469 usec per loop       <--- psutil==5.6.7
1000 loops, best of 3: 231 usec per loop       <--- pull request

Normal case (15 K connections):
$ python -m timeit "import psutil; psutil.Process().connections(kind='tcp4')"
10 loops, best of 3: 75.1 msec per loop       <--- psutil==5.6.7
10 loops, best of 3: 37.7 msec per loop       <--- pull request

Worsth case (end of a quick rampup from 0 K to 15 K connections):
$ python -m timeit -n 1 -r 1 "import psutil; psutil.Process().connections(kind='tcp4')"
1 loops, best of 1: 25.5 sec per loop       <--- psutil==5.6.7
1 loops, best of 1: 259 msec per loop     <--- pull request

The upper limit of connections I expect on our system is 1 M connections. In that case, a single call should be 2.5 seconds (CPU bound) without the PR and 1.3 second with the PR. Of course, I'm more affraid of the worsth case, e.g. a quick rampup of 1 M connections after a maintenance.

giampaolo:
OK, let's see. It's been a while since I took a look at that part of the code. I did it tonight, I realized it was sort of messy so I refactored it in https://github.com/giampaolo/psutil/pull/1678. At least now I understand what's going on.  

With that said, what's the part which takes more time? The first call to `GetExtendedTcpTable(NULL, ...)`? Is the second call to `GetExtendedTcpTable` faster? 

I'm not sure using the size of the last time `net_connections()` was call is wise since the number of connections can be completely different. It's based on pure guessing. Why not using a fixed increased size then (8192, 16536, ...) and avoid calling `GetExtendedTcpTable` at all then? (assuming that speed things up, which I'm not sure).

Despite the speedup is consistent I'm "meh" about using a global var and I am "meh" about guessing the size like this, since I'm not sure what the repercussions may be. Just trying to understand... =)

mlabonte:
The 2 calls take the same time, that's why it's a good idea to guess the size, adding a small margin. In my use case, we monitor the tcp connections periodically, so the table size is often similar and the guessed value is adjusted after each call. I found a 50% margin was a good compromise with memory usage since I expect it could use easily many MB of memory.

giampaolo:
How about this?

```c
DWORD g_TcpTableSize = 0;


static PVOID __GetExtendedTcpTable(ULONG family) {
    DWORD err;
    PVOID table;
    ULONG size;

    size = g_TcpTableSize;
    if (size == 0) {
        GetExtendedTcpTable(NULL, &size, FALSE, family,
                            TCP_TABLE_OWNER_PID_ALL, 0);
        // reserve 25% more space
        size = size + (size / 2 / 2);
        g_TcpTableSize = size;
    }

    table = malloc(size);
    if (table == NULL) {
        PyErr_NoMemory();
        return NULL;
    }

    // get connections
    err = GetExtendedTcpTable(table, &size, FALSE, family,
                              TCP_TABLE_OWNER_PID_ALL, 0);
    if (err == NO_ERROR)
        return table;

    free(table);
    if (err == ERROR_INSUFFICIENT_BUFFER || err == STATUS_UNSUCCESSFUL) {
        psutil_debug("GetExtendedTcpTable: retry with different bufsize");
        g_TcpTableSize = 0;
        return __GetExtendedTcpTable(family);
    }

    PyErr_SetString(PyExc_RuntimeError, "GetExtendedTcpTable failed");
    return NULL;
}

```

mlabonte:
I like it, it'a cleaner than my PR. I benchmarked it and it gives the same results for my basic cases. :-)

giampaolo:
Super seeded by #1679. Closing.

PR #1679, opened 2020-02-03
[Windows] speedup connections

giampaolo:
...a variant of https://github.com/giampaolo/psutil/pull/1676.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,6 +16,8 @@ XXXX-XX-XX
 - 1671_: [FreeBSD] add CI testing/service for FreeBSD (Cirrus CI).
 - 1677_: [Windows] process exe() will succeed for all process PIDs (instead of
   raising AccessDenied).
+- 1679_: [Windows] net_connections() and Process.connections() are twice as
+  fast.
 
 **Bug fixes**
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,8 +16,7 @@ XXXX-XX-XX
 - 1671_: [FreeBSD] add CI testing/service for FreeBSD (Cirrus CI).
 - 1677_: [Windows] process exe() will succeed for all process PIDs (instead of
   raising AccessDenied).
-- 1679_: [Windows] net_connections() and Process.connections() are twice as
-  fast.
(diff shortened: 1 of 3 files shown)


giampaolo:
@mlabonte I am benchmarking this PR with:

```python
import psutil, socket, time

ls = []
for x in range(5000):
    s = socket.socket()
    s.bind(('127.0.0.1', 0))
    s.listen(4)
    ls.append(s)

t = time.time()
for x in range(100):
    psutil.net_connections('tcp')
print(time.time() - t)
[x.close() for x in ls]
```
The results I get are worse than #1676. Without patch I get `3.35` secs, with patch `3.00`, which is around 9%.

mlabonte:
I cannot understand, I did the same on my workspace and 1679 is slighly faster than 1676. I heaven found one case where it's definively faster but still, I have no idea why as both PR do the same operations.

(pr1679) $ python -m timeit "import psutil; psutil.Process().connections(kind='tcp')"
1000 loops, best of 3: 439 usec per loop

(pr1676) $ python -m timeit "import psutil; psutil.Process().connections(kind='tcp')"
1000 loops, best of 3: 636 usec per loop

giampaolo:
Can you get similar results as https://github.com/giampaolo/psutil/pull/1676#issuecomment-580909687? This one:

```
Worsth case (end of a quick rampup from 0 K to 15 K connections):
$ python -m timeit -n 1 -r 1 "import psutil; psutil.Process().connections(kind='tcp4')"
1 loops, best of 1: 25.5 sec per loop <--- psutil==5.6.7
1 loops, best of 1: 259 msec per loop <--- pull request
```

Looks too much of a speedup. Are you sure you were measuring the right thing?

mlabonte:
Yes, I get similar results comparing master to 1679. It's a particular case: from approx. 9 K conxs, if the rate of new connections is big enough, the size of the table was always bigger than the previous call so we were stuck in the loop. I could get a single call blocked for minutes with ramping up to 60 K connections.

giampaolo:
Whatever measure I try I always get around 10% speedup. Anyway, it's fine. Merging.

mlabonte:
Thanks for merging this! :-) Do you have an ETA to release it in pypi?

PR #1631, opened 2019-11-29
[WIP] [Linux] Connections in separate network namespaces are invisible

davereikher:
Please see #1611 for explanation of the problem and related discussion.

Issue #1611, opened 2019-10-24
[Linux] Flag to get connections from all network namespaces


davereikher:
Currently the `psutil.net_connections()` function on Linux returns only connections found in the same network namespace as the Python process using psutil. That is because in the `psutil.Connections.process_inet()` method, psutil reads from `/proc/net`, which is a symlink to `/proc/[PID]/net` where `[PID]` is the pid of the python process.
According to the [network namespaces man page](http://man7.org/linux/man-pages/man7/network_namespaces.7.html):

> Network namespaces provide isolation of the system resources
>        associated with networking: network devices, IPv4 and IPv6 protocol
>        stacks, IP routing tables, firewall rules, the /proc/net directory
>        (which is a symbolic link to /proc/PID/net), the /sys/class/net
>        directory, various files under /proc/sys/net, port numbers (sockets),
>        and so on.  In addition, network namespaces isolate the UNIX domain
>        abstract socket namespace (see unix(7)).

I propose to add a boolean keyword argument `all_namespaces` to `net_connections()` to allow the user to choose to retrieve connections from the same network namespace or from all network spaces. If people agree, I am happy to submit a pull request :)

giampaolo:
I'm not sure I understand what this is for (first time I hear about network namespaces). What's the use case? What the PR would look like? In general, I'd say I'm against an argument which makes sense on 1 platform only.

davereikher:
I stumbled upon this issue while contributing to the [Rust clone](https://github.com/borntyping/rust-psutil) of psutil. I was working on adding a feature for listing the connections per process, which is currently missing there and for debugging I used the output of psutil for comparison. I noticed that the rust code listed pids and associated connections which were missing from the psutil output. I was also unaware of network namespaces back then, so I did some digging.
Network namespaces is a means in Linux to give a different view of the network stack to different processes. You could, for example, run two different processes with different routing tables, network interfaces, etc. Currently psutil for Linux lists only connections of processes running in the same network namespace as the Python process, for the reason stated in my above post. To show this, you can make the following experiment:

[...]

Issue #1581, opened 2019-09-10
[Linux] speed up net connections with NETLINK / DIAG sockets


giampaolo:
This would make `net_connections()` and `Process.connections()` on Linux considerably faster as we would avoid parsing `/proc/net/*`. See:
https://kristrev.github.io/2013/07/26/passive-monitoring-of-sockets-on-linux
https://github.com/kristrev/inet-diag-example/
This is the same technique used by `ss` command.
A quick benchmark shows this can be up to 2.5 / 3.0 times faster than the current implementation (see [netstat.py](https://github.com/giampaolo/psutil/blob/master/scripts/netstat.py)):

```
$ time python3 scripts/netstat.py
...
real	0m0,120s
user	0m0,086s
sys	0m0,033s

$ time ss
...
real	0m0,043s
user	0m0,004s
sys	0m0,028s
```

ImUrX:
I'm tempted to try doing this if you haven't yet tried

PR #1691, opened 2020-02-13
[Linux] Connections in separate network namespaces are invisible

davereikher:
This is a continuation of PR #1631 which was accidentally closed.
Please see #1611 for explanation of the problem and related discussion.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -868,6 +868,25 @@ class Connections:
         return _common.addr(ip, port)
 
     @staticmethod
+    def network_namespaces():
+        net_namespaces = defaultdict(list)
+        for pid in pids():
+            try:
+                ns = readlink("%s/%s/ns/net" % (get_procfs_path(), pid))
+            except (FileNotFoundError, ProcessLookupError):
+                continue
+            except OSError as err:
+                if err.errno == errno.EINVAL:
+                    # not a link
+                    continue
+                raise
+            else:
+                if ns.startswith('net:['):
+                    ns = ns[5:][:-1]
+                    net_namespaces[ns].append(pid)
+        return net_namespaces
+
+    @staticmethod
     def process_inet(file, family, type_, inodes, filter_pid=None):
         """Parse /proc/net/tcp* and /proc/net/udp* files."""
         if file.endswith('6') and not os.path.exists(file):
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -72,7 +72,7 @@ __extra__all__ = [
 # --- globals
 # =====================================================================
 
-
+# TODO remove this comment
 POWER_SUPPLY_PATH = "/sys/class/power_supply"
 HAS_SMAPS = os.path.exists('/proc/%s/smaps' % os.getpid())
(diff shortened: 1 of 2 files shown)


davereikher:
@giampaolo Is there a problem with AppVeyor builds? The errors there seem to be unrelated to my modifications.
If so, then this PR is ready for code review.

psutil/_pslinux.py
```diff
@@ -866,44 +866,74 @@ def decode_address(addr, family):
...
-                    #                      "PIDs references")
-                    pid, fd = inodes[inode][0]
-                else:
-                    pid, fd = None, -1
-                if filter_pid is not None and filter_pid != pid:
+    def network_namespaces():
+        net_namespaces = defaultdict(list)
+        for pid in pids():
+            try:
+                ns = readlink("%s/%s/ns/net" % (get_procfs_path(), pid))
```
giampaolo: use `self._procfs_path`

psutil/_pslinux.py
```diff
@@ -866,44 +866,74 @@ def decode_address(addr, family):
...
-                    else:
-                        status = _common.CONN_NONE
+                raise
+            else:
+                if ns.startswith('net:['):
+                    ns = ns[5:][:-1]
+                    net_namespaces[ns].append(pid)
+        return net_namespaces
+
+    def process_inet(self, file_name, family, type_, inodes, filter_pid=None):
```
giampaolo: This was a `staticmethod`

psutil/_pslinux.py
```diff
@@ -72,7 +72,7 @@
 # --- globals
 # =====================================================================
 
-
+# TODO remove this comment
```
giampaolo: ?

psutil/_pslinux.py
```diff
@@ -866,44 +866,74 @@ def decode_address(addr, family):
...
+                raise
+            else:
+                if ns.startswith('net:['):
+                    ns = ns[5:][:-1]
+                    net_namespaces[ns].append(pid)
+        return net_namespaces
+
+    def process_inet(self, file_name, family, type_, inodes, filter_pid=None):
+        """Parse /proc/*/net/tcp* and /proc/*/net/udp* files."""
+        net_ns = Connections.network_namespaces()
```
giampaolo: you can avoid calling this if `filter_pid is None`

giampaolo (commented):
I started looking into this PR. Here's some considerations.
This patch assumes that `/proc/{PID}/ns/net` (aka network namespaces) always exists. I suppose there may be circumstances (e.g. old Linux kernels) where this file may not exist, so I think the code should take that into account and fallback on using the old method (read /proc/net/tcp|udp).

Also, I suggest a refactoring: leave `process_inet()` and `process_unix()` methods alone. Instead of getting the namespace files in there, do that in `retrieve()` method and call `process_inet()` / `process_unix()` for each file.

Finally, but I don't think this is a problem for this PR, because we may implement #1581 sooner or later, the Python implementation will be altered. I cannot foresee how exactly that'll happen though, so ignore this part for now.

davereikher:
Thanks, I've addressed the review in ec3760b

PR #1693, opened 2020-02-13
[Windows] increase float precision of time APIs (boot_time(), users(), proc create_time()

giampaolo:
Right now both values are expressed in seconds and have a 1 second precision. 

```python
>>> import psutil
1581522914.0
>>> psuti.boot_time()
1581522614.0
```

With this PR we should get up to nano seconds precision (provided by 2 native system APIs):

```python
>>> import psutil
1581522914.3436192
>>> psuti.boot_time()
1581522614.02836111
```


giampaolo:
It turns out also `users()` login time can benefit of the better precision. Done in c67ab01.

Issue #1681, opened 2020-02-07
[Arch Linux x86_64] psutil.disk_partitions does not show swap


frankenstein91:
**Platform**
* OS: Arch Linux x86_64
* Kernel Linux 5.5.2-arch1-1 #1 SMP PREEMPT Tue, 04 Feb 2020 18:56:18 +0000 x86_64 GNU/Linux
* pip show psutil
Name: psutil
Version: 5.6.7

**Bug description**
As I understand it, the psutil.disk_partitions function should display all partitions. But my SWAP partitions are not listed.

```
NAME    MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda       8:0    0 931,5G  0 disk
├─sda1    8:1    0     2G  0 part /boot 
├─sda2    8:2    0    20G  0 part [SWAP]
└─sda3    8:3    0 909,5G  0 part /run/btrfs-root
nvme0n1 259:0    0 465,8G  0 disk
```

giampaolo:
did you try `psutil.disk_partitions(all=True)`?

frankenstein91:
yes 
missing on both

giampaolo:
Mmm I am not sure. Partitions are retrieved via `getmntent()` syscall which parses /etc/mtab: 
https://github.com/giampaolo/psutil/blob/05758f6978aefea709150898ae285b3ff3688c7c/psutil/_psutil_linux.c#L228
I assume that means /etc/mtab does not list swap partitions (I checked on my system: it's the same).

giampaolo:
I implemented this in 00a3398. You'll have to pass `all=True` to show swap partition.

frankenstein91:
First of all, thanks @giampaolo for the change. Unfortunately I still have a thought with your implementation. As I understand the documentation, all=False should filter on hardware. So USB sticks, hard drives and SSDs, but leave out virtual devices. I don't see the SWAP memory itself as such a virtual device, like a RAM disk. In my case it is not even a SWAP file, but a real partition. So I think it would be cool if this could be adjusted a bit more. 

I hope it is possible

giampaolo:
You got me thinking here so I investigate this topic a bit. 

The problem with swap partitions/files is that they are not like other regular partitions and most importantly they are not mounted (having `disk_partitions()`'s `moutpoint` value set to `None` is a first). You only have the reference to the device, which is a block type file ("b" column in ls). On my Linux box:

```
[...]

PR #1694, opened 2020-02-14
[WIP] psutil.disk_swaps() (swap partitions)

giampaolo:
This implements #1681, by adding a new `psutil.disk_swaps()` which lists swap partitions (and swap files). 
```python
# Linux
>>> import psutil
>>> psutil.disk_swaps()
[sdiskswaps(path='/dev/nvme0n1p3', total=11718652, used=2724, fstype='partition', priority=-2)]
```
```python
# Windows
>>> import psutil
>>> psutil.disk_swaps()
[sdiskswaps(path='C:\\pagefile.sys', total=1118652, used=5721, peak=3453)]
```

Linux and Windows only for now, but at least macOS and/or FreeBSD would also be nice to have.
I'm still not 100% sure about the name.

UPDATE - added support for:

- [x] Linux
- [x] Windows
- [x] FreeBSD
- [x] OpenBSD


giampaolo:
Current name:
* `disk_swaps()`
  * +1 because it's neutral about partitions vs. files (Linux can have both, Windows has files)
  * +1 /proc/swaps suggests someone already did the bikeshedding =) 

Other possible names:

* `disk_swap_partitions()`:
  *  -1 because we can also list swap files
* `swap_disks()`: 
   * +1 because we already have the "swap_" prefix of `swap_memory()`
   * -1 because it sounds like entries refer to an entire disk
* `swap_locations()`: 
   *  +1 (reuse "swap_" prefix)

PR #1696, opened 2020-02-16
Remove use of deprecated setuptools test_suite & tests_require

jdufresne:
Since setuptools v41.5.0 (27 Oct 2019), the 'test' command is formally
deprecated and should not be used. Test are still easy to run through
tox -- as before.


giampaolo:
Thanks. It's always good to see code being removed. =)

Issue #1620, opened 2019-11-13
[linux] cpu_count_physical is not correct


sctlee:
**Platform**
* linux
* psutil 5.6.2

**Bug description**
currently, psutil use '/sys/devices/system/cpu/cpu[0-9]*/topology/core_id' to calculate physical cores, but core id is only uniq in one socket.
```
def cpu_count_physical():
    """Return the number of physical cores in the system."""
    # Method #1
    core_ids = set()
    for path in glob.glob(
            "/sys/devices/system/cpu/cpu[0-9]*/topology/core_id"):
        with open_binary(path) as f:
            core_ids.add(int(f.read()))
    result = len(core_ids)
    if result != 0:
        return result
```

**Test results**
```
[root@dce100 ~]# lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    2
座：                 2
```

```
[root@dce100 ~]# cat /proc/cpuinfo | grep -E "^(physical|core|process)"
processor	: 0
physical id	: 0
core id		: 0
processor	: 1
physical id	: 0
core id		: 1
processor	: 2
physical id	: 1
core id		: 0
processor	: 3
physical id	: 1
core id		: 1
```

```
>>> import psutil
>>> psutil.cpu_count(logical=False)
2
```

amanusk:
This is also related to #1392  and #1558, currently there is no support for getting the number of socket in the system reliably. The number of sockets can be used as a sanity check for number of physical CPUs and can be interesting by itself.
Perhaps we should take a look at how this is done in lscpu:
https://github.com/karelzak/util-linux/blob/9cca1ef2d5283a461c3a6d9aa1a036bb4ada6e70/sys-utils/lscpu.c#L1078

DannyVanpoucke:
Could replacing the "set()" with an array resolve the problem? At least on one system I noticed the core_id to only be present for the physical cores...alternately this may provide a work-around to get the number of sockets as being the double-counting factor. (of course if some systems implement a core_id file for the logical cores as well this won't work)

Keou0007:
I've noticed this has broken on my system lately.
I run dual socket systems, some with hyperthreading disabled in the bios. OS is CentOS 6.

[...]

Issue #1404, opened 2019-02-06
psutil.cpu_count(logical=False) fails to report correct number or even gives None


pseudotensor:
On ppc systems, this just gives None, even though (say) the system has 8 hyperthreading cores per physical core and (say) 160 total hyperthreading cores.

Even worse, on x86 systems, it randomly gives 64k as the result.  A reboot changes that for the systems affected, but it eventually gets stuck that way again.

Since we were feeding that into a choice of physical cores, this led to major problems for any multi-threaded application (a thread bomb).

Perhaps psutil is not reading the data correctly in all cases, or makes assumptions about the data format that can change.

pseudotensor:
ppc:
```
jon@mr-0xp1:~$ python
Python 3.6.5 (default, Jan 23 2019, 17:10:07) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> max_cores = psutil.cpu_count(logical=False)
>>> max_cores
>>> 
```

```
jon@mr-0xp1:~$ cat /proc/cpuinfo 
processor       : 0
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 1
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 2
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 3
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 4
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 5
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 6
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 7
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 8
cpu             : POWER8E (raw), altivec supported
clock           : 2061.000000MHz
[...]

PR #1727, opened 2020-04-15
Patch cpu count physical

jandrovins:
Solves #1620 

Method 1 of `cpu_count_physical()` fails on several scenarios. I tested in a node with two sockets:
```
$ cat /sys/devices/system/cpu/cpu[0-9]*/topology/core_id | sort
0
0
1
1
2
2
3
3
4
4
5
5
6
6
7
7
```
Each of these values is added to a `set`, so this function would end up returning half the physical cores. And in my PC with 2 physical cores with Hyper Threadding:
```
 $ cat /sys/devices/system/cpu/cpu[0-9]*/topology/core_id
0
1
2
3
```
This function returned twice the physical cores.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -619,12 +619,12 @@ def cpu_count_logical():
 def cpu_count_physical():
     """Return the number of physical cores in the system."""
     # Method #1
-    core_ids = set()
+    thread_siblings_lists = set()
     for path in glob.glob(
-            "/sys/devices/system/cpu/cpu[0-9]*/topology/core_id"):
+            "/sys/devices/system/cpu/cpu[0-9]*/topology/thread_siblings_list"):
         with open_binary(path) as f:
-            core_ids.add(int(f.read()))
-    result = len(core_ids)
+            thread_siblings_lists.add(f.read())
+    result = len(thread_siblings_lists)
     if result != 0:
         return result
 


Campanula:
I track the commit history of method 1, it's not a perfect solution to delete it simply.

https://github.com/giampaolo/psutil/issues/1404

"/sys/devices/system/cpu/cpu[0-9]*/topology/core_id"
"/sys/devices/system/cpu/cpu[0-9]*/topology/physical_package_id"

Maybe we should combine the two.

jandrovins:
That would solve the issue on systems with more than one socket, but not in systems with Hyper Threading, as `physical_package_id` is the same on all logical cores, but `core_id` is different.

But you are right  in that the first method solved an existing problem. I believe that `ls /sys/devices/system/node/node[0-9]*/` list the attributes of each socket, and hence find each `core_id` associated with each socket. This might be the solution, but it's late where I live, I'll probably post another possible solution this weekend. :smile:

jandrovins:
With the new changes, `cpu_count_physical` fetches `thread_siblings_lists` values for each logical cpu  (i.e. `/sys/devices/system/cpu/cpu[0-9]*/topology/thread_siblings_list`). This file contains the logical core ids with which the logical core being fetch shares the core. This ids refers to `N` in `cpuN` from the paths accessed with glob. 

Without Hyper Threading:
```
$ cat /sys/devices/system/cpu/cpu30/topology/thread_siblings_list
30
$ cat /sys/devices/system/cpu/cpu30/topology/core_id              
15
```

With Hyper Threading:
```
$ cat /sys/devices/system/cpu/cpu30/topology/thread_siblings_list
30,62
$ cat /sys/devices/system/cpu/cpu30/topology/core_id
15

```

marxin:
I see very similar problem on 2 socket system in which I have a VM:

```python3
In [2]: psutil.cpu_count()                                                                                                                                                                                                                                                      
Out[2]: 64

In [3]: psutil.cpu_count(logical=False)                                                                                                                                                                                                                                         
Out[3]: 1
```

where I see:
```
$ cat /sys/devices/system/cpu/cpu[0-9]*/topology/core_id | sort | uniq -c
     64 0
```

So I see `0` 64 times.

marxin:
And `lscpu` tells:

```
lscpu
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   42 bits physical, 48 bits virtual
CPU(s):                          64
On-line CPU(s) list:             0-63
Thread(s) per core:              1
Core(s) per socket:              1
Socket(s):                       64
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           94
Model name:                      Intel Core Processor (Skylake, IBRS)
Stepping:                        3
CPU MHz:                         2800.046
BogoMIPS:                        5600.09
Virtualization:                  VT-x
Hypervisor vendor:               KVM
Virtualization type:             full
L1d cache:                       2 MiB
L1i cache:                       2 MiB
L2 cache:                        256 MiB
L3 cache:                        1 GiB
NUMA node0 CPU(s):               0-63
Vulnerability Itlb multihit:     KVM: Vulnerable
Vulnerability L1tf:              Mitigation; PTE Inversion; VMX vulnerable, SMT disabled
Vulnerability Mds:               Vulnerable; SMT Host state unknown
Vulnerability Meltdown:          Vulnerable
Vulnerability Spec store bypass: Vulnerable
Vulnerability Spectre v1:        Vulnerable: __user pointer sanitization and usercopy barriers only; no swapgs barriers
Vulnerability Spectre v2:        Vulnerable, IBPB: disabled, STIBP: disabled
Vulnerability Tsx async abort:   Vulnerable
Flags:                           fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology cpuid tsc_known_freq pni pclmulqdq vmx ssse3 fma cx16 pcid sse4_1 sse4_2
                                  x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch cpuid_fault invpcid_single ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2
                                  erms invpcid rtm rdseed adx smap xsaveopt arat umip md_clear arch_capabilities
```

ml-evs:
Thanks @jandrovins for this PR!

I've just tested it out on our dual-socket CPU E5-2660 v3 machine on Linux. It looks like the number of cores is now correctly reported, taking hyperthreading into account. Though these CPUs support hyperthreading (`flag: ht` present in `/proc/cpuinfo`), I believe we have it disabled (see `lscpu` output at bottom "Thread per core"). I guess it would be useful to test this on a multi-socket machine that has HT enabled?

```bash
$ uname -a
Linux node14 3.11.10-25-default #1 SMP Wed Dec 17 17:57:03 UTC 2014 (8210f77) x86_64 x86_64 x86_64 GNU/Linux
$ cat /proc/cpuinfo | grep processor | tail -n 1
processor       : 19
```

Old behaviour:

```python
>>> import psutil
>>> psutil.cpu_count()  # correct, by accident? i.e. guessing that HT is turned on, and so doubling the core count for single socket
20
>>> psutil.cpu_count(logical=False)
10
>>> psutil._pslinux.cpu_count_physical()
10
```

With this PR, everything seems correct:

```python
>>> import psutil
>>> psutil.cpu_count()
20
>>> psutil.cpu_count(logical=False)
20
>>> psutil._pslinux.cpu_count_physical()
20
```

`lscpu` output:
```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                20
On-line CPU(s) list:   0-19
Thread(s) per core:    1
Core(s) per socket:    10
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 63
Model name:            Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz
...
```

Keou0007:
I have also tested the new method 1 on dual socket machines both with and without HT. It returns correct values under CentOS 6 and 7.

dgold9670:
This looks to be ready to merge. Am I missing a reason it hasn't been?

ewilts:
@jandrovins will this patch also fix issue 1808 that I reported yesterday?  
https://github.com/giampaolo/psutil/issues/1808
My issue isn't hyper-threading related but simply a dual-socket 8-cores per socket system returning 15 cores.  The symptoms above don't appear to address an "off-by-1" issue.

ml-evs:
> @jandrovins will this patch also fix issue 1808 that I reported yesterday?
> #1808
> My issue isn't hyper-threading related but simply a dual-socket 8-cores per socket system returning 15 cores. The symptoms above don't appear to address an "off-by-1" issue.

Could you try it @ewilts? 

I've just tested on a dual-socket Xeon Gold 6244 (8 cores, 16 threads per socket) machine, and the current psutil release gives 12 (?!) and 32 for `psutil.cpu_count(logical={False, True})`. This looks similar to the machine you tested on in #1808.

Installing from this patch `pip install git+http://github.com/jandrovins/psutil@patch_cpu_count_physical` gives the correct 16/32.

PedanticHacker:
If you wanna know whether the processor running your code is utilizing hyperthreading and to then act accordingly, you can define your own function.

```
import psutil

def hyperthreading_enabled():
    return psutil.cpu_count(logical=True) > psutil.cpu_count(logical=False)

if hyperthreading_enabled():
    # Act on if hyperthreading is enabled!
else:
    # Act on if hyperthreading is disabled!
```

ml-evs:
> If you wanna know whether the processor running your code is utilizing hyperthreading and to then act accordingly, you can define your own function.
> 
> ```
> import psutil
> 
> def hyperthreading_enabled():
>     return psutil.cpu_count(logical=True) > psutil.cpu_count(logical=False)
> 
> if hyperthreading_enabled():
>     # Act on whether hyperthreading is enabled!
> else:
>     # Act on whether hyperthreading is disabled!
> ```

This doesn't work for multi-socket CPUs (as discussed above). If you have hyperthreading disabled and 2 sockets (a common configuration) then you get different answers, even with HT disabled.

PedanticHacker:
> This doesn't work for multi-socket CPUs (as discussed above). If you have hyperthreading disabled and 2 sockets (a common configuration) then you get different answers, even with HT disabled.

There are not a lot of people in these days who have more than 1 physical CPU on their motherboard. That was once a thing, but now it isn't anymore (since the multicore CPUs came into market). And even if one has a multisocket CPU, one must have Windows 10 Pro (the Home edition supports only single-socket CPUs).

I know that we want to detect every possible scenario, but if one has Windows 10 Home, then you can be sure that only 1 physical CPU is being used (albeit one has 2 or more physical CPUs).

I am googling around for a way to detect whether a multisocket CPU is utilized by the system. Haven't found anything useful yet.

ewilts:
>     There are not a lot of people in these days who have more than 1 physical CPU on their motherboard

For home computers, you're probably right.  For data centers, this is absolutely false.  

> And even if one has a multisocket CPU, one must have Windows 10 Pro 

Or Linux, which is where I tripped over this bug and where most of this thread has been focused.

ml-evs:
> I am googling around for a way to detect whether a multisocket CPU is utilized by the system. Haven't found anything useful yet.

This is precisely one of the issues that this PR solves.

PedanticHacker:
I wonder how this application counts CPU sockets.

PedanticHacker:
When is the new version of `psutil` gonna be available that fixes this?

giampaolo:
@all-involved: I trust this PR fixes the problem which unfortunately I cannot reproduce or investigate in detail due to lack of time. Before merging though, I would like to make sure that also "method 2" is not broken:
https://github.com/giampaolo/psutil/blob/ba0c0ab967b0f524807cedf06554b94b7aba3d70/psutil/_pslinux.py#L630-L651
"method 2" is used if `/sys/devices/system/cpu/cpu[0-9]*/topology/thread_siblings_list` files are not available. What happens if "method 1" is removed/commented-out? Does it give the expected results? Can somebody try?

Keou0007:
tested on CentOS 6 and 7, both method 1 and 2 independently return the correct number of physical cores across two sockets.

PedanticHacker:
Let me get this straight. So does the `cpu_count_physical()` return the number of `physical CPU cores` or the `physical CPU sockets`? The name is a little bit misleading.

Also, the name of the function `cpu_count()` suggests to me that the function counts how many physical CPUs a system has. I think it'd be better if the name was `thread_count()` for counting CPU threads, and `core_count()` for counting CPU cores, and then `cpu_count()` should count physical CPUs in a system, so CPU sockets.

Do you find any value in this?

dgold9670:
Using cat /proc/cpuinfo and lscpu as standins, it looks like physical_id is physical socket, core id is specific core number, and cpu cores the number of physical (non-hyperthreaded) cores per physical socket.
Output:
`bash-4.2# lscpu |grep -E 'CPU|Thread|Core|Socket'
CPU op-mode(s):        32-bit, 64-bit
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             2
CPU family:            6
Model name:            Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
CPU MHz:               3011.328
CPU max MHz:           3300.0000
CPU min MHz:           1200.0000
NUMA node0 CPU(s):     0-7,16-23
NUMA node1 CPU(s):     8-15,24-31
bash-4.2# cat /proc/cpuinfo |grep core|sort -u
core id         : 0
core id         : 1
core id         : 2
core id         : 3
core id         : 4
core id         : 5
core id         : 6
core id         : 7
cpu cores       : 8
bash-4.2# cat /proc/cpuinfo |grep physical|sort -u
physical id     : 0
physical id     : 1
`

giampaolo:
> Let me get this straight. So does the cpu_count_physical() return the number of physical CPU cores or the physical CPU sockets? The name is a little bit misleading.

It's supposed to return physical CPU cores. And yes, it's a bit misleading. To summarize:

- `cpu_count(logical=True)` = number of logical **usable** CPUs (including **hyper threading** ones)
- `cpu_count(logical=False)` = number of CPU **cores**

The number of CPU **sockets** is currently not implemented. We can debate whether to add it but:
- is there a use-case for it?
- I'm not sure how to expose that in terms of API; I wish `cpu_count(logical=bool)` was `cpu_count(type="string")`. That would have been good also because somebody requested the number of CPU NUMA nodes, but that boat has sailed long ago.

amanusk:
I agree that the information on number of sockets is missing, and would be intersted in having this available .
Some possible suggestions I would like to propose for not breaking APIs:
* A new call for `socket_count`, which will help distinguish CPU which usually refers to core/thread in psutil, and socket
* An optional second parameters to `cpu_count` with the index of the socket you wish to get information on (if possible). By default, it will return the sum of all sockets, buy on a 2 sockets system, `cpu_c

amanusk:
I agree this functionality is missing in psutil, and would be interested in having it.
I'd like to propose some suggestions for a non breaking API:
* A new call for `socket_count` for getting the number of socket. This might help distinguish CPU which usually means core/thread in psutil and an actual separate socket
* An optional second parameter for `cpu_count` which receives the index of the socket you wish to get information for. By default, the call will return the summary of both, but e.g. the call `cpu_count(logial=false, index=0)` would return only the number of physical cores on socket 0 in a multi socket system

zklaus:
Note that `thread_siblings_list` is deprecated (see [here](https://www.kernel.org/doc/Documentation/admin-guide/cputopology.rst)).
The new name is apparently `core_cpus_list`. I personally also had success with the following code that is based only on all `*_id` files.

```
def cpu_count_physical():
    # Adapted from psutil
    """Return the number of physical cores in the system."""
    IDS = ["physical_package_id",
           "die_id",
           "core_id",
           "book_id",
           "drawer_id"]
    # Method #1
    core_ids = set()
    for path in glob.glob(
            "/sys/devices/system/cpu/cpu[0-9]*/topology"):
        core_id = []
        for id in IDS:
            id_path = os.path.join(path, id)
            if os.path.exists(id_path):
                with open(id_path) as f:
                    core_id.append(int(f.read()))
        core_ids.add(tuple(core_id))
    result = len(core_ids)
    if result != 0:
        return result
    else:
        return None
```

giampaolo:
Merged. I also committed bfae1fc4a371c9e08f2c3f5053a80542e43d18f7 which takes into account that `*/thread_siblings_list` is deprcated and may not be available in the future (`*/core_cpus_list` is attempted first).
I also added a test which checks that "method 1" and "method 2" return the same value.

Issue #1737, opened 2020-04-28
[windows] subprocesses and signals


dothebart:
If one spawns several processes, using signals with them isn't directly possible. 

Please either directly add `kwargs['creationflags'] - subprocess.CREATE_NEW_PROCESS_GROUP` to the code, or indicate in the documentation, that one should do this. 
Else signalling subprocesses is only possible for *all* subprocesses at once. 

https://stefan.sofa-rockers.org/2013/08/15/handling-sub-process-hierarchies-python-linux-os-x/

giampaolo:
You mean in the test suite? `get_test_subprocess()` function?

dothebart:
it seems that this combination does the right thing; Launch the process like this:
```
ON_WINDOWS = (sys.platform == 'win32')
...
kwargs = {}
if ON_WINDOWS:
       kwargs['creationflags'] = subprocess.CREATE_NEW_PROCESS_GROUP
self.proc = subprocess.Popen(self.arguments, **kwargs)
```
and then in terminate instead of redirecting to kill dodge the self signal by temporarily setting a signal handler: 
```
def dodge_sigint_handler(signum, frame):
    if signum != signal.SIGINT:
        exit(1)
    pass

original_sigint_handler = signal.getsignal(signal.SIGINT)
signal.signal(signal.SIGINT, dodge_sigint_handler)
self.proc.send_signal(signal.CTRL_BREAK_EVENT)
self.proc.wait()
# restore original handler
signal.signal(signal.SIGINT, original_sigint_handler)
```

This will result in terminating *only* the intended process, as would sending SIG_TERM on linux.

giampaolo:
In the test suite it's not too important to kill a specific process only 'cause on `tearDown` we simply want to terminate them all, so I don't think it's worth it to add complexity. 
Signal handling is complex even without putting Windows in the mix: =)
http://grodola.blogspot.com/2016/02/how-to-always-execute-exit-functions-in-py.html

dothebart:
Please have a look at my PR. 
I think its not as complicated, as one may think ;-P

dothebart:
Please. We want to have a portable way. 
psutils should behave as similar on all available targets as possible. 
This patch will make it behave on the wintendo more like on a real OS.

giampaolo:
Closing after discussion in https://github.com/giampaolo/psutil/pull/1740.

ObiWahn:
[...]

PR #1740, opened 2020-04-29
fix invokation of windows processes, so we can terminate() them 

dothebart:
one by one without shooting ourselves.

this takes care of:
https://github.com/giampaolo/psutil/issues/1737

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -1240,12 +1240,23 @@ class Process(object):
     def terminate(self):
         """Terminate the process with SIGTERM pre-emptively checking
         whether PID has been reused.
-        On Windows this is an alias for kill().
+        On Windows this will only work for processes spawned through psutil
+        or started using
+           kwargs['creationflags'] = subprocess.CREATE_NEW_PROCESS_GROUP.
         """
         if POSIX:
             self._send_signal(signal.SIGTERM)
         else:  # pragma: no cover
-            self._proc.kill()
+            def sigint_boomerang_handler(signum, frame):
+                if signum != signal.SIGINT:
+                    exit(1)
+                pass
+            original_sigint_handler = signal.getsignal(signal.SIGINT)
+            signal.signal(signal.SIGINT, sigint_boomerang_handler)
+            self.send_signal(signal.CTRL_BREAK_EVENT)
+            self.wait()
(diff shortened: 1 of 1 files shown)


giampaolo:
I thought you were talking about the test suite. Here you're touching psutil itself. The point of `terminate()` method is being able to terminate *any* process, not only the ones created with the CREATE_NEW_PROCESS_GROUP flag. If that is your goal you should do it in your own code, registering your own signal handler. Doing so in a general purpose lib is not a good idea.

dothebart:
it will continue to be able to terminate *any* process. However, if you *SPAWN* them without, it will continue to kill all processes you spawned. 

In the current implementation state psutils will simply allways *hard-kill* processes on windows *allways* which to that regard is way worse!

giampaolo:
`psutil.Process` class can be used with any PID. In this PR you are registering a signal handler and using `exit()`. Both of those things affect the current process (`os.getpid()`), not the PID/process the `Process` instance is referring to.

dothebart:
since its only registered for SIGINT this case shouldn't happen? Whats would be a better way to do this?

giampaolo:
You mean avoid killing the entire group of process/PID you did NOT spawn? I don't know. But it seems to me your use case is about the processes created by yourself (os.getpid()), so you already have a solution. You have all the bits already, so just use it in your own code. =)

To fully recap, changing `Process.terminate()` the way you propose is a no-go because:
- it does not apply to any PID, just os.getpid() 
- it register a signal handler
- it would change the semantic of `Process.terminate()`, which should be the same as `subprocess.Popen.terminate()`, which is calling `TerminateProcess` Windows API

With that said, I agree that having something more high level (like a parameter to pass to `terminate()` or something) to avoid killing the entire process group (of any PID), would be nice to have, but as of now I can't say I understand the problem well enough to have a solid solution.

dothebart:
the problem is, with whoever is spawning processes without using CREATE_NEW_PROCESS_GROUP will have a sigint sent to all processes in that group - be they parented from you or not. 
Thus you cannot soft-kill (signal to terminate with a regular shutdown) single processes of a group alone. 

Ok, I will have to fork psutils then and stop taking the work to do PRs.

Issue #693, opened 2015-09-30
psutil compatibility on Windows Nano Server


alexpilotti:
psutil does not work on Nano Server, failing to load due to a missing DLL (wtsapi32) and some PSAPI exports:
- Usage of wtsapi32 API. Nano Server does not have any GUI related features, so Remote Desktop APIs are not included. This means that psutil_users needs to be rewritten using different APIs or dinamically load the function addresses (GetProcAddress) when not running on Nano.
- Missing PSAPI_VERSION 1 functions. On the currently available TP3 release, only PSAPI_VERSION 2 APIs are included. This is most probably going to be fixed in the Nano forwarders in the upcoming releases, so hopefully no changes will be required in psutil. 

Here's a quick patch that adds Nano Server TP3 compatibility by simply removing the psutil_users implementation and setting PSAPI_VERSION = 2. This is just meant as a PoC as psutil_users will need to be fleshed out before a PR:

https://github.com/cloudbase/psutil/commit/32053d8c5bdcde22c7b23b7558e9386d06794a09

Thanks!

giampaolo:
Wow! This is old and totally slipped under my radar, sorry. I'm not sure if you're still interested into this, but if your patch doesn't break other Windows versions I'm OK with merging it.

giampaolo:
PS: next time please make a PR. They are more "noticeable".

giampaolo:
Useful link: https://msdn.microsoft.com/en-us/library/aa383056(v=vs.85).aspx

giampaolo:
I cannot reproduce this due to lack of hardware, but I can try to change the C function so that it dynamically loads the necessary functions. Would you be able to test this if I do?

alexpilotti:
Sure, I can definitely test it!

smile21prc:
Hi @giampaolo, I also needs this fix -- has it been merged already? I still see error below with psyutil 5.2.1 on Nano.

...
    import psutil
  File "C:\Python\lib\site-packages\psutil\__init__.py", line 126, in <module>
    from . import _pswindows as _psplatform
  File "C:\Python\lib\site-packages\psutil\_pswindows.py", line 16, in <module>
    from . import _psutil_windows as cext
ImportError: DLL load failed: The specified module could not be found.

giampaolo:
[...]

PR #1768, opened 2020-05-25
Add support for Windows Nano Server

julien-lebot:
# Purpose
The purpose of this PR is to address #693 so that psutil can be used with Windows Nano Server containers.

# Additional notes

## ConnectTime
I changed the call that obtains the `ConnectTime` to use`wtsapi` since `WinStationQueryInformationW` seems deprecated (see [Microsoft documentation]( https://docs.microsoft.com/en-us/previous-versions/aa383827(v%3Dvs.85))).

## Dynamic module loading
The API from `wtsapi32.lib` is loaded dynamically as recommended [here](https://docs.microsoft.com/en-us/windows/win32/termserv/run-time-linking-to-wtsapi32-dll)

# Testing
I wrote some Dockerfile that will build a custom server nano image with python and then copy psutil into that container and ran a few commands:

## Nano Server Dockerfile
```
ARG WINDOWS_VERSION=1809
ARG PYTHON_VERSION=3.8.3
FROM python:$PYTHON_VERSION-windowsservercore-1809 as base
FROM mcr.microsoft.com/windows/nanoserver:$WINDOWS_VERSION

COPY --from=base ["Python", "Python"]

USER ContainerAdministrator
RUN setx /M PATH "%PATH%;c:\Python\;c:\Python\scripts\;"
USER ContainerUser

RUN mkdir c:\\Python\\lib\\site-packages\\psutil
COPY psutil c:\\Python\\lib\\site-packages\\psutil
ENTRYPOINT [ "python" ]
```
## Server Core Dockerfile
```
ARG WINDOWS_VERSION=1809
ARG PYTHON_VERSION=3.8.3
FROM python:$PYTHON_VERSION-windowsservercore-$WINDOWS_VERSION

RUN mkdir c:\\Python\\lib\\site-packages\\psutil
COPY psutil c:\\Python\\lib\\site-packages\\psutil
ENTRYPOINT [ "python" ]
```

### Test code

In `winmake.py`

```
def _test_windows_docker(tag, dockerfile, windows_ver = "1803", python_ver = "3.8.3"):
    docker_build = subprocess.run("docker build --isolation=hyperv --build-arg WINDOWS_VERSION=%s --build-arg PYTHON_VERSION=%s --no-cache -t %s -f %s ." % (windows_ver, python_ver, tag, dockerfile))
    if docker_build.returncode != 0:
        win_colorprint("Failed to build psutil docker image", RED)
        return
    result = subprocess.run("docker run --isolation=hyperv --rm %s -c \"import psutil\nprint(psutil.users())\"" % tag)
    print(result)


def test_docker():
    build()
    nano_server = "Dockerfiles\\windows\\Dockerfile.nanoserver"
    server_core = "Dockerfiles\\windows\\Dockerfile.servercore"
    test_matrix = [
        {
            "tag": "psutil:nanoserver-1809",
            "dockerfile": nano_server,
            "version": '1809'
        },
        {
            "tag": "psutil:servercore-1809",
            "dockerfile": server_core,
            "version": '1809'
        },
        {
            "tag": "psutil:servercore-ltsc2016",
            "dockerfile": server_core,
            "version": 'ltsc2016'
        }
    ]
    for test in test_matrix:
        print(test)
        _test_windows_docker(test["tag"], test["dockerfile"], test["version"])
```


psutil/arch/windows/ntextapi.h
```diff
@@ -551,6 +552,26 @@ DWORD (CALLBACK *_GetActiveProcessorCount) (
...
+    DWORD Reserved,
+    DWORD Version,
+    PWTS_SESSION_INFO* ppSessionInfo,
+    DWORD* pCount
+    );
+
+VOID(CALLBACK *_WTSFreeMemory)(
+    IN PVOID pMemory
+    );
+
```
giampaolo, line 574: Since these names are available on other Windows systems, I think there should be something like:

```c
#ifdef WINDOWS_NANO_OR_SOMETHING
...
#endif 
```

There's no need of the `_` prefix. Look at how other declarations are done. E.g.:

```c
PSTR (NTAPI * _RtlIpv6AddressToStringA) (
    struct in6_addr *Addr,
    PSTR P);

#define RtlIpv6AddressToStringA _RtlIpv6AddressToStringA
```

psutil/_psutil_common.c
```diff
@@ -320,6 +316,20 @@ psutil_loadlibs() {
     // minumum requirement: Win 7
     GetLogicalProcessorInformationEx = psutil_GetProcAddressFromLib(
         "kernel32", "GetLogicalProcessorInformationEx");
+    // minimum requirements: Windows Server Core
+#ifdef UNICODE
```
giampaolo, line 320: Mmm. Is this necessary? Is it possible for a system to not have unicode support?
Regardless from that, we already use other unicode "W" functions in many other places. If this is broken then they are all broken (in which case it should be a separate fix).

giampaolo (changes requested):

psutil/_psutil_common.c
julien-lebot: I wasn't sure if `psutil` was ever compiled with UNICODE unset, so I provided both "A" and "W". I will simplify this to just use the unicode methods.

psutil/arch/windows/ntextapi.h
julien-lebot: I'm not entirely sure I understand what you mean with the names being available on other Windows systems.

I think the function names are still available when compiling for Windows Nano, but the library is not present on the target system (i.e. Windows Nano is not a separate compilation target with a separate SDK).

If we're worried about name conflicts, what about using a distinctive name for the functions, like so:
```C#
BOOL(CALLBACK *psutil_WTSQuerySessionInformation) (
    HANDLE hServer,
    DWORD SessionId,
    WTS_INFO_CLASS WTSInfoClass,
    LPTSTR* ppBuffer,
    DWORD* pBytesReturned
    );

BOOL(CALLBACK *psutil_WTSEnumerateSessions)(
    HANDLE hServer,
    DWORD Reserved,
    DWORD Version,
    PWTS_SESSION_INFO* ppSessionInfo,
    DWORD* pCount
    );

VOID(CALLBACK *psutil_WTSFreeMemory)(
    IN PVOID pMemory
    );
...
    psutil_WTSEnumerateSessions = psutil_GetProcAddressFromLib(
        "wtsapi32.dll", "WTSEnumerateSessionsW");
    psutil_WTSQuerySessionInformation = psutil_GetProcAddressFromLib(
        "wtsapi32.dll", "WTSQuerySessionInformationW");
    psutil_WTSFreeMemory = psutil_GetProcAddressFromLib(
        "wtsapi32.dll", "WTSFreeMemory");
```
This way we also removes the need for the `_` prefix.

giampaolo: > I'm not entirely sure I understand what you mean

It's OK, let me try to rephrase it differently.
What I am aiming at is using the original API names, without the `_` or `psutil_` prefix (the original version of this file did exactly that, then I changed it). 
In order to do that you first have to "declare" the API name, like this:
https://github.com/giampaolo/psutil/blob/96cc7ea40d6f4f08e86677434213cef119cf1748/psutil/arch/windows/ntextapi.h#L501-L507
Then you'll have to "load" the API, in here, like this:
https://github.com/giampaolo/psutil/blob/96cc7ea40d6f4f08e86677434213cef119cf1748/psutil/_psutil_common.c#L217-L220
Is it more clear?

giampaolo: Not tested, but it should be something like this:

```c
// ntextapi.h
BOOL (CALLBACK *_WTSQuerySessionInformation) (
    HANDLE hServer,
    DWORD SessionId,
    WTS_INFO_CLASS WTSInfoClass,
    LPTSTR* ppBuffer,
    DWORD* pBytesReturned
    );

#define WTSQuerySessionInformation _WTSQuerySessionInformation
```

```c
// _psutil_common.c
WTSEnumerateSessions = psutil_GetProcAddressFromLib(
    "wtsapi32.dll", "WTSEnumerateSessions");
if (! WTSEnumerateSessions)
    return 1;
```

giampaolo: Also, MS doc says:
> The wtsapi32.h header defines WTSEnumerateSessions as an alias which automatically selects the ANSI or Unicode version of this function based on the definition of the UNICODE preprocessor constant. 

...so I think you should keep using `WTSEnumerateSessions` instead of `WTSEnumerateSessionsW`.

julien-lebot: Thanks for the explanation, I think I see what you're aiming for.
I included the header `Wtsapi32.h` in `ntextapi.h` to use the existing structures definition, but the point of `ntextapi.h` is to be independent of such header, right ?
So I should remove that `#include` and copy over all the definitions required to make it look like we are using the original header, correct ?

giampaolo: If you can fix the issue just by adding that new header (or from setup.py) there is no need to redefine anything in ntextapi.h.

Just define them explicitly in ntextapi.h only if you can’t solve it otherwise, in which case you should put the definition inside an “#ifdef nanoserver” because I think it will break other windows versions (but the CI tests should break)

julien-lebot: AFAIK there is no way to distinguish Windows Nano via preprocessor definitions - I think it is still using the same `_WIN32_WINNT` and `WINVER` as a regular Windows.
If we introduce a new prepocessor flag for Windows Nano, then we would have to maintain 2 versions of psutil,  one linked with `WtsApi32.lib` and one that loads dynamically the DLL at runtime.
Why not just keep the dynamically loaded version for all version of Windows and avoid that issue altogether ?
You mention it will break other Windows version, do you have something specific in mind ?
To me it seems the dynamic loading will also work on other Windows version (I ran my tests on many versions of Windows actually - I should probably update the PR description).

psutil/_psutil_windows.c
```diff
@@ -1210,6 +1209,15 @@ psutil_users(PyObject *self, PyObject *args) {
...
         return NULL;
 
+    // If either or those APIs are missing, we are probably running
+    // in a Windows Nano Server container
+    if (WTSEnumerateSessions == NULL ||
+        WTSQuerySessionInformation == NULL ||
+        WTSFreeMemory == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "wtsapi32.dll cannot be found.");
+        goto error;
+    }
```
giampaolo, line 1219: I'm not sure I understand. Are these names available on Windows Nano? Does this function (`users()`) work at all?

julien-lebot: No, those names are not available on Windows Nano, and consequently `users` does not work. If you try to run `psutil.users` in Windows Nano, that is what you get:
```
OSError: [WinError 120] This function is not supported on this system: '(originated from WTSEnumerateSessions)' 
```

julien-lebot: But by loading those symbols dynamically, at least `psutil` can be used on Windows Nano. Previously it would simply try to load `WtsApi32.dll` and fail even if you never called `users()`.

giampaolo: If Windows Nano does not support users at all then I think it makes more sense to just return an empty list. It's less disruptive than an error.

giampaolo: I think `Py_BuildValue("[]");` should do it.

psutil/_psutil_windows.c
```diff
@@ -1223,13 +1231,16 @@ psutil_users(PyObject *self, PyObject *args) {
...
+            WTSFreeMemory(buffer_info);
 
         buffer_user = NULL;
         buffer_addr = NULL;
+        buffer_info = NULL;
 
         // username
         bytes = 0;
-        if (WTSQuerySessionInformationW(hServer, sessionId, WTSUserName,
+        if (WTSQuerySessionInformation(hServer, sessionId, WTSUserName,
```
giampaolo, line 1243: I think it makes more sense to keep the old unicode name.

psutil/_psutil_windows.c
```diff
@@ -1263,26 +1275,23 @@ psutil_users(PyObject *self, PyObject *args) {
...
-                hServer,
-                sessionId,
-                WinStationInformation,
-                &station_info,
-                sizeof(station_info),
-                &returnLen))
-        {
-            PyErr_SetFromOSErrnoWithSyscall("WinStationQueryInformationW");
+        bytes = 0;
+        if (WTSQuerySessionInformation(hServer, sessionId, WTSSessionInfo,
```
giampaolo, line 1279: same here, keep the old name

giampaolo:
Merged. Glad to see a new platform being supported.

Issue #1791, opened 2020-07-21
[macOS] Use PAGE_SIZE macro instead of getpagesize() to build on macOS big sur.


doneria-anjali:
During psutil library installation on mac OS 11.0 (aka Big Sur), during building, the setup fails with error:
psutil/_psutil_osx.c:516:20: error: implicit declaration of function 'getpagesize' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        int pagesize = getpagesize();

Enhancement is to support installation of psutil on macOS 11.0 using python-pip by using PAGE_SIZE instead of getpagesize(). PAGE_SIZE is a macro that gives you the vm_page_size global.

giampaolo:
Please make a PR.

joefedorowicz:
Received the same error trying to install 5.7.3 using python 3.5.10.

Was this fixed in that version?

`    psutil/_psutil_osx.c:527:20: error: implicit declaration of function 'getpagesize' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        int pagesize = getpagesize();
                       ^
    psutil/_psutil_osx.c:567:20: error: implicit declaration of function 'getpagesize' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        int pagesize = getpagesize();
                       ^
    2 errors generated.
    error: command 'clang' failed with exit status 1
    ----------------------------------------
ERROR: Command errored out with exit status 1: /Users/josephfedorowicz/virtualenvs/ds2/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/3y/38591xfj2tx4jcd2fy208qk80000gn/T/pip-install-1yped8y1/psutil/setup.py'"'"'; __file__='"'"'/private/var/folders/3y/38591xfj2tx4jcd2fy208qk80000gn/T/pip-install-1yped8y1/psutil/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/var/folders/3y/38591xfj2tx4jcd2fy208qk80000gn/T/pip-record-vxxeg9sk/install-record.txt --single-version-externally-managed --compile --install-headers /Users/josephfedorowicz/virtualenvs/ds2/include/site/python3.5/psutil Check the logs for full command output.`

PR #1794, opened 2020-07-21
Use PAGE_SIZE macro instead of getpagesize() for macOS.

doneria-anjali:
For issue 1791 (https://github.com/giampaolo/psutil/issues/1791)

### DIFF
--- a/psutil/_psutil_osx.c
+++ b/psutil/_psutil_osx.c
@@ -513,7 +513,7 @@ psutil_virtual_mem(PyObject *self, PyObject *args) {
     uint64_t total;
     size_t   len = sizeof(total);
     vm_statistics_data_t vm;
-    int pagesize = getpagesize();
+    int pagesize = PAGE_SIZE;
     // physical mem
     mib[0] = CTL_HW;
     mib[1] = HW_MEMSIZE;
@@ -553,7 +553,7 @@ psutil_swap_mem(PyObject *self, PyObject *args) {
     size_t size;
     struct xsw_usage totals;
     vm_statistics_data_t vmstat;
-    int pagesize = getpagesize();
+    int pagesize = PAGE_SIZE;
 
     mib[0] = CTL_VM;
     mib[1] = VM_SWAPUSAGE;


kainjow:
@doneria-anjali `PAGE_SIZE` value is only known at compile time, while Apple Silicon needs to know this at runtime.

Can you try just adding `#include <unistd.h>`? This function should still be available AFAIK.

kainjow:
I confirmed the include is all you need with this little hack:

```sh
echo "#include <unistd.h>" > /tmp/test.h
CFLAGS="-include /tmp/test.h" pip install psutil
```

doneria-anjali:
It works! Thanks for the help Kevin.

Without this little hack, it gives error. Shall I change the use of PAGE_SIZE to something else that works both compile time and runtime?

doneria-anjali:
A friendly ping to follow up on this pull request. Thanks!

giampaolo:
Hello. I currently don't have an OSX box to test against, but this error:

```
ImportError: dynamic module does not define init function (init_psutil_osx)
```

...seems unrelated by the introduction of `#include <unistd.h>`. Are you sure you didn't change something else or you accidentally messed up psutil installation? In this case (messed up psutil installation) I suggest to run `make uninstall` (a couple of times) and `make clean` first, then retry.
It seems to me using `#include <unistd.h>` + `getpagesize()` should be the way to go, because:
1) it is documented by Apple:
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getpagesize.3.html
2) it is being used by cPython's `resource.getpagesize()`: https://github.com/python/cpython/blob/master/Modules/resource.c#L309-L325
3) in the unlikely event the OS/user changes the default page size, `getpagesize()` syscall is supposed to return an updated result, while `PAGE_SIZE` macro will not

Hope this helps.

giampaolo:
> In this case (messed up psutil installation) I suggest to run make uninstall (a couple of times) and make clean first, then retry.

Note: by default python3 will be used, so if you're on python 2 you actually want this instead:

```
make uninstall PYTHON=python
make uninstall PYTHON=python
make clean PYTHON=python
```

ThiagoIze:
I can also confirm that adding `#include <unistd.h>` fixes the build. Can we have that committed please?

giampaolo:
I proceeded by adding `#include <unistd.h>`. If this didn't work please reopen #1791.

PR #1800, opened 2020-08-01
Implement Process.environ() on *BSD family

### LINKED ISSUE #893: Implement Process.environ() on FreeBSD
`procstat -e PID` is able to get the environ; so should we.

ArminGruner:
Hi, I've added support for FreeBSD almost two years ago, but forgot to create a pull request ;-)
While I've been there, I also added support for NetBSD and OpenBSD lately.

Also, there are some fixes to make the C extension compile again on older FreeBSD, and recent NetBSD versions.

This closes #893

Greetings from Munich,
Armin


giampaolo:
Oh nice! I see some errors (errno 0):
https://github.com/giampaolo/psutil/pull/1800/checks?check_run_id=935540843
Since `kvm` is available from FreeBSD 8 and `libprocstat` requires FreeBSD9, why don't avoid using `libprocstat` and only rely on `kvm`?

ArminGruner:
> Oh nice! I see some errors (errno 0):
> https://github.com/giampaolo/psutil/pull/1800/checks?check_run_id=935540843

Hi,
actually I had some issues with the KVM interface with later FreeBSD versions. But I don't remember why I used the procstat interface (probably because I've checked both "ps" and "procstat" utils and copied code from there..).
But you're right, it look with some minor tweaks the KVM interfaces also works for later FreeBSD versions.
Let me check whether it doesn't break OpenBSD or NetBSD, I will update the pull request soon!
Thanks!

giampaolo:
Sounds good. ;)

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+    if (!PyArg_ParseTuple(args, "l", &pid))
+        return NULL;
+
+#if defined(PSUTIL_FREEBSD)
+    kd = kvm_openfiles(NULL, "/dev/null", NULL, 0, errbuf);
+#else
+    kd = kvm_openfiles(NULL, NULL, NULL, KVM_NO_FILES, errbuf);
+#endif
+    if (!kd) {
+        PyErr_SetString(PyExc_RuntimeError, errbuf);
```
giampaolo, line 425: There is a utility function, currently exposed for OpenBSD only, which can convert this kvm error:
https://github.com/giampaolo/psutil/blob/994c429c010049a6a0556ce3b0d1af1f86f27867/psutil/arch/openbsd/specific.c#L50-L61

I think it makes sense to move `convert_kvm_err` out of `psutil/arch/openbsd/specific.c` and put it into `psutil/_psutil_common.c` (guarded by an `#ifdef BSD...`) so that you can invoke it from this file, like this:

```c
    if (! kd) {
        convert_kvm_err("kvm_openfiles", errbuf);
        return NULL;
    }
```

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+    p = kvm_getprocs(kd, KERN_PROC_PID, pid, (int *)&cnt);
+#elif defined(PSUTIL_OPENBSD)
+    p = kvm_getprocs(kd, KERN_PROC_PID, pid, sizeof(*p), &cnt);
+#elif defined(PSUTIL_NETBSD)
+    p = kvm_getproc2(kd, KERN_PROC_PID, pid, sizeof(*p), &cnt);
+#endif
+
+    if (!p)
+    {
+        PyObject *ret = NoSuchProcess(kvm_geterr(kd));
```
giampaolo, line 439: I'm not completely sure this is the correct action. 
`kvm_getprocs` [doc](https://www.freebsd.org/cgi/man.cgi?query=kvm_getprocs&sektion=3&n=1) is not clear on whether you can/should use `kvm_geterr`, but I suppose you can. What I'm worried about here is that the error may not necessarily be a string like "no such process". Perhaps it can also be "access denied". Or something else (in which case we want `RuntimeError`). So you should probably parse the string and decide what to do.

You could try a test: call this function for a PID which does not exist. If `kvm_geterr` returns a string indicating that, then it means you can rely on `kvm_geterr`. 

In that case, what I recommend, is reusing and extending `convert_kvm_error` utility function (check if string is something like "no such process"), after you move it into `psutil/_psutil_common.c` (see my comment above).

The resulting code here should be something like this:

```c
    if (!p) {
        convert_kvm_err("kvm_getprocs", kvm_geterr(kd));
        kvm_close(kd);
        return NULL;
    }
```

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+        kvm_close(kd);
+        return ret;
+    }
+
+    if (!cnt)
+    {
+        kvm_close(kd);
```
giampaolo, line 446: I suggest to put all these `kvm_close(kd);` in the `error` section below, and do `goto error;` here.

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+    envs = kvm_getenvv2(kd, p, 0);
+#else
+    envs = kvm_getenvv(kd, p, 0);
+#endif
+
+    if (!envs)
+    {
+        kvm_close(kd);
+        errno = EPERM;
+        return PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo, line 460: You can just do:
```c
AccessDenied("kvm_getenvv");   // <- this will set EPERM
goto error;
```

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+    }
+
+    py_retdict = PyDict_New();
+    if (!py_retdict)
+        goto error;
+
+    for (i = 0; envs[i] != NULL; i++)
+    {
+        char *p;
+        PyObject *py_value;
```
giampaolo, line 470: both of these should be declared at the top

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+        }
+        Py_DECREF(py_value);
+    }
+
+    kvm_close(kd);
+    return py_retdict;
+
+error:
+    if (py_retdict)
+        Py_DECREF(py_retdict);
```
giampaolo, line 492: Add `Py_XDECREF(py_value);`

psutil/_psutil_bsd.c
```diff
@@ -391,6 +396,104 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+#endif
+
+    if (!envs)
+    {
+        kvm_close(kd);
+        errno = EPERM;
+        return PyErr_SetFromErrno(PyExc_OSError);
+    }
+
+    py_retdict = PyDict_New();
```
giampaolo, line 463: Please declare this at the top (and fail immediately if it fails).

psutil/arch/freebsd/specific.c
```diff
@@ -517,7 +517,7 @@ psutil_swap_mem(PyObject *self, PyObject *args) {
 }
 
 
-#if defined(__FreeBSD_version) && __FreeBSD_version >= 800000
+#if defined(__FreeBSD_version) && __FreeBSD_version >= 701000
```
giampaolo, line 520: Does this mean FreeBSD 7 was not supported and now it is?

setup.py
```diff
@@ -188,6 +188,10 @@ def get_winver():
 
 elif FREEBSD:
     macros.append(("PSUTIL_FREEBSD", 1))
+    libraries = ["devstat"]
+    os_major = int(platform.release().split(".")[0])
+    if os_major >= 8:
+        libraries.append("kvm")
```
giampaolo, line 194: Is this really necessary? It wasn't thus far. I'm worried about the string parsing. If it fails, the whole installation breaks.

psutil/_psutil_bsd.c
```diff
@@ -73,7 +74,11 @@
         #include <utmp.h>  // system users
     #else
         #include <utmpx.h>
-    #endif
+        #endif
```
giampaolo, line 77: no need to indent =)

giampaolo (changes requested):
Initial review. Little style request - please use:

```c
if (something) {
    ...
}
```

Instead of:

```c
if (something) 
{
    ...
}
```

psutil/_psutil_bsd.c
ArminGruner: Ah ok, nice, I've moved it.

psutil/_psutil_bsd.c
ArminGruner: I've checked the  `libkvm` sources and `/usr/src/bin/ps/ps.c` on all three BSD variants; they all use `kvm_err()` if there's a failing `kvm_*()` call.

I've written a tiny standalone test program to check for all return values (on all three platforms).
Fortunately, they all behave the same.

- `kvm_open*` isn't likely to fail with `EACCESS`, as it doesn't open `/dev/[k]mem` as in the old days.
It will only fail if the library cannot allocate internal memory, or cannot open the kernel boot file. `errno` will be set appropriate.
The question is which exception should be raised - `RuntimeError` or `OSError`?

- `kvm_getprocs` isn't likely to fail _itself_.
If the search for the PID fails, `cnt` is set to `0`. If the function itself returns `NULL`, it means that it was not able to allocate memory for the result, or the call parameters were invalid due to a programming error (`&cnt` pointing to invalid storage).
We can raise `RuntimeError` or `OSError` if the call itself failed.
If the call reports `cnt == 0`, this maps to `ESRCH` ("no such process").

I understand your concern mapping system errors to different Python errors;
now from what I've tested, `errno` is set appropriate after all failing `libkvm` calls.

There is just _one_ race which comes to my mind: When calling `kvm_getprocs()`, it may return success.
But before our call to `kvm_getenvv()`, we might have been preempted by the kernel, and the process in question might have terminated.
In this case, our later call to `kvm_getenvv()` will return `EINVAL`, and we probably should remap this to `ESRCH`?

psutil/arch/freebsd/specific.c
ArminGruner: Yes, I am able to compile on FreeBSD 7.1 after the change in `4b269a7`:

```shell
> python2 -c 'import os, psutil; print os.uname()[3], psutil.Process().environ()["OSTYPE"]'
FreeBSD 7.1-RELEASE #0: Thu Jan  1 08:58:24 UTC 2009     root@driscoll.cse.buffalo.edu:/usr/obj/usr/src/sys/GENERIC 
FreeBSD
```

Actually, we have bad rain here the whole day ;-),  and just for fun I've even managed to:
```shell
> python2 -c 'import os, psutil; print os.uname()[3], "\n", psutil.Process().environ()["OSTYPE"]'
FreeBSD 6.4-RELEASE #0: Wed Nov 26 08:21:48 UTC 2008     root@palmer.cse.buffalo.edu:/usr/obj/usr/src/sys/GENERIC 
FreeBSD
```
... though for compiling on `6.4`, I had to cut off some source bits in
`psutil/arch/freebsd/proc_socks.c`, `psutil/arch/freebsd/specific.c` and `psutil/arch/freebsd/sys_socks.c` because the 6.4 kernel interface is too old for that code...

setup.py
ArminGruner: Thanks for the reminder.
As mentioned, I actually more or less rebased the changes I made years ago, and for whatever reason it seemed neccessary then - probably I used a `.0` FreeBSD version which had some library glitches.

With all the (FreeBSD) versions mentioned in the commit log, it is not neccessary to specific `libkvm` at all, as `libdevstat` pulls it already!

psutil/_psutil_bsd.c
giampaolo: > kvm_open* isn't likely to fail with EACCESS, as it doesn't open /dev/[k]mem as in the old days.
It will only fail if the library cannot allocate internal memory, or cannot open the kernel boot file. errno will be set appropriate.
The question is which exception should be raised - RuntimeError or OSError?

If `errno` is set then it's easy: just use `PyErr_SetFromErrno(PyExc_OSError)`. My understanding though is that `kvm_` functions do not set `errno`, so we need to parse `kvm_geterr`. As a general rule, if `errno` is not set, parse `kvm_geterr` result and:

- set `ESRCH` if the string indicates the process is gone (use `NoSuchProcess("...");`)
- set `EPERM` if the string indicates some lack of privileges (use `AccessDenied("...");`)
- raise/call `PyErr_NoMemory()` if the string indicates it could not allocate memory
- raise `RuntimeError` in all other circumstances; if you can, also include the result of `kvm_geterr`

This logic (3 `if`s, 1 `else`) should be handled by `convert_kvm_err` utility function, and (if you feel like it, while you're at it =)) reused by all functions that use `kvm_*` APIs.
Maybe `convert_kvm_err` could have another `if` condition, which is executed first:

```c
if (errno != 0)
       PyErr_SetFromErrno(PyExc_OSError)
````

...but this assumes that one of the previous `kvm_` calls **did** set `errno`, but again, my understanding is that they usually don't (the various man pages never mention `errno`).

> kvm_getprocs isn't likely to fail itself. If the search for the PID fails, cnt is set to 0 [...]

Then raise `NoSuchProcess("...");`

> [...] If the function itself returns NULL, it means that it was not able to allocate memory for the result, or the call parameters were invalid due to a programming error (&cnt pointing to invalid storage, kd invalid or NULL, meaning a previous kvm_open* did not suceed). We can raise RuntimeError or OSError if the call itself failed.

Mmmm... if we have a way to differentiate with `kvm_geterr` (is it ENOMEM? is it EINVAL? ....) then I would say to just stick with `convert_kvm_err` utility fun. 

> If the call reports cnt == 0, this maps to ESRCH ("no such process"). This will also happen if the process with the given PID exists, but sysctl security.bsd.see_other_uids == 0.

Damn... This I don't know. =)

> There is just one race which comes to my mind: When calling kvm_getprocs(), it may return success.
But before our call to kvm_getenvv(), we might have been preempted by the kernel, and the process in question might have terminated. In this case, our later call to kvm_getenvv() will return EINVAL, and we probably should remap this to ESRCH?

Mmm... hard call. I would say yes, let's map it to `ESRCH`.

Comment (ArminGruner) [reply psutil/_psutil_bsd.c]: The latest update of the pull requests contains the changes I've mentioned so far.

Now `psutil/tests/test_process.py` passes all `Process().environ()` related stuff on all `*BSD` variants.

Unfortunately, CirrusCI is now not happy with these latest changes.
Interestingly, the version yesterday passed the checks, which makes me a bit puzzled.

```code
ERROR: psutil.tests.test_contracts.TestFetchAllProcesses.test_all

File "/tmp/cirrus-ci-build/psutil/_psbsd.py", line 674, in environ
    return cext.proc_environ(self.pid)
OSError: [Errno 22] Invalid argument
```

Is it possible that there is some incosistency what test_process.py checks, and what test_contracts.py checks? It looks that someone is expecting an `OSError` exception where the other one expects a `NoSuchProcess` exception?

The whole test suite does not pass altogether, though it looks that it doesn't pass on the `5.7.2` release tag either?
That happens on all `*BSD` platforms here..

ArminGruner: > My understanding though is that kvm_ functions do not set errno, so we need to parse kvm_geterr

Actually, almost _all_ library functions do some system calls, *especially* the `kvm*` stuff.
So, after _any_ system call which failed, there will be `errno` set.

Usually library utility functions do not clear `errno` if they return an error condition themself.

This avoids inventing a dozen more library error codes if most stuff may be mapped on typical `errno` codes anyway.

giampaolo: > Actually, almost all library functions do some system calls, especially the kvm* stuff. So, after any system call which failed, there will be errno set.

Got it. Then I think it makes sense to do this:

```c
#if defined(PSUTIL_FREEBSD) || defined(PSUTIL_OPENBSD) || defined(PSUTIL_NETBSD)
void
convert_kvm_err(const char *syscall, char *errbuf) {
    char fullmsg[8192];

    sprintf(fullmsg, "(originated from %s: %s)", syscall, errbuf);
    if (errno != 0)
        PyErr_SetFromOSErrnoWithSyscall(syscall);
    else if (strstr(errbuf, "Permission denied") != NULL)
        AccessDenied(fullmsg);
    else if (strstr(errbuf, "Operation not permitted") != NULL)
        AccessDenied(fullmsg);
    else
        PyErr_Format(PyExc_RuntimeError, fullmsg);
}
#endif
```

As for `EINVAL`, it's not clear to me where it originates from in `psutil_proc_environ` function. You should put `printf()`s to figure that out. At that point I'd say you'll have to decide what to do: whether to raise `NoSuchProcess` or `AccessDenied`.
From C you can use `psutil_pid_exists` to check if the PID exists. If it does raise AD, else NSP. It's done elsewhere in the C code: grep for `psutil_pid_exists`.

giampaolo: ...e.g. here:
https://github.com/giampaolo/psutil/blob/5e47e0bcba7585909a11e9c7424e864fc26a7fd2/psutil/arch/osx/process_info.c#L166-L174
Even if this is OSX, it may be that on BSD `kvm_getenvv` may raise `EINVAL` in case of zombie process.

giampaolo:
Judging from the [latest test failure](https://github.com/giampaolo/psutil/pull/1800/checks?check_run_id=939681032) it seems you cannot rely on `errno` after `kvm_getenvv`, or at least not all the time.
I took a peek at FreeBSD source code for `kvm_getenvv`. It seems sometimes you can rely on `errno`:
https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L755
...sometimes you don't:
https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L757
...and sometimes you can only rely on the error string:
https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L723

So maybe, something like this (not tested)?

```c
#if defined(PSUTIL_NETBSD)
    envs = kvm_getenvv2(kd, p, 0);
#else
    envs = kvm_getenvv(kd, p, 0);
#endif
    if (!envs) {
        err = kvm_geterr(kd);
        if (err) {  // probably wrong
            convert_kvm_err("kvm_getenvv", err);
            goto error;
        }
        if (psutil_pid_exists(pid))
            py_err = AccessDenied("kvm_getenvv");
        else
            py_err = NoSuchProcess("kvm_getenvv");
        goto error;
    }
```

ArminGruner:
Well, let's have some fun ;-)

For a moment, let's assume that the code is _correct_,
- probably because we have a gut feeling it must be right
(and the code is so short anyway, it uses system interfaces which exist stable for multiple decades),
- because our standalone test code says so and works exactly what we expect,
- and probably also because code review from other sources indicate that it should work how it's written
(namely, /usr/src/bin/ps/ps.c, /usr/src/usr.bin/procstat/*.c).

So, why there are failures then? Well, probably because of good old [nightmare] reasons:
- unstable build environment
- unknown run environment
- unstable test suite
- wrong tests, assuming same standard behavior for all supported platforms, but platforms don't implement the same POSIX version

For example:

PLATFORM A:
FreeBSD ikarus.fritz.box 12.1-RELEASE-p5 FreeBSD 12.1-RELEASE-p5 GENERIC  amd64
(my notebook, Thinkpad T530 i7-3520M, Standard FreeBSD installation)
Python 3.7.8 (default, Jul 11 2020, 01:22:11)
Sources: origin https://github.com/giampaolo/psutil.git
Head: a0967043 - (3 weeks ago) pre-release - Giampaolo Rodola (HEAD -> master, tag: release-5.7.2, origin/master, origin/HEAD)
```shell
> gmake clean install test
FAIL: psutil.tests.test_posix.TestSystemAPIs.test_disk_usage
Ran 542 tests in 7.023s
FAILED (failures=1, skipped=220)
```
Oh well, one test probably broken. Again:

```shell
> gmake test
FAIL: psutil.tests.test_posix.TestSystemAPIs.test_disk_usage
AssertionError: 0.0 != 100.0 within 1 delta (100.0 difference)
FAIL: psutil.tests.test_bsd.BSDTestCase.test_disks
  File "/home/ag/repos/git-repos/giampaolo-psutil/psutil/tests/test_bsd.py", line 116, in test_disks
    self.assertEqual(usage.total, total)
AssertionError: 17994129408 != 17993093120
Ran 542 tests in 6.367s
FAILED (failures=2, skipped=220)
```
Hm, well, now two test failing. Interesting! So let's just run tests in a loop to see what's happening!

```shell
> while :
do
gmake test 2>&1 | fgrep "FAILED "
done
FAILED (failures=3, skipped=220)
FAILED (failures=2, skipped=220)
FAILED (failures=2, skipped=220)
FAILED (failures=3, errors=1, skipped=220)
FAILED (failures=3, skipped=220)
```

Okay, so now we have at least _one_ more test constantly failing.. uhm, probably related to our redirected stdout?!
(I will send a fix for this when I'm back in swing mood..)

Now let's be more brave, another platform with the same OS and Python version!

PLATFORM B:
FreeBSD magnus.fritz.box 12.1-RELEASE-p3 FreeBSD 12.1-RELEASE-p3 GENERIC  amd64
(my server, tons of disk space, Core i7, Standard FreeBSD installation)
Python 3.7.7 (default, Apr 28 2020, 01:32:17)
So, nearly identical (OS some patch levels behind, Python one minor version behind).

Sources: origin https://github.com/giampaolo/psutil.git
Head: a0967043 - (3 weeks ago) pre-release - Giampaolo Rodola (HEAD -> master, tag: release-5.7.2, origin/master, origin/HEAD)
```shell
gmake test 2>&1  | fgrep FAIL
psutil.tests.test_contracts.TestFetchAllProcesses.test_all ... FAIL
psutil.tests.test_posix.TestSystemAPIs.test_disk_usage ... FAIL
psutil.tests.test_process.TestProcess.test_terminal ... FAIL
psutil.tests.test_bsd.BSDTestCase.test_disks ... FAIL
psutil.tests.test_bsd.FreeBSDSystemTestCase.test_cpu_stats_ctx_switches ... FAIL
psutil.tests.test_bsd.FreeBSDSystemTestCase.test_cpu_stats_interrupts ... FAIL
FAIL
FAIL: psutil.tests.test_contracts.TestFetchAllProcesses.test_all
    FAIL: test_exe pid=1426, ret='/usr/local/sbin/cupsd'
FAIL: psutil.tests.test_posix.TestSystemAPIs.test_disk_usage
FAIL: psutil.tests.test_process.TestProcess.test_terminal
FAIL: psutil.tests.test_bsd.BSDTestCase.test_disks
FAIL: psutil.tests.test_bsd.FreeBSDSystemTestCase.test_cpu_stats_ctx_switches
FAIL: psutil.tests.test_bsd.FreeBSDSystemTestCase.test_cpu_stats_interrupts
FAIL: psutil.tests.test_bsd.FreeBSDSystemTestCase.test_cpu_stats_syscalls
FAILED (failures=7, errors=2, skipped=223)
```

Same platform, same OS versions, same python, a server with some more long-running processes, but now 7 failures? Oh heck..

Okay, well. Now let's try a different BSD!

OpenBSD openbsd6.localdomain 6.7 GENERIC.MP#182 amd64
(vagrant virtualbox)
Python 3.7.7 (default, May  8 2020, 13:05:15)

Sources: origin https://github.com/giampaolo/psutil.git
Head: a0967043 - (3 weeks ago) pre-release - Giampaolo Rodola (HEAD -> master, tag: release-5.7.2, origin/master, origin/HEAD)

```shell
gmake clean all install
psutil.tests.test_process.TestProcess.test_terminal ... FAIL
psutil.tests.test_system.TestCpuAPIs.test_cpu_times_comparison ... FAIL
psutil.tests.test_connections.TestSystemWideConnections.test_it ... FAIL
FAIL
psutil.tests.test_contracts.TestSystemAPITypes.test_net_connections ... FAIL
FAIL: psutil.tests.test_process.TestProcess.test_terminal
FAIL: psutil.tests.test_system.TestCpuAPIs.test_cpu_times_comparison
FAIL: psutil.tests.test_connections.TestSystemWideConnections.test_it
FAIL: psutil.tests.test_contracts.TestFetchAllProcesses.test_all
    FAIL: test_connections pid=0
FAIL: psutil.tests.test_contracts.TestSystemAPITypes.test_net_connections
FAILED (failures=5, errors=2, skipped=266)
```

Okay, 5 failures on recent OpenBSD version. Now let's see if the failure rate keeps stable:

```shell
while :; do make test 2>&1 | fgrep --line-buffered "FAIL "; done
```
No output anymore! Output redirection now causes the test suite to hang.. oh well.
Let's check it other way:

```shell
script -c "while :; do gmake test; done" /tmp/make.out
~~~ wait 5 minutes ~~~
CTRL-C
fgrep "FAILED " /tmp/make.out
FAILED (failures=4, errors=2, skipped=266)
FAILED (failures=4, errors=2, skipped=266)
FAILED (failures=4, errors=2, skipped=266)
FAILED (failures=4, errors=2, skipped=266)
FAILED (failures=4, errors=2, skipped=266)
```
Okay, on this platform there _seems_ to be no timing or other race in either the code under test, or the test suite.

Now let's move on! Back to the laptop, now with the new sources:

Let's check what's Cirrus CI says! (no idea why it's labelled "freebsd_13_py2", actually we shall not fool ourself, actually it's
*12.1 and py3*!

```shell
Cirrus CI / freebsd_11_py2 failed 13 hours ago in 2m 49s
Task Summary
Instruction main failed in 00:17
Details
x 00:17 main

  File "/tmp/cirrus-ci-build/psutil/_psbsd.py", line 673, in environ
    return cext.proc_environ(self.pid)
ProcessLookupError: [Errno 3] No such process (originated from kvm_getenvv)
psutil.NoSuchProcess: psutil.NoSuchProcess process no longer exists (pid=1, name='init')
AssertionError: NoSuchProcess not raised by Process

The above exception was the direct cause of the following exception:

  File "/tmp/cirrus-ci-build/psutil/tests/test_contracts.py", line 397, in test_all
```

Okay, then now let's run this test suite locally!

Sources: origin git@github.com:ArminGruner/psutil.git
Head: c08876b3 - (2 days ago) Implement Process.environ() on *BSD family - Armin Gruner (HEAD -> master, origin/master, origin/HEAD)

```shell
gmake clean install
» py psutil/tests/test_contracts.py
psutil.tests.test_contracts.TestAvailConstantsAPIs.test_PROCFS_PATH ... OK
psutil.tests.test_contracts.TestAvailConstantsAPIs.test_linux_ioprio_linux ... OK
psutil.tests.test_contracts.TestAvailConstantsAPIs.test_linux_ioprio_windows ... OK
psutil.tests.test_contracts.TestAvailConstantsAPIs.test_linux_rlimit ... OK
psutil.tests.test_contracts.TestAvailConstantsAPIs.test_win_priority ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_cpu_affinity ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_cpu_num ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_environ ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_gids ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_io_counters ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_ionice ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_memory_maps ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_num_fds ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_num_handles ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_rlimit ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_terminal ... OK
psutil.tests.test_contracts.TestAvailProcessAPIs.test_uids ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_battery ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_cpu_freq ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_sensors_fans ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_sensors_temperatures ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_win_service_get ... OK
psutil.tests.test_contracts.TestAvailSystemAPIs.test_win_service_iter ... OK
psutil.tests.test_contracts.TestFetchAllProcesses.test_all ... OK
psutil.tests.test_contracts.TestProcessWaitType.test_negative_signal ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_boot_time ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_cpu_count ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_cpu_freq ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_cpu_percent ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_cpu_times ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_cpu_times_percent ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_disk_io_counters ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_disk_partitions ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_net_connections ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_net_if_addrs ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_net_if_stats ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_net_io_counters ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_sensors_fans ... skipped: not supported
psutil.tests.test_contracts.TestSystemAPITypes.test_sensors_temperatures ... OK
psutil.tests.test_contracts.TestSystemAPITypes.test_users ... OK

Ran 40 tests in 1.276s

OK (skipped=1)
SUCCESS
```

Okay. So, where do we go from here? (-:     To be continued...

giampaolo:
Hehe. Keeping psutil unit tests stable has always been like a never ending battle that I never really managed to win. There's always something that breaks occasionally. The "matrix" (different OSes, OS versions, CI services, python versions, etc.) is very complex, and 95% of the times I'm on Linux (my laptop) so I'm not sure how stable tests are on other platforms.
FreeBSD is definitively the platform which received more love amongst *BSDs, but I'm not sure when was the last time I turned on my virtualized OpenBSD and NetBSD boxes, so it's very likely that some tests on those platforms are broken and require a separate fix (either in the implementation or in the test). 

With that said, for this specific PR feel free to ignore any failure which is unrelated with this change (memory, disk, etc...). 
Judging from the latest test failures:
https://cirrus-ci.com/task/5945661432528896
https://cirrus-ci.com/task/4819761525686272
I see 2 problems:
- it seems we cannot determine environ for PID 0; this is common also on other platforms. For this I suggest to add a check at the top of the function and return an empty dict if PID == 0
- as for the second error:

```
FAIL: psutil.tests.test_contracts.TestFetchAllProcesses.test_all
multiprocessing.pool.RemoteTraceback: 
"""
  File "/tmp/cirrus-ci-build/psutil/_psbsd.py", line 551, in wrapper
  File "/tmp/cirrus-ci-build/psutil/_psbsd.py", line 673, in environ
    return cext.proc_environ(self.pid)
ProcessLookupError: [Errno 3] No such process (originated from kvm_getenvv)
  File "/tmp/cirrus-ci-build/psutil/tests/test_contracts.py", line 367, in proc_info
    info[fun_name] = fun()
  File "/tmp/cirrus-ci-build/psutil/__init__.py", line 862, in environ
    return self._proc.environ()
  File "/tmp/cirrus-ci-build/psutil/_psbsd.py", line 556, in wrapper
    raise NoSuchProcess(self.pid, self._name)
psutil.NoSuchProcess: psutil.NoSuchProcess process no longer exists (pid=1, name='init')
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/tmp/cirrus-ci-build/psutil/tests/test_contracts.py", line 369, in proc_info
    check_exception(exc, proc, name, ppid)
  File "/tmp/cirrus-ci-build/psutil/tests/test_contracts.py", line 344, in check_exception
    tcase.assertProcessGone(proc)
  File "/tmp/cirrus-ci-build/psutil/tests/__init__.py", line 902, in assertProcessGone
    self.assertRaises(psutil.NoSuchProcess, psutil.Process, proc.pid)
  File "/usr/local/lib/python3.6/unittest/case.py", line 733, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/local/lib/python3.6/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
  File "/usr/local/lib/python3.6/unittest/case.py", line 201, in __exit__
    self.obj_name))
  File "/usr/local/lib/python3.6/unittest/case.py", line 135, in _raiseFailure
    raise self.test_case.failureException(msg)
AssertionError: NoSuchProcess not raised by Process
```

I'm pretty sure the test is legitimate (aka not broken). What is broken is the implementation. I think that's because this part is not correct:

```c
    if (!envs) {
        if (errno == EPERM)
            py_err = AccessDenied("kvm_getenvv");
        else
            py_err = NoSuchProcess("kvm_getenvv");
        goto error;
    }
```

I think a possible solution is what I suggested in my previous comment https://github.com/giampaolo/psutil/pull/1800#issuecomment-667948629.  I didn't try it myself so I can be wrong (but I can try myself after pulling your branch if you're still in trouble - not sure when though). I think we're basically there though.

ArminGruner:
> I'm pretty sure the test is legitimate (aka not broken). What is broken is the implementation. I think that's because this part is not correct:
> 
>     if (!envs) {
>         if (errno == EPERM)
>             py_err = AccessDenied("kvm_getenvv");
>         else
>             py_err = NoSuchProcess("kvm_getenvv");
>         goto error;
>     }

You're surely right.
In languages without support for **promises** or similar stuff one should _never assume_ preconditions are met,
if they are actually not _explicitly_ checked ;-)


> Judging from the latest test failure it seems you cannot rely on errno after kvm_getenvv, or at least not all the time.

You know that I'm a great lover of *BSD.
I usually get into _defend BSD mode_ if I read sentences like this ;-)
(with respect to *BSD library code/release engineering quality, not my code bits!)
Mostly, but not always, this trigger is correct.. ;-)

So:

> I took a peek at FreeBSD source code for kvm_getenvv. It seems sometimes you can rely on errno:
> https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L755

Well, no, and ... yes ;-)

If `kvm_getenvv()` returns `NULL`, but `errno` == 0, this means the result did not fit into the buffer which was allocated.

You're right, my code assumes that it always fits, because `kenv_getenvv()` is called with `nchr=0`.

This actually indicates to the library that it should allocate _how much it takes_ (quoting Mario Draghi ;-)),
but it may fail in the third code branch you are mentioning (see below).
So my calling code must also assume (and handle) different `errno` codes than just `ESRCH` or `EPERM`.

> ...sometimes you don't:
> https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L757

No, this is clearly stated in the man page.

It states _if the result does not fit into the allocated buffer, it will be truncated_.
That's what the code is doing, and obviously it has to clear `errno` here,
because `sysctl()` has set it to indicate that the returned result
did not fit into the supplied buffer and _actually was_ truncated.

> ...and sometimes you can only rely on the error string:
>  https://github.com/freebsd/freebsd/blob/b4d009417b7bed8b076f0f5637885ce165ed3b02/lib/libkvm/kvm_proc.c#L723

No, this is not true, because if `bufp` is `NULL`, this means that `bufp = malloc(sizeof(char *) * argc)` failed,
and thus `malloc()` has set `errno = ENOMEM`.  ;-)


... and that's exactly what Cirrus CI shows with the latest pull request update.

Now, I am a bit lost and happy for any suggestions:

I don't know the exact runtime environment of Cirrus CI.
I tried to reproduce the currently failing unit test, but did not manage yet.
Even if I lower resource limits to provoke an memory allocation error, I do not get one.

I will have to investigate later, now for some less serious, but more vital work ;-)

ArminGruner:
Good evening .. well, what do you think?  -> https://cirrus-ci.com/task/5900109814693888

giampaolo:
Hey Armin. Mmm... since it seems we cannot avoid `ENOMEM` for certain PIDs, and since it's a critical error which would inevitably break code using `psutil.process_iter()`, I think the best compromise is to turn `ENOMEM` into `AccessDenied`. Something like this (not tested):

```c
    if (!envs) {
        /* Map the most obvious stuff to general high-level exceptions exported by our "psutil" */
        if (errno == EPERM)
            AccessDenied("kvm_getenvv");
        else if (errno == ESRCH)
            NoSuchProcess("kvm_getenvv");
        else if (errno == ENOMEM) {
            psutil_debug("kvm_getenvv -> ENOMEM, turned into AccessDenied");
            AccessDenied("kvm_getenvv -> ENOMEM, turned into AccessDenied");
        }
        else {
            sprintf(errbuf, "kvm_getenvv(pid=%d)", pid);
            PyErr_SetFromOSErrnoWithSyscall(errbuf);
        }
        goto error;
    }
```

ArminGruner:
Hi Giampaolo,
I've added some workarounds to avoid this error so far.

I've filed a bug report on FreeBSD, let's see if the can confirm that there is a problem (meanwhile I'm pretty sure there is; I also find stuff like
```shell
» fgrep getenvv /var/log/messages
Jul 23 15:03:47 ikarus console-kit-daemon[1625]: WARNING: kvm_getenvv failed: 
```
in syslog files)

I need to check the other *BSD platforms if they are still ok. The CI builds are all **green** now (well some macosx build seems to be flaky, but it's not related to my changes)

giampaolo:
Hey Armin. Thanks for investigating this issue so thoroughly. I seem to remember there was a similar problem with "corrupted env" on other platforms, but I don't remember the exact details (or maybe it was the cmdline(), anyway, I remember a similar problem occurring when a process env or cmdline was changed *after* the process was started). 

AFAICT, given how thorny this issue appears to be, I think it would be fine to just raise `NoSuchProcess` if PID does not exist, else give up and do `AccessDenied`, without additional logic which tries to determine what went wrong exactly.
In the end the result is the same: we won't get the information, so I think it's better to fail with `AccessDenied` rather than `RuntimeError`.

psutil/_psutil_bsd.c
```diff
@@ -74,6 +75,11 @@
     #else
         #include <utmpx.h>
     #endif
+    #include <fcntl.h>
+    #include <sys/param.h>
+    #include <sys/sysctl.h>
+    #include <sys/user.h>
+    #define PSUTIL_PROC_SKIP(p) (!((p)->ki_flag & P_INMEM) || ((p)->ki_flag & P_SYSTEM))
```
giampaolo, line 82: Is this necessary? If it is, I would rather call it `PSUTIL_PROC_ENV_SKIP`

psutil/_psutil_bsd.c
```diff
@@ -83,6 +89,7 @@
     #include <sys/file.h>
     #undef _KERNEL
     #include <sys/sched.h>  // for CPUSTATES & CP_*
+    #define PSUTIL_PROC_SKIP(p) ((p)->p_flag & P_SYSTEM)
```
giampaolo, line 92: Ditto.

psutil/_psutil_bsd.c
```diff
@@ -93,6 +100,7 @@
     #ifndef DTYPE_VNODE
         #define DTYPE_VNODE 1
     #endif
+    #define PSUTIL_PROC_SKIP(p) ((p)->p_stat == SZOMB)
```
giampaolo, line 103: Ditto.

psutil/_psutil_bsd.c
```diff
@@ -391,6 +399,132 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+
+#if defined(PSUTIL_FREEBSD)
+    p = kvm_getprocs(kd, KERN_PROC_PID, pid, &cnt);
+#elif defined(PSUTIL_OPENBSD)
+    p = kvm_getprocs(kd, KERN_PROC_PID, pid, sizeof(*p), &cnt);
+#elif defined(PSUTIL_NETBSD)
+    p = kvm_getproc2(kd, KERN_PROC_PID, pid, sizeof(*p), &cnt);
+#endif
+    if (!p) {
+        py_err = NoSuchProcess("kvm_getprocs");
```
giampaolo, line 443: You can avoid using `py_err = ...`. Just Calling NSP/AD will set the exception.

psutil/_psutil_bsd.c
```diff
@@ -391,6 +399,132 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+     *
+     * Some system process have no stats attached at all (they are marked with P_SYSTEM.)
+     *
+     * On FreeBSD, it's possible that the process is swapped or paged out,
+     * then we have no access to the environment stored in the process' user area.
+     *
+     * On NetBSD, we cannot call kvm_getenvv2() for a zombie process.
+     *
+     * To make unittest suite happy, return an empty environment.
+     */
```
giampaolo, line 462: Just a minor style complaint: could you please use `//` comments and limit the length of the string to 80 chars?

giampaolo (changes requested):
Hello Armin. I added a couple more (minor) comments. The rest LGTM.

psutil/_psutil_bsd.c
```diff
@@ -391,6 +392,145 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
...
+    }
+
+    kvm_close(kd);
+    return py_retdict;
+
+error:
+    Py_XDECREF(py_value);
+    Py_XDECREF(py_retdict);
+    kvm_close(kd);
+    return py_err;
```
giampaolo, line 531: Just `return NULL;` and get rid of `py_err`

giampaolo:
Are we adding support for FreeBSD 7 and/or other BSD platforms? (asking for the HISTORY.rst file entry)

ArminGruner:
> Are we adding support for FreeBSD 7 and/or other BSD platforms? (asking for the HISTORY.rst file entry)

Well, the code compiles under FreeBSD 7.1, but I cannot run the test suite.
Support for unittest2 and dependent packages is practially gone.

I've (manually) tested that _my_ code implementing Process.environ() works, and that the C code still compiles to a loadable exxtension. So I don't see a regression here.

For NetBSD and OpenBSD, I was able to run the test suite,  and the Process.environ() related changes work; though there are some tests failing which are unrelated to Process.environ().

giampaolo:
Merged. Thanks a lot for the great work Armin. It was a good patch. If you want to fix other BSD-related stuff you're very welcome. ;-) BSD* definitively need more maintenance.

ArminGruner:
Hi Giampaolo, thanks for merging!

I'm still in evaluation for an automated CI runner for the other *BSD family members.
Unfortunately, the _renowned_ candidates such as Appveyor or even GitHub CI require a dotnet 3.x environment (sic!). 
There has been some effort to bring dotnet to both OpenBSD and NetBSD, I will check this after my vacation to Denmark ;-)

giampaolo:
To be honest, having / maintaining extra CI for Open/NetBSD would be an additional effort which I'd prefer to avoid. I'd rather fix whatever is broken on those platforms and then leave them alone. The code for Open/NetBSD changes pretty rarely anyway. 
Have a good time in Denmark. I'm staying in Italy this time, and will spend occasional weekends camping in the nearby (north) mountains. =)

Issue #1830, opened 2020-09-20
[Linux] Network interface `isup` returned from `net_if_stats` doesn't represent interface state


phobozad:
**Platform**
* Debian 10.4
* psutil version: 5.5.1
* Python 3.7.3

**Bug description**
On Linux, the value of `isup` for network interfaces isn't accurate to the operational state of the interface (i.e. link is up).  

Currently the code looks for the flag `IFF_UP` to set the value of `isup`.  However, the `IFF_UP` flag only represents the interface being administratively up (i.e. not disabled) but not the operational state (i.e. connected).  The operational state is represented by the `IFF_RUNNING` flag.
https://www.kernel.org/doc/Documentation/networking/operstates.txt
https://stackoverflow.com/questions/11679514/what-is-the-difference-between-iff-up-and-iff-running

I suspect this issue is also why the test for `isup` in `test_linux.py` was disabled - the test was looking for the `RUNNING` flag but the code was looking at the `UP` flag.  Thus if there was an interface that was administratively up but didn't have a cable connected, it would fail the test suite.  In this case, the test failing was correct since the code wasn't actually checking the correct flag.

This same issue is also mentioned in relation to OSX Wireless status in #805.

giampaolo:
According to https://stackoverflow.com/a/11679939... it seems this make sense. I will change the doc in accordance.
As for Windows, it seems there's a similar distinction:
https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ms894611(v=msdn.10)
Currently we're doing:

```c
        // is up?
        if ((pIfRow->dwOperStatus == MIB_IF_OPER_STATUS_CONNECTED ||
                pIfRow->dwOperStatus == MIB_IF_OPER_STATUS_OPERATIONAL) &&
                pIfRow->dwAdminStatus == 1 ) {
            py_is_up = Py_True;
        }
        else {
            py_is_up = Py_False;
        }
```

I cannot disconnect the Ethernet cable on Windows because it's virtualized, but I did `ipconfig /release` (release IP address assigned by DHCP) and the status keeps being `MIB_IF_OPER_STATUS_OPERATIONAL`, which seems correct (the same as on Linux). As such, it appears Windows code is fine as it is.

Issue #805, opened 2016-04-15
speed/isup returned by net_if_stats are incorrect on OSX


weyou:
The link *speed* is always 0.
The link status *isup* is always *True* even the wireless is turn off or the cable is disconnected.

giampaolo:
We may have a better chance to figure out what's wrong with this when we get around #898.

alexbakker:
This issue also exists on Linux.

ewedlund:
"isup" reflects the value of the flag IFF_UP.

In OSX the flag IFF_UP is set even if the airport is disabled (using OSX tools, such as 'networksetup'), but the flag IFF_RUNNING reflects the airport state. Also, there is no way to set an ethernet interface to down (IFF_UP not set) and  IFF_RUNNING remains set even if the cable is disconnected. (It is however possible to manually set the IFF_UP flag through 'ifconfig')

In linux the flags are set as one would expect, i.e. IFF_UP is set if the interface is enabled/configured and IFF_RUNNING if it is connected.

Here is a description of the two (from [Stack Overflow):](https://stackoverflow.com/questions/11679514/what-is-the-difference-between-iff-up-and-iff-running)
___
From [Linux Device Drivers 3](http://www.xml.com/ldd/chapter/book/ch14.html#t3):

IFF_UP This flag is read-only for the driver. The kernel turns it on when the interface is active and ready to transfer packets.
...
IFF_RUNNING This flag indicates that the interface is up and running. It is mostly present for BSD compatibility; the kernel makes little use of it. Most network drivers need not worry about IFF_RUNNING.

Digging a bit deeper, it seems that there is one significant difference:

IFF_RUNNING is supposed to reflect the operational status on a network interface, rather than its administrative one. To provide an example, an Ethernet interface may be brought UP by the administrator (e.g. ifconfig eth0 up), but it will not be considered operational (i.e. RUNNING as per RFC2863) if the cable is not plugged in.
___
**I think there are two options, either to add 'is_running' to the interface dict, or to consider checking IFF_RUNNING instead of IFF_UP for 'isup'.**

Both the above solutions would solve the problem with WiFi status in OSX, but not the Ethernet status, since it appears to always be both "UP" and "RUNNING". However, the speed will be 0 on a disconnected Ethernet interface.

PR #1831, opened 2020-09-20
Use IFF_RUNNING instead of IFF_UP for Linux

phobozad:
Fix for #1830 and part of #805 

IFF_UP is administrative state
IFF_RUNNING is operational state

Value of `isup` should be based on operational state (i.e. link up or down).


giampaolo:
According to https://stackoverflow.com/a/11679939... it seems this make sense. I will change the doc in accordance.

PR #1837, opened 2020-09-26
Expanded battery sensor detection

aristocratos:
This expands the battery sensors detection for directories not starting with "BAT" but containing "battery" in the name. (As example on a Pinebook Pro running the default Manjaro distro the folder will be named "cw2015-battery").

It also looks for "/time_to_empty_now" (minutes) which can be used to get "secsleft" if either "energy_now" or "power_now" fails to get a value.

The TestSensorsBattery tests will however fail on a machine that has the "/time_to_empty_now" file present since it won't return "None" in that case. I figured I probably shouldn't mess with the tests and let you do it if this PR gets accepted.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1340,7 +1340,8 @@ def sensors_battery():
                 return int(ret) if ret.isdigit() else ret
         return None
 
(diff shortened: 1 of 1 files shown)


giampaolo:
Thank you.

PR #1843, opened 2020-10-05
remove old unused code

wiggin15:
"let's keep it, just in case" - it's been more than 6 years, and the code isn't Python 3 compatible. I think we can let it go ;)


giampaolo:
=)

PR #1847, opened 2020-10-11
Windows: Add mising system DLLs import libraries.

Biswa96:


giampaolo:
It looks like this breaks python 2.7
https://ci.appveyor.com/project/giampaolo/psutil/builds/35688603/job/q3d29ymct4bwyeou
What problem does this solve? Can you provide an error/traceback?

Biswa96:
This helps to build without using `GetProcAddress` function. The build error in Appveyor seems suspicious. The error is `LINK : fatal error LNK1181: cannot open input file 'ntdll.lib'`. It is related to Visual Studio. But `ntdll.lib` is present in any version of Windows SDK.

Biswa96:
This helps to build without using `GetProcAddress` function (here `psutil_loadlibs`). The build error in Appveyor seems suspicious. The error is `LINK : fatal error LNK1181: cannot open input file 'ntdll.lib'`. It is related to Visual Studio. But `ntdll.lib` is present in any version of Windows SDK.

giampaolo:
What's the problem with `GetProcAddress` / `psutil_loadlibs`?

Biswa96:
While compiling with GCC in Windows, there are may errors about multiple definition at linking time. Probably because many APIs are redefined in `ntexapi.h` file.

giampaolo:
GCC on Windows? Are you trying to compile for cygwin? Because in that case it's not supported.

Biswa96:
It's MSYS2/mingw.

giampaolo:
Also not supported (explicitly, I dropped mingw support years ago because supporting both compilers was a pain + the VS suite for compiling became free). You should use Visual Studio. setup.py could be be made a bit smarter to detect that and recommend VS though.

Biswa96:
> Also not supported

OK, as you wish.

> You should use Visual Studio. 

Visual Studio, the second most bloated IDE after Android Studio.

PR #1853, opened 2020-10-15
Fix labeling in Process cpu_times() documentation

kellurs:
Changed system_user to children_system to match the named tupple and expected behaviour.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1426,7 +1426,7 @@ Process class
     - **system**: time spent in kernel mode.
     - **children_user**: user time of all child processes (always ``0`` on
       Windows and macOS).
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1426,7 +1426,7 @@ Process class
     - **system**: time spent in kernel mode.
     - **children_user**: user time of all child processes (always ``0`` on
       Windows and macOS).
-    - **system_user**: user time of all child processes (always ``0`` on
+    - **children_system**: user time of all child processes (always ``0`` on
```
giampaolo, line 1429: Oh! Good catch. Also the explanation is wrong:

```
user time of all child processes ...
```

...it should be

```
system time of all child processes ...
```

PR #1861, opened 2020-10-24
Remove Python 3.5 from Travis

gliptak:


giampaolo:
Mmm... let's wait a little while longer. 3.5 is 5 years old, it's not so old. Let's get rid of it when 3.10 comes out.

PR #1862, opened 2020-10-24
Pin Python 3.5.10 in Travis

gliptak:
https://github.com/giampaolo/psutil/pull/1861


giampaolo:
That section is for OSX. On OSX I think it's not worth it (test runs are very slow).

PR #1869, opened 2020-11-04
'.exe()' method now returns "System Idle Process" and "System"

noobaly:
When you try to use `.exe()` method on a `Process` that has a pid of either '0' or '4' the program raises the `psutil.AccessDenied` exception. Instead of handling this with a `try: except psutil.AccessDenied:` I added a few lines of code so that the program would just return "System Idle Process" or "System"

**BEFORE:**
image
image

**AFTER:**
image

### DIFF
--- a/psutil/_pswindows.py
+++ b/psutil/_pswindows.py
@@ -752,6 +752,14 @@ class Process(object):
     @wrap_exceptions
     @memoize_when_activated
     def exe(self):
+
+        # Accessing protected system processes raises an error, this is to prevent that from happening.
+        if self.pid == 0:
+            return "System Idle Process"
+        if self.pid == 4:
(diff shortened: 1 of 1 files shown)


giampaolo:
`exe()` is supposed to return the absolute executable path, not the name. What you did in there is returning the name, which is the same logic currently defined in `name()` method on Windows.

noobaly:
> `exe()` is supposed to return the absolute executable path, not the name. What you did in there is returning the name, which is the same logic currently defined in `name()` method on Windows.

Oh, i thought since it returns "Registry" as "Registry" it'd would be fine, returning an empty string would cause some "new lines"
that's why i returned their names. Thank your for your time anyways 👍

PR #1880, opened 2020-11-15
Remove Travis and Cirrus, use GH also for FreeBSD

giampaolo:
Get rid of Travis (yes, after so many years) and Cirrus services for CI testing. Reason: both of them are no longer free for public repos (require payment). From now on, GitHub actions will take care of CI testing and will generate wheels on:

* Linux (tests + wheels)
* macOS (tests + wheels)
* FreeBSD (tests only) 

It turns out it is possible to run FreeBSD on GitHub Actions by making it run macOS and virtualize FreeBSD on top of it. 

AppVeyor will take care of Windows (tests + wheels). Also GitHub Actions is able to do that (I left the commented code in place), but it takes around 18 mins to complete. AppVeyor is also relatively slow, but this way we split the load and keep Linux/macOS/FreeBSD builds faster and more agile when pushing changes.

Related to #1879.


fkorotkov:
@giampaolo just FYI Cirrus CI is still free and was free at the time of this PR. We don't have any monthly limits on CPU minutes used as you might imply.

PR #1895, opened 2020-12-20
#1892 fix cpu_freq() broken on Apple M1

### LINKED ISSUE #1892: [MacOS] psutil.cpu_freq() broken on Apple M1
## Summary

* OS: MacOS
* Architecture: Apple M1
* Psutil version: N/A
* Python version: N/A
* Type:  core

## Description

Sorry that I couldn't provide some of the info in the template as I don't own an Apple M1 device, but it is highly suspected that calling `psutil.cpu_freq()` on Apple M1 is failing as the `hw.cpufrequency` sysctl call ([used here](https://github.com/giampaolo/psutil/blob/60475128881f3c89608dfc45e62bf3872c32a427/psutil/_psutil_osx.c#L696)) was removed on this arch (can be checked with `sysctl hw.cpufrequency` on any Apple M1 device, while it is working on amd64).

See https://github.com/shirou/gopsutil/pull/999 and https://github.com/shirou/gopsutil/issues/1000

In https://github.com/shirou/gopsutil/pull/999 a fix was suggested but I couldn't confirm it.

psutil user-base is probably bigger than gopsutil's so I hope this will help both projects find a solution or confirm @shoenig fix in his PR.

giampaolo:
## Summary

* OS: macOS M1
* Bug fix: yes
* Type: core
* Fixes: #1892, #1456

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,5 +1,17 @@
 *Bug tracker at https://github.com/giampaolo/psutil/issues*
 
+5.8.0 (IN DEVELOPMENT)
+======================
+
+XXXX-XX-XX
+
+**Bug fixes**
+
+- 1456_: [macOS] psutil.cpu_freq()'s min and max are set to 0 if can't be
+  determined (instead of crashing).
+- 1892_: [macOS] psutil.cpu_freq() broken on Apple M1.
+
+
 5.8.0
 =====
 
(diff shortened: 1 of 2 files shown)


giampaolo:
This was not fixed, see https://github.com/giampaolo/psutil/issues/1892#issuecomment-1483995613.

Issue #1877, opened 2020-11-13
OSError: [WinError 0] The operation completed successfully


rhazz:
**Platform**
* { Windows nt 2012 }
* { psutil version: 5.7.3" }
* { python 3.8.2}

**Bug description**
Version 5.7.3 again presented this error corrected in version 3.6.7, I solved my problem by going back to version 3.6.7 ... the error ran on windows server 2012 R2 Datacenter virtualized by VMware

**Test results**
```

import psutil
from datetime import datetime, timedelta
processName='chrome'
listOfProcessObjects = []

for proc in psutil.process_iter():
    try:
        pinfo = proc.as_dict(attrs=['pid', 'name', 'create_time'])
        pinfo['started'] = datetime.fromtimestamp(pinfo['create_time'])
          # Check if process name contains the given name string.
        if pinfo['name'] and processName.lower() in pinfo['name'].lower():
            listOfProcessObjects.append(pinfo)
    except (psutil.PermissionError, psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
        pass

print(listOfProcessObjects)
```

```
Traceback (most recent call last):
  File ".\teste.py", line 6, in <module>
    for proc in psutil.process_iter():
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\__init__.py", line 1457, in process_iter
    yield add(pid)
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\__init__.py", line 1432, in add
    proc = Process(pid)
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\__init__.py", line 346, in __init__
    self._init(pid)
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\__init__.py", line 373, in _init
    self.create_time()
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\__init__.py", line 723, in create_time
    self._create_time = self._proc.create_time()
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\_pswindows.py", line 681, in wrapper
    raise convert_oserror(err, pid=self.pid, name=self._name)
  File "C:\Users\user10\AppData\Roaming\Python\Python38\site-packages\psutil\_pswindows.py", line 671, in convert_oserr
or
    raise exc
[...]

PR #1904, opened 2021-01-08
windows: Preserve GetLastError() value before calling sprintf

alxchk:
## Summary

* OS: Windows
* Bug fix: yes
* Type: core
* Fixes: https://github.com/giampaolo/psutil/pull/1887

## Description
https://github.com/giampaolo/psutil/issues/1877#issuecomment-756342272


giampaolo:
Thanks for the investigation at https://github.com/giampaolo/psutil/issues/1877#issuecomment-756342272. I imagine this was painful to track down. Hopefully the "error 0" problem is finally gone.

Issue #1874, opened 2020-11-11
[Solaris] Swap output wrong on some OS versions


jomann09:
**Platform**
* Solaris 11.4
* psutil 5.7.3
* Any version

**Bug description**
It looks like on Solaris 11.4 the output of the `swap` command can have an "encrypted" column.
```
$ /usr/sbin/swap -l
swapfile                 dev            swaplo      blocks        free encrypted
/dev/zvol/dsk/rpool/swap 303,5               0     8388608     7072944      yes
```
which is likely causing an issue with this line here: https://github.com/giampaolo/psutil/blob/master/psutil/_pssunos.py#L158 and may be able to be remedied by using `line[3:4]` instead of `line[-2:]`

giampaolo:
Please make a PR.

jomann09:
Alright added a PR for it.

PR #1914, opened 2021-02-04
Fix incorrect range for Solaris swap output (#1874)

jomann09:
## Summary

* OS: Solaris
* Bug fix: yes
* Type: core
* Fixes: Typo in the range for swap output

## Description

Not sure how I missed this in the issue, but when I submitted the PR the range was not both columns, just one column. Updated to 3:5 instead of 3:4 to grab both blocks and free columns.


dimpase:
I just stumbled on this with psutil 5.8.0 - good that's fixed and I don't need to do a PR :-)

PR #1917, opened 2021-02-11
Linux: wait_procs ignoring timeout (#1913)

### LINKED ISSUE #1913: [Linux] wait_procs seemingly ignoring timeout, TimeoutExpired thrown
## Summary

* OS: Ubuntu 18.04
* Architecture: 64bit
* Psutil version: Both 5.6.1 and 5.8.0
* Python version: 3.7.5
* Type: core

## Description

The bug can be reproduced by creating these two files and running `python a.py`:

a.py:

```py
import signal
import subprocess
import psutil
import time

my_process = subprocess.Popen(["python", "b.py"])
time.sleep(2)
try:
    my_process.send_signal(signal.SIGTERM)
    _, alive = psutil.wait_procs([my_process], timeout=5)
    if alive:
        alive[0].kill()
        psutil.wait_procs(alive, timeout=5)

except Exception as exc:
    print("Sigterm failed: ", str(exc))
```

b.py

```py
import time
import signal

def handler(_signum, _frame):
    time.sleep(5)

signal.signal(signal.SIGTERM, handler)

time.sleep(50)
```

On my machine, this will print `Sigterm failed  Command '['python', 'b.py']' timed out after 1.0 seconds`. Exception is an instance of `subprocess.TimeoutExpired`. By debugging a bit I see that [this piece of code](https://github.com/giampaolo/psutil/blob/master/psutil/__init__.py#L1503-L1514) is where the exception initially appears. And since `psutil.TimeoutExpired` and `subprocess.TimeoutExpired` are different, it doesn't get caught.

I'm sure I'm doing something wrong as the example in the documentation is using a higher timeout, but I can't see the problem.

guille:
The function was exiting after one second due to a subprocess.TimeoutException

Fixes #1913

## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: #1913

## Description

Ensures the exception caught includes subprocess.TimeoutException as well

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -47,6 +47,7 @@ from ._common import ZombieProcess
 from ._compat import long
 from ._compat import PermissionError
 from ._compat import ProcessLookupError
+from ._compat import SubprocessTimeoutExpired as _SubprocessTimeoutExpired
 from ._compat import PY3 as _PY3
 
 from ._common import CONN_CLOSE
@@ -1504,6 +1505,8 @@ def wait_procs(procs, timeout=None, callback=None):
         try:
             returncode = proc.wait(timeout=timeout)
         except TimeoutExpired:
+            pass
+        except _SubprocessTimeoutExpired:
             pass
         else:
             if returncode is not None or not proc.is_running():
--- a/CREDITS
+++ b/CREDITS
@@ -736,3 +736,7 @@ I: 1822
 N: marxin
 W: https://github.com/marxin
 I: 1851
+
+N: guille
+W: https://github.com/guille
+I: 1913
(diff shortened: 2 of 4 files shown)


guille:
@giampaolo could you take a look when you have some time, please?

giampaolo:
`subprocess.TimeoutExpired` is not available in Python 2. I think you can add this to `psutil/_compat.py`:

```python
# python 3.3
try:
    from subprocess import TimeoutExpired as SubprocessTimeoutExpired
except ImportError:
    class SubprocessTimeoutExpired:
        pass
 ```
 
 ...and do `from psutil._compat import SubprocessTimeoutExpired` in `psutil/__init__.py`.

**Comment by guille** [comment]
Thanks for the pointer. Updated.

File: psutil/__init__.py [hunk 1]
```diff
@@ -47,6 +47,7 @@
 from ._compat import long
 from ._compat import PermissionError
 from ._compat import ProcessLookupError
+from ._compat import SubprocessTimeoutExpired
```
giampaolo, line 50: Sorry, on a second thought actually I prefer `from ._compat import SubprocessTimeoutExpired as _SubprocessTimeoutExpired` (let's keep it private in the main psutil namespace). Can you update it?

guille: Updated properly now, sorry for the noise.

Issue #1941, opened 2021-05-02
difference-method for timing results [OS] all


Barry1:
## Summary

* OS: { every supported }
* Type: { core, new-api }

## Description

The Timing results from `psutil` - let's begin with `cpu_times()` most often only makes sense as the difference between two calls. Haven `__sub__` from those `namedtuples` might help a lot.

### possible use case

I used methods from `cpu_load_generator` just for showing

```python
import psutil


def generatecpuload():
    from cpu_load_generator import load_single_core, load_all_cores

    load_single_core(
        core_num=0, duration_s=20, target_load=0.4
    )  # generate load on single core (0)
    load_all_cores(duration_s=30, target_load=0.2)  # generates load on all cores


beginsys = psutil.cpu_times()
beginproc = psutil.Process().cpu_times()
# do something which needs cpu ...
generatecpuload()
endsys = psutil.cpu_times()
endproc = psutil.Process().cpu_times()
print(f"sys difference {endsys-beginsys}")
print(f"process difference {endproc-beginproc}")
```

At the moment, the differences in the print statement are not available - but I think it might help. I try to send a merge request related to this issue soon.  
Maybe there are other positions, where it makes sense.

giampaolo:
Mmm... Interesting idea. There are some things to consider here. The most common use case for CPU times is to subtract them, but sometimes CPU times can **decrease** over time (yes):
https://github.com/giampaolo/psutil/blob/dd237b19a4ca286ed7ce9f89b213b2848c9078cf/psutil/__init__.py#L1668-L1680
Should `__sub__` take that into account and trim negative results to zero? I don't know.
Should we support `__add__` other than `__sub__` for consistency? Are there use cases for that? I don't know.
On one hand this seems like a nice to have feature, on the other hand having nameduples with the extra ability of being subtracted feels like API over engineering and a bit too "magical". You'd know that's possible only by reading psutil doc. Plus, at the end of the day, making `nt1 - nt2` possible work would only save one line of code, really. Despite this an interesting idea, I'm more keen on rejecting this proposal, sorry. You had me thinking though. =]

PR #1942, opened 2021-05-02
implemented differences for (p,s)cputimes

Barry1:
## Summary

* OS: { all supported }
* Bug fix: { no }
* Type: { core, new-api }
* Fixes: { #1941  }

## Description

For every implementation of the classes pcputimes and scputimes  I inherited from the "former idea" namedtuple(...) and added 
```python
    def __sub__(self, other):
        return self.__class__(*(self[i] - other[i] for i in range(len(self))))
```
for enabling easy differences between timing information.

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -211,8 +211,9 @@ sfan = namedtuple('sfan', ['label', 'current'])
 # --- for Process methods
 
 # psutil.Process.cpu_times()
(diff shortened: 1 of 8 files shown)


Barry1:
Sorry, was closed automatically from my renaming of the brach. Just opened again unter #1943

PR #1943, opened 2021-05-02
Implemented timing differences

Barry1:
## Summary

* OS: { all supported }
* Bug fix: { no }
* Type: { core, new-api }
* Fixes: { #1941  }

## Description

For every implementation of the classes pcputimes and scputimes  I inherited from the "former idea" namedtuple(...) and added 
```python
    def __sub__(self, other):
        return self.__class__(*(self[i] - other[i] for i in range(len(self))))
```
for enabling easy differences between timing information.

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -211,8 +211,9 @@ sfan = namedtuple('sfan', ['label', 'current'])
 # --- for Process methods
 
 # psutil.Process.cpu_times()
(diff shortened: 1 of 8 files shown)


Barry1:
Happy to see, all test passed - looking forward for merge - open for any questions.

giampaolo:
Rejected as per https://github.com/giampaolo/psutil/issues/1941#issuecomment-831191657.

PR #1947, opened 2021-05-12
add type hints to _psosx.py and _psposix.py

### LINKED ISSUE #1946: Type hints & PEP 561 packaging
## Summary

* OS: n/a
* Type: wheels

## Description

Add [PEP 484](https://www.python.org/dev/peps/pep-0484/) type hints to all (at least public API) code and make the distribution [PEP 561](https://www.python.org/dev/peps/pep-0561/) compliant. This allows `mypy` (and other tools?) to find the type hints and use them in linting.

For PEP 561, add an empty `awscrt/py.typed` file, and include it in the package:

```py
setup(
    package_data={'psutli': ['py.typed']},
    zip_safe=False,  # not needed with wheels, AFAIK
)
```

pawl:
## Summary

* OS: n/a
* Bug fix: no
* Type: tests
* Fixes: #1946 

## Description

This PR makes some initial progress on #1946 by adding type hints as comments (py2-compatible style) to _psosx.py and _psposix.py.

Here's more info on the process I used to generate these type hints: https://gist.github.com/pawl/15ea14d194b48a4000dd6446b9b3f35d

Here are the mypy errors after this change: https://gist.github.com/pawl/ef927fb239f9ec351975dc9c7d320fac

### DIFF
--- a/psutil/_psosx.py
+++ b/psutil/_psosx.py
@@ -9,6 +9,7 @@ import errno
 import functools
 import os
 from collections import namedtuple
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 from . import _common
 from . import _psposix
@@ -110,6 +111,7 @@ pfullmem = namedtuple('pfullmem', pmem._fields + ('uss', ))
 
 
 def virtual_memory():
+    # type: () -> svmem
     """System virtual memory as a namedtuple."""
     total, active, inactive, wired, free, speculative = cext.virtual_mem()
     # This is how Zabbix calculate avail and used mem:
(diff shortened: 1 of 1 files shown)


giampaolo:
Looking back at this: it looks unnecessary to add this and maintain it for every .py files (_pslinux.py, _pswindows.py, etc.). 
We should do this for the main psutil namespace only, meaning [psutil/__init__.py](https://github.com/giampaolo/psutil/blob/master/psutil/__init__.py). 
I'm not sure I like adding this in form of comments though. In #1946 you mention we can use .pyi files. How would that look in practice?

PR #1949, opened 2021-05-12
Fix thread safety of cached functions lock-freely

### LINKED ISSUE #1948: [Linux] Thread-unsafe cache functions
## Summary

* OS: Linux Ubuntu 16.04 LTS
* Architecture: 64bit
* Psutil version: 5.8.0
* Python version: Python 3.9.5
* Type: core

## Description

The cached attributes decorated with the wrapper `memoize_when_activated` are unsafe for multi-threading applications.

Traceback:

```pytb
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_common.py", line 447, in wrapper
    ret = self._cache[fun]
KeyError: <function Process.uids at 0x7f789385d0d0>

  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_common.py", line 447, in wrapper
    ret = self._cache[fun]
KeyError: <function Process._read_status_file at 0x7f7893854790>

  File "/home/panxuehai/.linuxbrew/bin/nvitop", line 8, in <module>
    sys.exit(main())
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/cli.py", line 82, in main
    top.loop()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/top.py", line 199, in loop
    self.redraw()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/top.py", line 190, in redraw
    self.draw()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/top.py", line 182, in draw
    super().draw()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/displayable.py", line 188, in draw
    displayable.draw()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/panels/process.py", line 368, in draw
    if self.selected.owned():
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/ui/panels/process.py", line 80, in owned
    return self.is_set() and (IS_SUPERUSER or self.process.username() == CURRENT_USER)
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/core/process.py", line 52, in wrapped
    return func(self, *args, **kwargs)
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/nvitop/core/process.py", line 225, in username
    return self.host.username()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/__init__.py", line 695, in username
    real_uid = self.uids().real
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_common.py", line 454, in wrapper
    ret = self._cache[fun] = fun(self)
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/__init__.py", line 733, in uids
    return self._proc.uids()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_pslinux.py", line 1576, in wrapper
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_pslinux.py", line 2146, in uids
    data = self._read_status_file()
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_pslinux.py", line 1576, in wrapper
  File "/home/panxuehai/.linuxbrew/opt/python@3.9/lib/python3.9/site-packages/psutil/_common.py", line 454, in wrapper
    ret = self._cache[fun] = fun(self)
AttributeError: _cache
```

Relevant code:

https://github.com/giampaolo/psutil/blob/d5e9ea17e5068d89b741a2f9dc8dd22d24b2545a/psutil/_common.py#L417-L471

A thread lock is needed for `wrapper`, `cache_activate`, and `cache_deactivate`.

XuehaiPan:
## Summary

* OS: Linux Ubuntu 16.04 LTS
* Bug fix: yes
* Type: core
* Fixes: #1948 

## Description

Add thread locks to cached functions, fixes issuse #1948.


giampaolo:
This is definitively needed. I benchmarked it though, and it introduces a 11% slowdown.
Unfortunately that's a lot for this part of the code, because it's called multiple times, for different Process methods, typically in a loop (imagine an app monitoring CPU percentages of **all** processes every second, which is a typical use-case for psutil). 
This would penalize both threaded and non-threaded apps too much. 

I have an idea on how to solve this, but first I would like to find a way to reproduce the original issue that you reported in https://github.com/giampaolo/psutil/issues/1948

I tried this, but it doesn't work:

```python
import psutil, threading


def worker():
    while 1:
        with p.oneshot():
            p.cpu_times()


p = psutil.Process()
for x in range(1000):
    t = threading.Thread(target=worker)
    t.start()
```

XuehaiPan:
> This is definitively needed. I benchmarked it though, and it introduces a 11% slowdown.
> Unfortunately that's a lot for this part of the code, because it's called multiple times, for different Process methods, typically in a loop (imagine an app monitoring CPU percentages of all processes every second, which is a typical use-case for psutil).
> This would penalize both threaded and non-threaded apps too much.

Indeed add a thread lock to these commonly used functions will cause heavy overhead, especially for some apps monitoring all processes (up to ~10k).

I use multi-threading to gather process information asynchronously in my own app [`nvitop`](https://github/XuehaiPan/nvitop), which is to take snapshots for all GPU Processes every 2 seconds in 3 separated threads. There could be multiple threads to manipulate with the same `Process` instance at the same time. My current solution is to add a global snapshot lock to all these threads to ensure there is only one thread gathering information. But I think to make `psutil` a thread-safe library is better.

> I have an idea on how to solve this, but first I would like to find a way to reproduce the original issue that you reported in #1948

If the thread lock introduces too much overhead. I think just fix the `AttributeError` (lock-less) is fine:

```python
    @functools.wraps(fun)
    def wrapper(self):
        try:
            # case 1: we previously entered oneshot() ctx
            ret = self._cache[fun]
        except AttributeError:
            # case 2: we never entered oneshot() ctx
            return fun(self)
        except KeyError:
            # case 3: we entered oneshot() ctx but there's no cache
            # for this entry yet
            ret = fun(self)
            try:
                self._cache[fun] = ret
            except AttributeError:
                pass  # inconsistency caused by multi-threading, just ignore it
            
        return ret
```

> I tried this, but it doesn't work

It is hard to reproduce issue #1948, sometimes it will take hours of running.

giampaolo:
> If the thread lock introduces too much overhead. I think just fix the AttributeError (lock-less) is fine:
> [snippet]

This is exactly what I had in mind: sacrifice caching if we bump in this corner case. If you update your PR I'll gladly merge it.

XuehaiPan:
I have just updated the content and title of this PR.

giampaolo:
Xiexie Xuehai. Merged.

XuehaiPan:
Hi, I find out why the snippet in https://github.com/giampaolo/psutil/pull/1949#issuecomment-932730572 cannot reproduce the race condition in issue #1948. There is a thread lock in `oneshot()`:

https://github.com/giampaolo/psutil/blob/53a6c03e36db43d70b3b4e18abe7b81eaab235ce/psutil/__init__.py#L455-L456

It is thread safe within statement `with proc.oneshot()`.

------

Here I provide two snippets to reproduce issue #1948:

```python
import psutil
import threading


def worker_A(process):
    while True:
        process.cpu_times.cache_activate(process)
        process.cpu_times()
        process.cpu_times.cache_deactivate(process)


def worker_B(process):
    while True:
        process.cpu_times.cache_activate(process)
        # do something faster than worker A
        # delete cache before worker A's uncached function returning
        pass  # do nothing
        process.cpu_times.cache_deactivate(process)


this = psutil.Process()
threading.Thread(target=worker_A, args=(this,)).start()
threading.Thread(target=worker_B, args=(this,)).start()
```

Another snippet to reproduce:

```python
import psutil
import threading
import time
import traceback


def worker_A(process, event):
    while event.is_set():
        with process.oneshot():
            process.cpu_times()
        time.sleep(0.1)

    event.clear()


def worker_B(process, event):
    while event.is_set():
        try:
            # worker B accidentally uses worker A's one-shot cache
            process.cpu_times()
        except Exception:
            # the cache can only deactivated by worker A, so the exception
            # only raises on worker B
            traceback.print_exc()
            break

    event.clear()


this = psutil.Process()
alive = threading.Event()
alive.set()
thread_A = threading.Thread(name='Worker A', target=worker_A, args=(this, alive))
thread_A.start()
thread_B = threading.Thread(name='Worker B', target=worker_B, args=(this, alive))
thread_B.start()

thread_A.join()
thread_B.join()
```

The second snippet could be a common use case, should we add this to the test cases?

giampaolo:
I don't think it's worth adding them as a test case, but thanks for the repro snippets. (I verified they indeed work without patch)

PR #1961, opened 2021-06-23
[Windows] Reset `mp_flag` after each drive on `psutil.disk_partitions()`

mx-psi:
## Summary

* OS: Windows
* Bug fix: yes
* Type: core
* Fixes: #1598

## Description

Fixes #1598; before this only the first drive partitions were gathered, since `mp_flag` was never reset to `TRUE`.

Signed-off-by: Pablo Baeyens <pablo.baeyens@datadoghq.com>

PR #1977, opened 2021-08-09
[Windows] Ignore disk I/O counters for drives that are not accessible

### LINKED ISSUE #1974: [Windows] Ignore permissions errors on `disk_io_counters`
## Summary

* OS: Windows
* Type: core

## Description

On some Windows disk configurations `psutil.disk_io_counters` can raise a `PermissionError: [WinError 5]` due to a `DeviceIoControl` trying to access a disk for which it does not have permissions. To be able to access the rest of the disks info, I see two ways out of this

- (Option 1) Ignore WinError 5 like we do with other errors https://github.com/giampaolo/psutil/blob/a0967043b5819b2edc61d9a12306289d5e7f98c2/psutil/arch/windows/disk.c#L137-L142
- (Option 2) add an off-by-default `ignore_errors` flag to `disk_io_counters` to ignore the errors from `DeviceIOControl` on Windows (and no-op on other platforms).

I am happy to do a PR for either of these but I would love to know what maintainers and contributors think is the best option first. 

cc @giampaolo, could you give me some feedback as to what would be preferable?

mx-psi:
## Summary

* OS: Windows
* Bug fix: yes (debatable?)
* Type: core
* Fixes: #1974

## Description

Ignore permission errors on Windows when fetching I/O counters. I would classify this as a bug fix since this error only happens because of the way the function works (by exhaustively checking every possible drive), and nothing in the function description indicates it should work like this.

Nonetheless I understand it is debatable (see proposed alternative on linked issue if this is not an acceptable patch).

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -761,7 +761,7 @@ D: fix typos in documentation
 
 N: Pablo Baeyens
 W: https://github.com/mx-psi
-I: 1598
+I: 1598, 1974
 
 N: Xuehai Pan
 W: https://github.com/XuehaiPan
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -50,6 +50,7 @@ XXXX-XX-XX
   by PetrPospisil)
 - 1991_: process_iter() can raise TypeError if invoked from multiple threads
   (not thread-safe).
+- 1974_: [Windows] psutil.disk_io_counters() raises permission error.
 
 5.8.0
 =====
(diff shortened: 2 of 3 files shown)


mx-psi:
@giampaolo Would it be possible for you to have a look at this? 

Also, I wonder if I can have a timeline on when this or my previous PR #1961 will be in a released version of psutil. It would be awesome to be able to share a rough date with customers of my employer who need these fixes.

Thanks in advance :)

PR #1978, opened 2021-08-10
Do not fail proc_basic_info if status isn't readable on AIX

albertvaka:
## Summary

* OS: AIX
* Bug fix: No
* Type: Error handling
* Fixes: N/A

## Description

On AIX, `/proc/<pid>/status` might have more restrictive permissions
than `/proc/<pid>/psinfo`. If only the later can be accessed, we can
still return most of the information instead of failing.

### DIFF
--- a/psutil/_psutil_aix.c
+++ b/psutil/_psutil_aix.c
@@ -122,6 +122,7 @@ psutil_proc_basic_info(PyObject *self, PyObject *args) {
         } else {
             // Can't access /proc/<pid>/status (eg: access denied)
             // Continue without the process status
+            PyErr_Clear();
             pr_stat = 0;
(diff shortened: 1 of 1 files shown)


giampaolo:
Historically we don't swallow "access denied" when it happens. We let it propagate. Usually it happens for PIDs owned by other users (root).

albertvaka:
Makes total sense. It's a pity that we fail the entire method just because one field couldn't be returned, but I agree that swallowing the error isn't good either. I'll close this, thanks for your feedback :)

Maybe the best solution would be to split the information coming from two different places in two different methods, so we don't fail when we only need the stuff we can access, but that needs a bigger refactor.

PR #1987, opened 2021-09-12
for building on windows, support blanks in the path of python interpreter

bitranox:
## Summary

* OS: Windows
* Bug fix: yes
* Type: sycripts
* Fixes: support blanks in the path of python interpreter

## Description

support blanks in the path of python interpreter

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -1,23 +1,24 @@
 @echo off
 
-rem ==========================================================================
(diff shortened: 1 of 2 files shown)


bitranox:
paolo, can You please merge that . It does not hurt.
Just to be able to compile on windows, if the python interpreter is sitting on a path with a blank ....
please be so kind
bitranox

PR #1995, opened 2021-10-08
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
## Summary

* OS: Ubuntu
* Bug fix: yes
* Type: tests

## Description

Deprecated unittest aliases were removed in Python 3.11 in https://github.com/python/cpython/pull/28268 . This is the only instance in the codebase as per command below.

```
rg -t py 'assertEquals|assertNotEquals|assertAlmostEquals|assertNotAlmostEquals|assertRegexpMatches|assertNotRegexpMatches|assertRaisesRegexp|failUnlessEqual|failIfEqual|failUnlessAlmostEqual|failUnless|failUnlessRaises|failIf'
```


giampaolo:
Than you :)

PR #2000, opened 2021-10-14
[Linux] convert /dev/root device to real path (fixes #1999)

### LINKED ISSUE #1999: [Linux] psutil.disk_partitions() returns /dev/root on Ubuntu 20.04 on AWS
## Summary

* OS: Ubuntu 20.04
* Architecture: 64bit
* Psutil version: 5.8.0
* Python version: 3.8.10
* Type: core

## Description

On a default AWS instance running Ubuntu 20.04, `psutil.disk_partitions()` returns `/dev/root` instead of `/dev/xvda1`.

This is because `/proc/mounts` contain this entry: 
```
/dev/root / ext4 rw,relatime,discard 0 0
```

However, the actual root disk device is `/dev/xvda1`. Both `mount` and `findmnt` do return the right device:
```
$ mount
/dev/xvda1 on / type ext4 (rw,relatime,discard)
[...]
```
```
$ findmnt -n -o SOURCE /
/dev/xvda1
```

I found some people doing a [workaround](https://github.com/scylladb/scylla/commit/d1ab37654ec065c98b33533d861d676a50ab5380) to call `findmnt` manually in this case, but I think this should be handled by psutil.

giampaolo:
## Summary

* OS: Linux
* Bug fix: no
* Type: core
* Fixes: #1999 

## Description

`disk_partitions()` may return partitions with `device == "/dev/root"` or `"rootfs"`. Add different methods to try to obtain the real device path. Resources:
* https://bootlin.com/blog/find-root-device/
* https://www.systutorials.com/how-to-find-the-disk-where-root-is-on-in-bash-on-linux/

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,8 @@ XXXX-XX-XX
 - 1992_: error classes (NoSuchProcess, AccessDenied, etc.) now have a better
   formatted and separated `__repr__` and `__str__` implementations.
 - 1996_: add support for MidnightBSD.  (patch by Saeed Rasooli)
+- 1999_: [Linux] disk_partitions(): convert "/dev/root" device (an alias used
+  on some Linux distros) to real root device path.
 
 **Bug fixes**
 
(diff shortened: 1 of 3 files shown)


albertvaka:
I can confirm this fixes the issue for me 😄 

Is there a date for a future release that includes this change?

PR #1958, opened 2021-06-06
convert mach tick units to nsecs

rmalouf:
## Summary

* OS: macOS 11.4
* Bug fix: yes
* Type: core
* Fixes: #1956, nicolargo/glances#1827

## Description

{{{
Mach sys calls report cpu time in "tick units" -- on intel cpus, these happen to be nsecs but on M1 machines they're about 40 nsecs. This PR corrects for this, but I think it still needs two improvements: the value of mach_timebase_info should be cached somewhere, and there are probably other places where the code depends on 1 tick=1 nsec that also need to be updated. Sounds reasonable? I can work on both of those, but I wanted to get some feedback first before I spend any more time on this.
}}}

PR #2011, opened 2021-11-04
Timebase

### LINKED ISSUE #1956: [macOS / M1] cpu_percent() gives incorrect information
## Summary

* OS: macOS 11.4
* Architecture: ARM
* Psutil version: 5.8.0
* Python version: 3.9.4
* Type: core

## Description

Process.cpu_percent() always reports very low cpu usage, in contradiction to top and Activity monitor. This test program:

```
from concurrent.futures import ThreadPoolExecutor
import psutil
import time
from math import log

def work(n):
    m = 0
    for i in range(1, n**2):
        m += log(i)
    return True
    
executor = ThreadPoolExecutor(max_workers=2)
task = executor.submit(work, 10000)

proc = psutil.Process()
while not task.done():
    print(proc.cpu_percent(interval=1))
```

consistently reports 2.4% usage on the Mac (vs. ~100% usage, as expected, on linux with the same version of psutil).

odormond:
## Summary

* OS: macOS 12.0.1, 11.6.1
* Bug fix: yes
* Type: core
* Fixes: #1956

## Description

Mach sys calls report cpu time in "tick units" -- on intel cpus, these happen to be nsecs but on M1 machines they're about 40 nsecs. This PR corrects for this, but I think it still needs two improvements: the value of mach_timebase_info should be cached somewhere, and there are probably other places where the code depends on 1 tick=1 nsec that also need to be updated. Sounds reasonable? I can work on both of those, but I wanted to get some feedback first before I spend any more time on this.

Based on @rmalouf PR #1958.

### DIFF
--- a/psutil/_psutil_common.c
+++ b/psutil/_psutil_common.c
@@ -170,6 +170,15 @@ convert_kvm_err(const char *syscall, char *errbuf) {
 }
 #endif
 
+// ====================================================================
+// --- macOS
+// ====================================================================
+
+#ifdef PSUTIL_OSX
+#include <mach/mach_time.h>
+
+struct mach_timebase_info MACH_TIMEBASE_INFO;
+#endif
 
 // ====================================================================
 // --- Windows
@@ -403,6 +412,10 @@ psutil_setup(void) {
         return 1;
     GetSystemInfo(&PSUTIL_SYSTEM_INFO);
     InitializeCriticalSection(&PSUTIL_CRITICAL_SECTION);
+#endif
+
+#ifdef PSUTIL_OSX
+    mach_timebase_info(&MACH_TIMEBASE_INFO);
 #endif
 
     return 0;
--- a/psutil/_psutil_common.c
+++ b/psutil/_psutil_common.c
@@ -177,7 +177,7 @@ convert_kvm_err(const char *syscall, char *errbuf) {
 #ifdef PSUTIL_OSX
 #include <mach/mach_time.h>
 
-struct mach_timebase_info MACH_TIMEBASE_INFO;
+struct mach_timebase_info PSUTIL_MACH_TIMEBASE_INFO;
 #endif
 
 // ====================================================================
@@ -415,7 +415,7 @@ psutil_setup(void) {
 #endif
 
 #ifdef PSUTIL_OSX
-    mach_timebase_info(&MACH_TIMEBASE_INFO);
+    mach_timebase_info(&PSUTIL_MACH_TIMEBASE_INFO);
 #endif
 
     return 0;
--- a/psutil/_psutil_osx.c
+++ b/psutil/_psutil_osx.c
@@ -279,17 +279,18 @@ static PyObject *
 psutil_proc_pidtaskinfo_oneshot(PyObject *self, PyObject *args) {
     pid_t pid;
     struct proc_taskinfo pti;
-    uint64_t total_user, total_system;
+    uint64_t total_user;
+    uint64_t total_system;
 
     if (! PyArg_ParseTuple(args, _Py_PARSE_PID, &pid))
         return NULL;
     if (psutil_proc_pidinfo(pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti)) <= 0)
         return NULL;
 
-    total_user = pti.pti_total_user * MACH_TIMEBASE_INFO.numer;
-    total_user /= MACH_TIMEBASE_INFO.denom;
-    total_system = pti.pti_total_system * MACH_TIMEBASE_INFO.numer;
-    total_system /= MACH_TIMEBASE_INFO.denom;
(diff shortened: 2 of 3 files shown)


giampaolo:
>  the value of mach_timebase_info should be cached somewhere

Makes sense. You can do that in `psutil_setup()` function, which is called on module init:
https://github.com/giampaolo/psutil/blob/8fa1746685a86b8ddda454e2e2746d535e2b6d89/psutil/_psutil_common.c#L395
You should define the new global var as a C `extern` defined in `_psutil_common.h`.  grep for `PSUTIL_SYSTEM_INFO` to see how it's done.

psutil/_psutil_osx.c
```diff
@@ -279,16 +280,24 @@ static PyObject *
...
+    total_user /= info.denom;
+    total_system = pti.pti_total_system * info.numer;
+    total_system /= info.denom;
+
     return Py_BuildValue(
         "(ddKKkkkk)",
-        (float)pti.pti_total_user / 1000000000.0,     // (float) cpu user time
-        (float)pti.pti_total_system / 1000000000.0,   // (float) cpu sys time
+        (float)total_user / 1000000000.0,     // (float) cpu user time
+        (float)total_system / 1000000000.0,   // (float) cpu sys time
```
giampaolo, line 300: Since you did the new conversion above, is the `/ 1000000000.0` division still necessary?

odormond:
> > the value of mach_timebase_info should be cached somewhere
> 
> Makes sense. You can do that in `psutil_setup()` function, which is called on module init:
> 
> https://github.com/giampaolo/psutil/blob/8fa1746685a86b8ddda454e2e2746d535e2b6d89/psutil/_psutil_common.c#L395
> 
> You should define the new global var as a C `extern` defined in `_psutil_common.h`. grep for `PSUTIL_SYSTEM_INFO` to see how it's done.

Sure. I just pushed another commit doing just that. If it's good for you I can squash it to replace the previous two.

About the division by a billion, the mach_timebase_info give the conversion ratio from ticks to nanoseconds so the result still needs to be converted to seconds.

psutil/_psutil_common.c
```diff
@@ -170,6 +170,15 @@ convert_kvm_err(const char *syscall, char *errbuf) {
...
 
+// ====================================================================
+// --- macOS
+// ====================================================================
+
+#ifdef PSUTIL_OSX
+#include <mach/mach_time.h>
+
+struct mach_timebase_info MACH_TIMEBASE_INFO;
+#endif
```
giampaolo, line 181: In order to be consistent with `PSUTIL_SYSTEM_INFO`, `PSUTIL_CRITICAL_SECTION` and `PSUTIL_DEBUG` I suggest to:

* call this `PSUTIL_MACH_TIMEBASE_INFO`
* define it in `_psutil_common.h` file

psutil/_psutil_common.c
```diff
@@ -405,5 +414,9 @@ psutil_setup(void) {
     InitializeCriticalSection(&PSUTIL_CRITICAL_SECTION);
 #endif
 
+#ifdef PSUTIL_OSX
+    mach_timebase_info(&MACH_TIMEBASE_INFO);
+#endif
```
giampaolo, line 419: I would invoke this from `psutil_setup()` function instead, so that all init functions are in the same place.

psutil/_psutil_osx.c
```diff
@@ -279,16 +279,22 @@ static PyObject *
 psutil_proc_pidtaskinfo_oneshot(PyObject *self, PyObject *args) {
     pid_t pid;
     struct proc_taskinfo pti;
+    uint64_t total_user, total_system;
```
giampaolo, line 282: one declaration per line

giampaolo (changes requested):

psutil/_psutil_common.c
odormond: > In order to be consistent with `PSUTIL_SYSTEM_INFO`, `PSUTIL_CRITICAL_SECTION` and `PSUTIL_DEBUG` I suggest to:
> 
>     * call this `PSUTIL_MACH_TIMEBASE_INFO`

That makes more sense indeed. Sorry.

>     * define it in `_psutil_common.h` file

I don't understand this. The header file does have the extern definition and storage can only be in the .c file. What am I missing?

psutil/_psutil_common.c
odormond: It is. Do you prefer that I move it before the Windows conditional block?

psutil/_psutil_osx.c
odormond: Yes. The mach timebase conversion ratio produces nanoseconds.

psutil/_psutil_common.c
giampaolo: Oh didn't realize you were already in psutil_setup() function, sorry

giampaolo:
Merged. Thanks!

PR #2019, opened 2021-11-19
Handle missing dependencies on MidnightBSD

laffer1:
## Summary

* OS: MidnightBSD 2.1.2
* Bug fix: no
* Type: scripts
* Fixes: 

## Description

MidnightBSD uses a different package manager than FreeBSD, but we're grouping the support together so just use which to figure out the correct package manager with FreeBSD's first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -424,7 +424,10 @@ def main():
                 print(hilite("XCode (https://developer.apple.com/xcode/) "
                              "is not installed"), color="red", file=sys.stderr)
             elif FREEBSD:
-                missdeps("pkg install gcc python%s" % py3)
+                if which('pkg'):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -424,7 +424,10 @@ def main():
                 print(hilite("XCode (https://developer.apple.com/xcode/) "
                              "is not installed"), color="red", file=sys.stderr)
             elif FREEBSD:
-                missdeps("pkg install gcc python%s" % py3)
+                if which('pkg'):
+                    missdeps("pkg install gcc python%s" % py3)
+                elif which('mport'):
```
giampaolo, line 429: Please add:

```
elif which('mport'):  # MidnightBSD
```

laffer1: added

PR #2037, opened 2021-12-20
Add in support for network interface flags.

clalancette:
## Summary

* OS: Linux, macOS
* Bug fix: no
* Type: add more information to net_if_addrs()

## Description

Add in support for network interface flags.

That is, return the raw flag integer as reported by getifaddrs()
on POSIX-compatible systems.  On Windows systems (and any other
ones that do not support flags), return None instead.

I was careful to add the `flags` field to the end of the `snicaddr` named tuple so that the API wouldn't be broken.

Full disclosure: I've tested this on Linux and on macOS, and it seems to work OK on both of those.  I have not tested on Windows, since I don't have a Windows machine handy.

I've made a few choices here that we could change:
1. I've decided to add this additional functionality to `net_if_addrs()` as opposed to `net_if_stats()` or adding a new API.  That makes the most sense to me, but I'm happy to change it if you feel otherwise.
2. The flags are just the raw integer that comes from the return value of `getifaddrs()` on POSIX.  The meaning of each of the bits is standardized by POSIX, so it should be the same on all platforms that support this.  That said, we could make this "nicer" and possibly more Pythonic by expanding those bitfields into human-readable flags.
3. I decided to return "None" when the flags aren't supported (like on Windows).  That seemed like a good way to go here, but if we wanted to make sure it was always an integer we could return -1 or something on Windows.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -735,7 +735,12 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
-  - **flags**: a string of comma-separate flags on the interface (may be ``None``).
+  - **flags**: a string of comma-separated flags on the interface (may be the empty string).
+    Possible flags are: ``up``, ``broadcast``, ``debug``, ``loopback``,
+    ``pointopoint``, ``notrailers``, ``running``, ``noarp``, ``promisc``,
+    ``allmulti``, ``master``, ``slave``, ``multicast``, ``portsel``,
+    ``dynamic``, ``oactive``, ``simplex``, ``link0``, ``link1``, ``link2``,
+    and ``d2`` (some flags are only available on certain platforms).
 
   Example:
 
@@ -750,7 +755,7 @@ Network
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
-  .. versionchanged:: 5.9.1 *flags* field was added.
+  .. versionchanged:: 5.9.1 *flags* field was added on POSIX.
 
 Sensors
 -------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -742,6 +742,8 @@ Network
     ``dynamic``, ``oactive``, ``simplex``, ``link0``, ``link1``, ``link2``,
     and ``d2`` (some flags are only available on certain platforms).
 
+    Availability: UNIX
+
   Example:
 
     >>> import psutil
@@ -755,7 +757,7 @@ Network
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
-  .. versionchanged:: 5.9.2 *flags* field was added on POSIX.
+  .. versionchanged:: 5.9.3 *flags* field was added on POSIX.
 
 Sensors
 -------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -735,19 +735,22 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
+  - **flags**: a string of comma-separate flags on the interface (may be ``None``).
 
   Example:
 
     >>> import psutil
     >>> psutil.net_if_stats()
-    {'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500),
-     'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536)}
+    {'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500, flags='up,broadcast,running,multicast'),
+     'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536, flags='up,loopback,running')}
 
   Also see `nettop.py`_ and `ifconfig.py`_ for an example application.
 
   .. versionadded:: 3.0.0
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
+
+  .. versionchanged:: 5.9.1 *flags* field was added.
 
 Sensors
 -------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -735,7 +735,7 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
-  - **flags**: a string of comma-separated flags on the interface (may be the empty string).
+  - **flags**: a string of comma-separated flags on the interface (may be an empty string).
     Possible flags are: ``up``, ``broadcast``, ``debug``, ``loopback``,
     ``pointopoint``, ``notrailers``, ``running``, ``noarp``, ``promisc``,
     ``allmulti``, ``master``, ``slave``, ``multicast``, ``portsel``,
@@ -755,7 +755,7 @@ Network
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
-  .. versionchanged:: 5.9.1 *flags* field was added on POSIX.
+  .. versionchanged:: 5.9.2 *flags* field was added on POSIX.
 
 Sensors
 -------
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -381,7 +381,7 @@ def net_if_stats():
     for name in names:
         try:
             mtu = cext_posix.net_if_mtu(name)
-            isup = cext_posix.net_if_is_running(name)
+            flags = cext_posix.net_if_flags(name)
             duplex, speed = cext_posix.net_if_duplex_speed(name)
         except OSError as err:
             # https://github.com/giampaolo/psutil/issues/1279
@@ -390,7 +390,14 @@ def net_if_stats():
         else:
             if hasattr(_common, 'NicDuplex'):
                 duplex = _common.NicDuplex(duplex)
-            ret[name] = _common.snicstats(isup, duplex, speed, mtu)
+            flag_list = []
+            for flagname, bit in _psposix.POSIX_NET_FLAGS:
+                if flags & (1 << bit):
+                    flag_list.append(flagname)
+
+            output_flags = ','.join(flag_list)
+            isup = 'running' in output_flags
+            ret[name] = _common.snicstats(isup, duplex, speed, mtu, output_flags)
     return ret
 
 
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -390,13 +390,8 @@ def net_if_stats():
         else:
             if hasattr(_common, 'NicDuplex'):
                 duplex = _common.NicDuplex(duplex)
-            flag_list = []
-            for flagname, bit in _psposix.POSIX_NET_FLAGS:
-                if flags & (1 << bit):
-                    flag_list.append(flagname)
-
-            output_flags = ','.join(flag_list)
-            isup = 'running' in output_flags
+            output_flags = ','.join(flags)
+            isup = 'running' in flags
             ret[name] = _common.snicstats(isup, duplex, speed, mtu, output_flags)
     return ret
 
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -392,7 +392,8 @@ def net_if_stats():
                 duplex = _common.NicDuplex(duplex)
             output_flags = ','.join(flags)
             isup = 'running' in flags
-            ret[name] = _common.snicstats(isup, duplex, speed, mtu, output_flags)
+            ret[name] = _common.snicstats(isup, duplex, speed, mtu,
+                                          output_flags)
     return ret
 
 
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1071,7 +1071,8 @@ def net_if_stats():
         else:
             output_flags = ','.join(flags)
             isup = 'running' in flags
-            ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu, output_flags)
+            ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu,
+                                          output_flags)
     return ret
 
 
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1070,7 +1070,7 @@ def net_if_stats():
                 debug(err)
         else:
             output_flags = ','.join(flags)
-            isup = 'running' in output_flags
+            isup = 'running' in flags
             ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu, output_flags)
     return ret
 
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1060,7 +1060,7 @@ def net_if_stats():
     for name in names:
         try:
             mtu = cext_posix.net_if_mtu(name)
-            isup = cext_posix.net_if_is_running(name)
+            flags = cext_posix.net_if_flags(name)
             duplex, speed = cext.net_if_duplex_speed(name)
         except OSError as err:
             # https://github.com/giampaolo/psutil/issues/1279
@@ -1069,7 +1069,9 @@ def net_if_stats():
             else:
                 debug(err)
         else:
-            ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu)
+            output_flags = ','.join(flags)
+            isup = 'running' in output_flags
+            ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu, output_flags)
     return ret
 
 
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -293,7 +293,7 @@ def net_if_stats():
         isup, duplex, speed, mtu = items
         if hasattr(_common, 'NicDuplex'):
             duplex = _common.NicDuplex(duplex)
-        ret[name] = _common.snicstats(isup, duplex, speed, mtu)
+        ret[name] = _common.snicstats(isup, duplex, speed, mtu, None)
     return ret
 
 
--- a/psutil/_pssunos.py
+++ b/psutil/_pssunos.py
@@ -293,7 +293,7 @@ def net_if_stats():
         isup, duplex, speed, mtu = items
         if hasattr(_common, 'NicDuplex'):
             duplex = _common.NicDuplex(duplex)
-        ret[name] = _common.snicstats(isup, duplex, speed, mtu, None)
+        ret[name] = _common.snicstats(isup, duplex, speed, mtu, '')
     return ret
 
 
--- a/psutil/_psutil_posix.c
+++ b/psutil/_psutil_posix.c
@@ -430,23 +430,21 @@ error:
     return PyErr_SetFromErrno(PyExc_OSError);
 }
 
-static bool
-check_and_append_iff_flag(PyObject *py_retlist, short int flags, short int flag_to_check, const char * flag_name)
+static int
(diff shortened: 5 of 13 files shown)


giampaolo:
These flags are pretty pointless since there are no exposed constants to compare them to (`IFF_BROADCAST`, `IFF_LOOPBACK`, etc). A solution to this could be turning them into strings, like:

```
flags="multicast,loopback,running"
```

...similarly to `psutil.disk_partitions()`'s `opts` field. According to the man page, the flags are:

```
       SIOCGIFFLAGS, SIOCSIFFLAGS
              Get or set the active flag word of the device.  
              ifr_flags contains a bit mask of the following values:

                                      Device flags
              IFF_UP            Interface is running.
              IFF_BROADCAST     Valid broadcast address set.
              IFF_DEBUG         Internal debugging flag.
              IFF_LOOPBACK      Interface is a loopback interface.
              IFF_POINTOPOINT   Interface is a point-to-point link.
              IFF_RUNNING       Resources allocated.
              IFF_NOARP         No arp protocol, L2 destination address not
                                set.
              IFF_PROMISC       Interface is in promiscuous mode.
              IFF_NOTRAILERS    Avoid use of trailers.
              IFF_ALLMULTI      Receive all multicast packets.

              IFF_MASTER        Master of a load balancing bundle.
              IFF_SLAVE         Slave of a load balancing bundle.
              IFF_MULTICAST     Supports multicast
              IFF_PORTSEL       Is able to select media type via ifmap.
              IFF_AUTOMEDIA     Auto media selection active.
              IFF_DYNAMIC       The addresses are lost when the interface
                                goes down.
              IFF_LOWER_UP      Driver signals L1 up (since Linux 2.6.17)
              IFF_DORMANT       Driver signals dormant (since Linux 2.6.17)
              IFF_ECHO          Echo sent packets (since Linux 2.6.25)
```

What makes me skeptical about this is that most of the flags above seem to refer to the network interface itself, and not the individual addresses. E.g. associating "IFF_PROMISC", "IFF_PORTSEL" or "IFF_AUTOMEDIA" to a network address it would not make too much sense: those are things that belong to a network interface.

As such, perhaps the right place to do this is `net_if_stats()` and not `net_if_addrs()`:
https://psutil.readthedocs.io/en/latest/#psutil.net_if_stats

clalancette:
> flags="multicast,loopback,running"

That's fine by me.  I'll go ahead and implement it that way instead.

> As such, perhaps the right place to do this is `net_if_stats()` and not `net_if_addrs()`:

OK, that sounds good to me as well.  I'll add a new extension API for getting the flags and add it to the `snicstats` API instead.

clalancette:
And this has all been done in b75a234.  Please take a look when you get a chance.

clalancette:
I've now rebased this onto the latest, and resolved the conflicts in HISTORY.rst based on the release.  This is ready for another review when you have time.

clalancette:
Rebased onto the latest again.  Please take a look.

clalancette:
Friendly ping on this one; do you think we could get this in?

giampaolo:
Hi Chris and sorry for being late but life got busier. =) I re-read your patch and I'm happy to include this functionality.  I don't like hard-coding C constants in python though. I prefer if that is done in C. I remember we do something like this in here:
https://github.com/giampaolo/psutil/blob/df3fba560125a1694140418458d85076ebc546fd/psutil/_psutil_bsd.c#L730
I guess you can do something like that and remove `","` at the beginning/end of the resulting string if needed.

clalancette:
> I re-read your patch and I'm happy to include this functionality. I don't like hard-coding C constants in python though. I prefer if that is done in C.

Thanks for the feedback.  I went through and rewrote most of it in C, but in a slightly different way (see 1a5842b).  In particular, what I'm doing there is constructing a list of strings to return to the Python layer, which I can then easily combine using `join`.  Please take a look and let me know what you think.

By the way, it turns out that it is much, much better to do this at the C layer.  The `IFF_` bit flags are not standardized, and so different platforms can report different values there (for instance, on Linux IFF_MULTICAST is 0x1000, while on macOS it is 0x8000).  So this is absolutely the right move, thanks for noticing it.

psutil/_pssunos.py
```diff
@@ -293,7 +293,7 @@ def net_if_stats():
         isup, duplex, speed, mtu = items
         if hasattr(_common, 'NicDuplex'):
             duplex = _common.NicDuplex(duplex)
-        ret[name] = _common.snicstats(isup, duplex, speed, mtu)
+        ret[name] = _common.snicstats(isup, duplex, speed, mtu, None)
```
giampaolo, line 296: let's enable this on sunos as well (see my previous comment)

giampaolo (changes requested):

docs/index.rst
```diff
@@ -735,20 +735,23 @@ Network
...
+    {'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500, flags='up,broadcast,running,multicast'),
+     'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536, flags='up,loopback,running')}
 
   Also see `nettop.py`_ and `ifconfig.py`_ for an example application.
 
   .. versionadded:: 3.0.0
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
+  .. versionchanged:: 5.9.1 *flags* field was added.
```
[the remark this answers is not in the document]
clalancette: Fixed in e05f926.

docs/index.rst
```diff
@@ -735,20 +735,23 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
+  - **flags**: a string of comma-separate flags on the interface (may be ``None``).
```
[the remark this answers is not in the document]
clalancette: Changed to an empty string everywhere below.  Also listed all of the possible string values in e05f926.

psutil/_psbsd.py
```diff
@@ -390,7 +390,14 @@ def net_if_stats():
         else:
             if hasattr(_common, 'NicDuplex'):
                 duplex = _common.NicDuplex(duplex)
-            ret[name] = _common.snicstats(isup, duplex, speed, mtu)
+            flag_list = []
+            for flagname, bit in _psposix.POSIX_NET_FLAGS:
+                if flags & (1 << bit):
+                    flag_list.append(flagname)
```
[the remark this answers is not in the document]
clalancette: Yep, good call.  Removed in e05f926.

psutil/_psutil_posix.c
```diff
@@ -430,6 +430,106 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+
+    if (py_retlist == NULL)
+        return NULL;
+
+    if (! PyArg_ParseTuple(args, "s", &nic_name))
+        return NULL;
+
+    sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1)
+        return NULL;
```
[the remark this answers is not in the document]
clalancette: Fixed in e05f926.

psutil/_psutil_posix.c
```diff
@@ -430,6 +430,106 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+ * Get all of the NIC flags and return them.
+ */
+static PyObject *
+psutil_net_if_flags(PyObject *self, PyObject *args) {
+    char *nic_name;
+    int sock = -1;
+    int ret;
+    struct ifreq ifr;
+    PyObject *py_retlist = PyList_New(0);
+    PyObject *py_flag = NULL;
```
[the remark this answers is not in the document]
clalancette: Fixed in e05f926.

psutil/_psutil_posix.c
```diff
@@ -430,6 +430,106 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+    short int flags;
+
+    if (py_retlist == NULL)
+        return NULL;
+
+    if (! PyArg_ParseTuple(args, "s", &nic_name))
+        return NULL;
```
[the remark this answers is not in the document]
clalancette: Fixed in e05f926.

psutil/_psutil_posix.c
```diff
@@ -430,6 +430,106 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+    ret = ioctl(sock, SIOCGIFFLAGS, &ifr);
+    if (ret == -1)
+        goto error;
+
+    close(sock);
+    sock = -1;
+
+    flags = ifr.ifr_flags & 0xFFFF;
+
+    if (flags & IFF_UP) {
```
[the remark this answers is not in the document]
clalancette: Yes, this is a fantastic idea.  It also allows us to specify additional flags that only occur on certain platforms.  Changed in e05f926.

psutil/_psutil_posix.c
```diff
@@ -430,6 +430,106 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+        py_flag = PyUnicode_DecodeFSDefault("up");
```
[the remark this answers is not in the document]
clalancette: Yes, good call.  I added a helper function for this in e05f926, which I think cleans up the code nicely.

psutil/_pswindows.py
```diff
@@ -386,7 +386,7 @@ def net_if_stats():
         isup, duplex, speed, mtu = items
         if hasattr(_common, 'NicDuplex'):
             duplex = _common.NicDuplex(duplex)
-        ret[name] = _common.snicstats(isup, duplex, speed, mtu)
+        ret[name] = _common.snicstats(isup, duplex, speed, mtu, None)
```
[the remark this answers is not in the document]
clalancette: Fixed in e05f926.

psutil/_psbsd.py
```diff
@@ -390,7 +390,14 @@ def net_if_stats():
...
+
+            output_flags = ','.join(flag_list)
```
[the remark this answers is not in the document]
clalancette: So I kept this as `output_flags` currently, as we still need to use `flags` for the list of flags.  I'm happy to change the name of both if you want, just let me know what you would prefer.

psutil/_pslinux.py
```diff
@@ -1069,7 +1069,9 @@ def net_if_stats():
             else:
                 debug(err)
         else:
-            ret[name] = _common.snicstats(isup, duplex_map[duplex], speed, mtu)
+            output_flags = ','.join(flags)
```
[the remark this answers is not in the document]
clalancette: Same comment here.

psutil/_pssunos.py
clalancette: So SunOS is a completely different beast than Linux, BSD, macOS, or Windows.  Looking at the code in https://github.com/giampaolo/psutil/blob/4446e3bd5fd29e0598f2ca6672d5ffe8ca40d442/psutil/_psutil_sunos.c#L1484 , it looks like it doesn't even use the same `ioctl` (it is using something called SIOCGLIFFLAGS), and then it verifies that data with another structure (kstat?).  So I don't think I can easily enable SunOS, and I don't have access to a SunOS machine.

clalancette:
@giampaolo Thanks for the very detailed review!  I've made a number of changes based on your review, though I opted not to make some of the changes you requested (individual comments have details why).  Please take another look!

docs/index.rst
```diff
@@ -735,20 +735,28 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
+  - **flags**: a string of comma-separated flags on the interface (may be the empty string).
```
giampaolo, line 738: "may be AN empty string"

docs/index.rst
```diff
@@ -735,20 +735,28 @@ Network
...
+    {'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500, flags='up,broadcast,running,multicast'),
+     'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536, flags='up,loopback,running')}
 
   Also see `nettop.py`_ and `ifconfig.py`_ for an example application.
 
   .. versionadded:: 3.0.0
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
+  .. versionchanged:: 5.9.1 *flags* field was added on POSIX.
```
giampaolo, line 758: 5.9.2

psutil/_psutil_posix.c
```diff
@@ -429,6 +430,186 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+            return false;
+        if (PyList_Append(py_retlist, py_str)) {
+            Py_DECREF(py_str);
+            return false;
+        }
+        Py_CLEAR(py_str);
+    }
+
+    return true;
+}
```
giampaolo, line 450: I don't like doing the flag comparison / check in here (`if flags & flag_to_check ...`). That should happen in the main function. The utility function (this one) should only add the item to the dict.

Also, instead of return `true` or `false` I would just return `1` or `0`_, which is consistent with the rest of the code base and avoids including `<stdbool.h>` (I have some memories of this header file causing problems on some platform).

psutil/_psutil_posix.c
```diff
@@ -429,6 +430,186 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+    sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1) {
+        PyErr_SetFromErrno(PyExc_OSError);
+        goto error;
+    }
+
+    PSUTIL_STRNCPY(ifr.ifr_name, nic_name, sizeof(ifr.ifr_name));
+    ret = ioctl(sock, SIOCGIFFLAGS, &ifr);
+    if (ret == -1) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
giampaolo, line 479: Please use `PyErr_SetFromOSErrnoWithSyscall("ioctl(SIOCGIFFLAGS)");`.

psutil/_psutil_posix.c
```diff
@@ -429,6 +430,186 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+
+    close(sock);
+    sock = -1;
+
+    flags = ifr.ifr_flags & 0xFFFF;
+
+    // Linux/glibc IFF flags: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/net/if.h;h=251418f82331c0426e58707fe4473d454893b132;hb=HEAD
+    // macOS IFF flags: https://opensource.apple.com/source/xnu/xnu-792/bsd/net/if.h.auto.html
+    // AIX IFF flags: https://www.ibm.com/support/pages/how-hexadecimal-flags-displayed-ifconfig-are-calculated
+    // FreeBSD IFF flags: https://www.freebsd.org/cgi/man.cgi?query=if_allmulti&apropos=0&sektion=0&manpath=FreeBSD+10-current&format=html
```
giampaolo, line 491: very good

giampaolo (changes requested):

docs/index.rst
clalancette: Fixed in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

docs/index.rst
clalancette: Fixed in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

psutil/_psutil_posix.c
clalancette: That's fair.  I've switched this helper function to return an int, and I'm now doing the check in the main function.  All in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

psutil/_psutil_posix.c
```diff
@@ -429,6 +430,186 @@ psutil_net_if_mtu(PyObject *self, PyObject *args) {
...
+
+    if (py_retlist == NULL)
+        return NULL;
+
+    if (! PyArg_ParseTuple(args, "s", &nic_name))
+        goto error;
+
+    sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1) {
+        PyErr_SetFromErrno(PyExc_OSError);
```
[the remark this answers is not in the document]
clalancette: No problem, it happens.  Fixed in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

psutil/_psutil_posix.c
clalancette: Fixed in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

psutil/tests/test_system.py
```diff
@@ -808,7 +808,7 @@ def test_net_if_stats(self):
                         psutil.NIC_DUPLEX_UNKNOWN)
         for name, stats in nics.items():
             self.assertIsInstance(name, str)
-            isup, duplex, speed, mtu = stats
+            isup, duplex, speed, mtu, flags = stats
```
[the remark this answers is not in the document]
clalancette: Fixed in d83c959a981c68ff7ff3d60a155b4f35640e9fa9.

clalancette:
Sorry for the delay here (I was on vacation).  I've now gone through and fixed up all of the latest issues; please take another look!

docs/index.rst
```diff
@@ -735,20 +735,28 @@ Network
   - **speed**: the NIC speed expressed in mega bits (MB), if it can't be
     determined (e.g. 'localhost') it will be set to ``0``.
   - **mtu**: NIC's maximum transmission unit expressed in bytes.
+  - **flags**: a string of comma-separated flags on the interface (may be an empty string).
+    Possible flags are: ``up``, ``broadcast``, ``debug``, ``loopback``,
+    ``pointopoint``, ``notrailers``, ``running``, ``noarp``, ``promisc``,
+    ``allmulti``, ``master``, ``slave``, ``multicast``, ``portsel``,
+    ``dynamic``, ``oactive``, ``simplex``, ``link0``, ``link1``, ``link2``,
+    and ``d2`` (some flags are only available on certain platforms).
```
giampaolo, line 743: Please add `. Availability: UNIX`. After merge I'll try to take a look on how feasible it is to do this on Windows.

docs/index.rst
```diff
@@ -735,20 +735,28 @@ Network
...
+    {'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500, flags='up,broadcast,running,multicast'),
+     'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536, flags='up,loopback,running')}
 
   Also see `nettop.py`_ and `ifconfig.py`_ for an example application.
 
   .. versionadded:: 3.0.0
 
   .. versionchanged:: 5.7.3 `isup` on UNIX also checks whether the NIC is running.
 
+  .. versionchanged:: 5.9.2 *flags* field was added on POSIX.
```
giampaolo, line 758: I released 5.9.2 yesterday. Please set this to 5.9.3.

psutil/_psutil_posix.c
```diff
@@ -9,6 +9,7 @@
 #include <Python.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <stdbool.h>
```
giampaolo, line 12: I think you can remove this, right?

giampaolo:
I added some final comments but overall it's good to go.

docs/index.rst
clalancette: I've added it in 30825261446cadb71a7201e154bb40feee9eccd4 .  Let me know if that is where you want it.

docs/index.rst
clalancette: Fixed in 30825261446cadb71a7201e154bb40feee9eccd4

psutil/_psutil_posix.c
clalancette: Good call, removed in 30825261446cadb71a7201e154bb40feee9eccd4

clalancette:
> I added some final comments but overall it's good to go.

Thanks!  Besides your comments, I also made small changes in 30825261446cadb71a7201e154bb40feee9eccd4 that should hopefully make linting happier.

giampaolo:
Perfect. I'm happy with this change. Thanks.

clalancette:
Thank you so much!  I appreciate all of the good feedback and the merge of the PR.

PR #2046, opened 2022-01-04
Multiply cpuinfo reported frequency by 1000

MadnessASAP:
## Summary

* OS: { Linux }
* Bug fix: { yes }
* Type: { core }
* Fixes: { None }

## Description

proc/cpuinfo reports frequencies in MHz, sysfs in kHz. Currently psutil assumes kHz across the board and as such reports the wrong frequency when using the cpuinfo cache. To fix this I've multiplied the cpuinfo reported number by a   1000 so it matched the sysfs number.


MadnessASAP:
This was fixed by #2049

Issue #1989, opened 2021-09-18
[Linux] Process.cpu_percent is incorrect when instantiated with a tid (thread ID)


benhsmith:
## Summary

* OS: { Linux }
* Architecture: { 64bit, 32bit, ARM, PowerPC, s390 }
* Psutil version: { 5.8.0 }
* Python version: { 3.6.8 }
* Type: { core }

## Description

Assume I have a process with pid 100 and 2 threads with tids 101 and 102.

```
p = Process(101) 
p.cpu_percent(1)
```

I would expect `cpu_percent` to return the CPU usage of thread 101 but, instead, it returns thread usage for the entire process (100). Process.threads() does return user and system time, and I can get cpu percent from those but then I have to calculate cpu percent usage in my own code, which is what `cpu_percent` is for.

Ideally, when the process is a thread, `_pslinux._pslinux._parse_stat_file` would figure out that its pid is really a tid and then read the proc/PID/task/TID/stat file (which contains per-thread CPU times) instead of /proc/TID/stat.

Unfortunately, for some reason tgid, which is what would tell you the PID of the thread, is not in /proc/PID/stat, it's only in /proc/PID/status. So it looks like figuring out that a PID is actually a TID would require also reading /proc/PID/status to get the tgid value. So the downsides to this approach would be reading an additional file and a slight change to behavior. If anyone was counting on getting CPU percentage for the entire process when they create a Process from a tid, this would break that. However, being able to easily get per thread CPU usage seems much more useful.

I'd be happy to put together a PR but I wanted to check first in case this has already been considered and rejected. I looked in the issues and didn't find anything.

benhsmith:
I was thinking about this again and it occurred to me another way to provide pre-thread CPU usage would be to just make the objects returned by Process.threads() full Process objects. Then you could do something like:
```p = Process(101)
for thread in p.threads():
    print(f'[thread.name()]: [thread.cpu_percent()]')
```
that makes a bit more sense since it corresponds to the actual layout of the /proc directories.

giampaolo:
Interesting and complicated subject. Some considerations:

* Linux is the only platform allowing to instantiate a thread via `Process` class as in `thread = Process(tid)`, and query the thread as if it was a process.
* As you pointed out, there currently is no distinction in the code whether `Process` points to a PID or TID, meaning we read thread info from `/proc/{pid}/*` instead of `/proc/{pid}/task/{tid}*`. Note that this applies to **all** process info, not only CPU timings.

[...]

PR #2061, opened 2022-01-17
- Adding patch from issue #1989 to correct cpu_percent on thread ID

RobinC-18:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: 1989

## Description

Applying the patch described in issue #1989, to enable a user to query the cpu_percent of a TID and it to return the correct values for that specific thread, rather than of the master_pid.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1577,6 +1577,31 @@ def pids():
     return [int(x) for x in os.listdir(b(get_procfs_path())) if x.isdigit()]
 
 
+def pid_for_tid(pid_or_tid):
+    """If the ID refers to a thread, returns the parent/master PID of the
+    process which spawned the thread, else None.
+    """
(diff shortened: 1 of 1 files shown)


foamdino:
This patch is something we're using in production in a local fork at the moment.  Would it be possible to get this merged into mainline so we can retire our local fork?

ajmmm:
Would be great if this PR was accepted - also having to maintain my own fork right now. :(

ajmmm:
@giampaolo is there any chance of getting this PR reviewed and merged please?
Thanks!

RobinC-18:
Hi @giampaolo, thanks for the reply.

We could look at a refactor that removes the check for the PID identity and whether it is a TID into an API that is only run when the caller explicitly asks for it to be checked? That means the `__init__` performance would stay the same, however the option would be there is a caller wanted to know whether a PID was a TID or not?

PR #2062, opened 2022-01-17
[Linux] Properly raise ZombieProcess in exe() method

Jongy:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: None

## Description
`psutil.Process(pid).exe()` returns `""` for zombie processes, incorrectly. It should raise `ZombieProcess`, and return `""` only for kernel threads.

This PR fixes the logic (I wrote something quick, I suppose that the logic can be reordered more nicely maybe).

Now we get:
* Alive processes - their `exe`
* Non-existing PIDs - `NoSuchProcess`
* Kernel threads - `""`
* Zombie processes - `ZombieProcess`.

Tested on kernel 5.4.

Also fixed & improved the comment along the way.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1741,16 +1741,16 @@ class Process(object):
         try:
             return readlink("%s/%s/exe" % (self._procfs_path, self.pid))
         except (FileNotFoundError, ProcessLookupError):
-            # no such file error; might be raised also if the
-            # path actually exists for system processes with
-            # low pids (about 0-20)
-            if os.path.lexists("%s/%s" % (self._procfs_path, self.pid)):
-                return ""
+            if not pid_exists(self.pid):
+                raise NoSuchProcess(self.pid, self._name)
             else:
-                if not pid_exists(self.pid):
-                    raise NoSuchProcess(self.pid, self._name)
-                else:
+                # its /proc/pid directory exists, but the /exe file does not.
+                # it's either a zombie
+                if self.status() == "zombie":
                     raise ZombieProcess(self.pid, self._name, self._ppid)
+                else:
+                    # or a kernel thread, for whom we return an empty string.
+                    return ""
         except PermissionError:
             raise AccessDenied(self.pid, self._name)
 


Jongy:
I will look into the failing test later today

Jongy:
I looked into the test. The logic here is incorrect:
```
                # ...but if /proc/pid no longer exist we're supposed to treat
                # it as an alias for zombie process
                with mock.patch('psutil._pslinux.os.path.lexists',
                                return_value=False):
                    self.assertRaises(
                        psutil.ZombieProcess, psutil.Process().exe)
```

if `/proc/pid` no longer exists, it's not a zombies. Zombies have their `/proc/pid`. It's a zombie if `/exe` link is missing & `.status()` is `"zombie"`.

How should I fix the test? The `lexists` patch is relevant for `NoSuchProcess`, so I can change that, & add another mock for `status()` which makes it look like a zombie?

giampaolo:
> psutil.Process(pid).exe() returns "" for zombie processes, incorrectly. It should raise ZombieProcess, and return "" only for kernel threads.

Mmm why are you saying so? Can you provide a test case?
We do have a unit test for zombie processes (and it passes on Linux):
https://github.com/giampaolo/psutil/blob/1da9f7928e79ecc127768e7a5bc4eb7f785f4513/psutil/tests/test_process.py#L1297
Also, if there really is a problem with zombie detection, then we also have a problem with other Process methods, not only `exe()`.

With that said, by looking at the Linux code I also suspect we're not detecting/handling zombie properly, because BSD and OSX implementations do something different, similar to what you're doing in this PR:
* OSX: https://github.com/giampaolo/psutil/blob/1da9f7928e79ecc127768e7a5bc4eb7f785f4513/psutil/_psosx.py#L331-L336
* BSD: https://github.com/giampaolo/psutil/blob/1da9f7928e79ecc127768e7a5bc4eb7f785f4513/psutil/_psbsd.py#L545-L550

giampaolo:
OK, I see what you mean now. `exe()` returns an empty string and `test_zombie_process` is too relaxed and does not test this condition (assumes `""` is a legit return value).

Jongy:
> OK, I see what you mean now. `exe()` returns an empty string and `test_zombie_process` is too relaxed and does not test this condition (assumes `""` is a legit return value).

Precisely :)

I'm available to work on the PR if you want any changes to get this merged.

Jongy:
We're using a patched version to work around this bug:
```
def process_exe(process: psutil.Process) -> str:
    """
    psutil.Process(pid).exe() returns "" for zombie processes, incorrectly. It should raise ZombieProcess, and return ""
    only for kernel threads.

    See https://github.com/giampaolo/psutil/pull/2062
    """
    exe = process.exe()
    if exe == "":
        if is_process_zombie(process):
            raise psutil.ZombieProcess(process.pid)
        raise MissingExePath(process)
    return exe
```

I'd love to remove it and get back to mainline haha.

giampaolo:
I'm afraid this may affect other methods other than `exe()` and requires a more generic solution.

giampaolo:
As I suspected, it turns out the problem not only affects  process `exe()`. We have it also in `cmdline()` and `memory_maps()`. I created https://github.com/giampaolo/psutil/pull/2288 which fixes the problem in a more generic way, and also expands tests for zombie processes.

Jongy:
> As I suspected, it turns out the problem not only affects process `exe()`. We have it also in `cmdline()` and `memory_maps()`. I created #2288 which fixes the problem in a more generic way, and also expands tests for zombie processes.

That's great! Thanks.

PR #2079, opened 2022-02-23
Fixes #2078: Import the inspect module on use

### LINKED ISSUE #2078: Calling _set_debug(True) doesn't import `inspect`
When the tests enable debug messages, they call this function and pass `True`:

https://github.com/giampaolo/psutil/blob/dd57b9a109d6db9a87247d5f47303b87672dfa78/psutil/__init__.py#L2349-L2355

However, if the environment variable wasn't already set, the import of `inspect` was skipped and any attempt to display a debug message will fail.

https://github.com/giampaolo/psutil/blob/dd57b9a109d6db9a87247d5f47303b87672dfa78/psutil/_common.py#L46-L48

The import of `inspect` should probably just be done inside the debug function.

zooba:
## Summary

* OS: All
* Bug fix: yes
* Type: tests
* Fixes: 2078

## Description

This avoids issues when debug messages are enabled at runtime (for example, in the test runner) as the module-level import may have been skipped.

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -44,8 +44,6 @@ else:
 # can't take it from _common.py as this script is imported by setup.py
 PY3 = sys.version_info[0] == 3
 PSUTIL_DEBUG = bool(os.getenv('PSUTIL_DEBUG', 0))
-if PSUTIL_DEBUG:
-    import inspect
 _DEFAULT = object()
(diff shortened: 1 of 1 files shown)


giampaolo:
Hey! Hi Steve. :)

PR #2080, opened 2022-02-25
Improves robustness of some tests that were too environment-specific

zooba:
## Summary

* OS: Windows
* Bug fix: yes
* Type: tests
* Fixes: 

## Description

I've been running the test suite in a slightly different setup recently, and these tests were failing due to assumptions about the environment. I've tried to tweak them to be more robust in general, and happy to tweak more if you like.

### DIFF
--- a/psutil/tests/test_process.py
+++ b/psutil/tests/test_process.py
@@ -799,19 +799,29 @@ class TestProcess(PsutilTestCase):
         init = p.nice()
         try:
             if WINDOWS:
-                for prio in [psutil.NORMAL_PRIORITY_CLASS,
-                             psutil.IDLE_PRIORITY_CLASS,
+                # A CI runner may limit our maximum priority, which will break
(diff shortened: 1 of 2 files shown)


psutil/tests/test_process.py
```diff
@@ -799,19 +799,32 @@ def test_nice(self):
...
         try:
             if WINDOWS:
-                for prio in [psutil.NORMAL_PRIORITY_CLASS,
-                             psutil.IDLE_PRIORITY_CLASS,
+                # A CI runner may limit our maximum priority, which will break
+                # this test. Instead, we test in order of increasing priority,
+                # and match either the expected value or the highest so far.
+                # We assume at least NORMAL will work, so complete failure will
+                # be caught on lower priorities.
+                highest_prio = psutil.NORMAL_PRIORITY_CLASS
```
zooba, line 807: Actually, this needs tweaking already or the diagnostic message on failure is going to be bad.

Issue #1909, opened 2021-01-26
[Linux] Test failure due to broken handling of IPv6 [FAIL: psutil.tests.test_linux.TestSystemNetIfAddrs.test_ips]


torsava:
## Summary

* OS: Linux
* Architecture: 64bit (x86_64)
* Psutil version: 5.8.0
* Python version: Python 3.9.1
* Type: tests

## Description

Test `psutil.tests.test_linux.TestSystemNetIfAddrs.test_ips` fails when an interface has multiple IPv6 addresses. 

Test failure:

```
...
======================================================================
FAIL: psutil.tests.test_linux.TestSystemNetIfAddrs.test_ips
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/psutil-release-5.8.0/psutil/tests/test_linux.py", line 947, in test_ips
    self.assertEqual(address, get_ipv6_address(name))
AssertionError: '2620:52:0:1330:d7b2:99a4:96ab:a55' != 'fe80::f1d4:c17e:7251:901b'
- 2620:52:0:1330:d7b2:99a4:96ab:a55
+ fe80::f1d4:c17e:7251:901b
----------------------------------------------------------------------
Ran 543 tests in 3.524s
FAILED (failures=1, skipped=182)
```

Test case:
``` 
   def test_ips(self):
        for name, addrs in psutil.net_if_addrs().items():
            for addr in addrs:
                if addr.family == psutil.AF_LINK:
                    self.assertEqual(addr.address, get_mac_address(name))
                elif addr.family == socket.AF_INET:
                    self.assertEqual(addr.address, get_ipv4_address(name))
                    self.assertEqual(addr.netmask, get_ipv4_netmask(name))
                    if addr.broadcast is not None:
                        self.assertEqual(addr.broadcast,
                                         get_ipv4_broadcast(name))
                    else:
                        self.assertEqual(get_ipv4_broadcast(name), '0.0.0.0')
                elif addr.family == socket.AF_INET6:
                    # IPv6 addresses can have a percent symbol at the end.
                    # E.g. these 2 are equivalent:
                    # "fe80::1ff:fe23:4567:890a"
                    # "fe80::1ff:fe23:4567:890a%eth0"
                    # That is the "zone id" portion, which usually is the name
                    # of the network interface.
                    address = addr.address.split('%')[0]
                    self.assertEqual(address, get_ipv6_address(name))
```

Current code of the `get_ipv6_address` function:

```
[...]

PR #2086, opened 2022-03-18
Fixed an issue that caused linux tests to fail when running on a system with multiple ipv6 addresses per interface

myheroyuki:
## Summary

* OS: { Ubuntu 20.04}
* Bug fix: { yes }
* Type: { tests }
* Fixes: {Linux tests no longer fail when one interface has multiple ipv6 addresses }

## Description

  When one interface has two ipv6 addresses a test would fail because it would expect one address to be equal to the other. It now checks the ipv6 address against all other ipv6 addresses on the interface.


myheroyuki:
This was mentioned in issue #1909

PR #2088, opened 2022-03-28
[NetBSD] get data about cached memory usage from /proc/meminfo as well

0-wiz-0:
## Summary

* OS: NetBSD
* Bug fix: yes
* Type: core
* Fixes: -
* 
## Description

This fixes the reporting of 'cached' memory on NetBSD.


0-wiz-0:
Somethings up with the CIs:
The linter complains about whitespace in a file I didn't touch, and the 'issues workflow run' fails because it wants to add a label, but isn't allowed to.

PR #2039, opened 2021-12-25
Drop Python 2.6 support

mayeut:
## Summary

* OS: ALL
* Bug fix: no
* Type: core
* Fixes: #1053

## Description

Let's drop python 2.6 support. It's not being tested anymore, its usage was mostly linked to CentOS 6 (c.f. #1053) which is now EOL. The stats for python 2.6 downloads also show that it's not being used anymore (0 downloads for 5.8.0 version, 67 downloads across all versions for the last month).

```
pypinfo -l 25 --percent --where 'file.version = "5.8.0" AND file.project = "psutil" AND details.installer.name = "pip"' psutil pyversion version  
Served from cache: False
Data processed: 1.47 GiB
Data billed: 1.47 GiB
Estimated cost: $0.01

| python_version | version | percent | download_count |
| -------------- | ------- | ------- | -------------- |
| 3.7            | 5.8.0   |  55.45% |     15,460,755 |
| 3.8            | 5.8.0   |  16.87% |      4,703,807 |
| 3.6            | 5.8.0   |  10.90% |      3,040,032 |
| 3.9            | 5.8.0   |   8.56% |      2,387,500 |
| 2.7            | 5.8.0   |   5.31% |      1,480,235 |
| 3.5            | 5.8.0   |   2.02% |        562,382 |
| 3.10           | 5.8.0   |   0.88% |        246,692 |
| 3.4            | 5.8.0   |   0.00% |          1,213 |
| 3.11           | 5.8.0   |   0.00% |            301 |
| 3.2            | 5.8.0   |   0.00% |              2 |
| Total          |         |         |     27,882,919 |
```
```
pypinfo --percent psutil pyversion
Served from cache: False
Data processed: 1.19 GiB
Data billed: 1.19 GiB
Estimated cost: $0.01

| python_version | percent | download_count |
| -------------- | ------- | -------------- |
| 3.7            |  52.29% |     16,852,218 |
| 3.8            |  16.73% |      5,391,531 |
| 3.6            |  12.94% |      4,170,766 |
| 2.7            |   7.77% |      2,504,686 |
| 3.9            |   7.59% |      2,447,477 |
| 3.5            |   1.85% |        596,800 |
| 3.10           |   0.80% |        258,639 |
| 3.4            |   0.02% |          5,583 |
| 3.11           |   0.00% |            304 |
| 2.6            |   0.00% |             67 |
| Total          |         |     32,228,071 |
```

PR #1178, opened 2017-11-20
Drop support for EOL Python 2.6

hugovk:
https://github.com/giampaolo/psutil/issues/1053 says to drop Python 2.6 support in unit tests and scripts, and to issue a deprecation warning in psutil.

In case or until you want to go further, this also removes Python 2.6 code, plus already dropped 3.2-3.3.

PR #2099, opened 2022-04-15
Drop Python 2.6 support

### LINKED ISSUE #1053: Drop Python 2.6 support in unit tests and scripts, deprecate it in psutil
Maintaining Python 2.6 support is a pain in the butt from a CI / testing perspective for a variety of reason (unittest2, a specific version of mock, setuptools, etc.). Also, both [pip](https://github.com/pypa/pip/issues/3955) and [setuptools](https://github.com/pypa/setuptools/issues/878) dropped support for 2.6, so installing psutil on 2.6 is becoming harder (although still possible). For these reasons I would like to drop python 2.6 for tests only.

There is an argument for **not** dropping 2.6 support in psutil "as a lib" (so excluding tests) yet: CentOS 6. CentOS 6 still ships 2.6 by default and AFAIK it is still in use on a non negligible number of servers (unfortunately). Being psutil a "system library" I am kind of worried about that so I think it'd be nice to keep support for a little while longer and just deprecate it by issuing a warning on import. 

CentOS 6 EOL is scheduled for November 2020. I am not sure I want to maintain support 'till then, but the point is that there currently are no 2.7-only features such as set literals, set/dict comprehensions, multiple ctx managers or OrderedDict which I may want to use in psutil code, so why break it? 

On the other hand, some of those 2.7-only features may be used in the test suite (e.g. replace `contextlib.nexted` with multiple ctx managers). As far as 2.6 goes I would still like travis to test it by installing psutil on 2.6 and import it (that's it), but the test suite will not be run. 
At the same time, `setup.py` may print a warning telling that 2.6 is supported but not testable.

As for scripts (/scripts dir): we can also drop support (and e.g. be free to use the `argparse` module).

TODOs:

* drop 2.6 support in unit tests and scritps
* deprecate 2.6 support in psutil by issuing a deprecation warning

hugovk:
## Summary

* OS: ALL
* Bug fix: no
* Type: core
* Fixes: https://github.com/giampaolo/psutil/issues/1053

## Description

Follow on from https://github.com/giampaolo/psutil/pull/2039, remove a few extra redundant Python 2.6 bits from https://github.com/giampaolo/psutil/pull/1178.

Also bump GitHub Actions to v3 and cache pip.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -17,7 +17,7 @@ XXXX-XX-XX
 
 - 2048_: ``AttributeError`` is raised if ``psutil.Error`` class is raised
   manually and passed through ``str``.
-- 2049_, [Linux]: `cpu_freq`_ erroneously returns ``curr`` value in GHz while
+- 2049_, [Linux]: `cpu_freq()`_ erroneously returns ``curr`` value in GHz while
   ``min`` and ``max`` are in MHz.
 - 2050_, [Linux]: `virtual_memory()`_ may raise ``ValueError`` if running in a
   LCX container.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,7 +7,7 @@ XXXX-XX-XX
 
 **Enhancements**
 
-- 1053_: dropped support for Python 2.6.  (patch by Matthieu Darbois)
+- 1053_: dropped support for Python 2.6.  (patches by Matthieu Darbois and Hugo van Kemenade)
 - 2050_, [Linux]: increase ``read(2)`` buffer size from 1k to 32k when reading
   ``/proc`` pseudo files line by line. This should help having more consistent
   results.
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -31,6 +31,7 @@ import tempfile
 import textwrap
 import threading
 import time
+import unittest
 import warnings
 from socket import AF_INET
 from socket import AF_INET6
@@ -56,11 +57,6 @@ from psutil._compat import u
 from psutil._compat import unicode
 from psutil._compat import which
 
-
-if PY3:
-    import unittest
-else:
-    import unittest2 as unittest  # requires "pip install unittest2"
 
 try:
     from unittest import mock  # py3
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -858,6 +858,11 @@ class TestCase(unittest.TestCase):
         def runTest(self):
             pass
 
+        @contextlib.contextmanager
+        def subTest(self, *args, **kw):
+            # fake it for python 2.7
+            yield
+
 
 # monkey patch default unittest.TestCase
 unittest.TestCase = TestCase
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,10 +46,12 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
       with:
         python-version: 3.9
+        cache: pip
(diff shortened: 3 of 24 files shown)


HISTORY.rst
```diff
@@ -11,6 +11,7 @@ XXXX-XX-XX
   ``/proc`` pseudo files line by line. This should help having more consistent
   results.
 - 2057_, [OpenBSD]: add support for `cpu_freq()`_.
+- 2039_, 2099_: drop support for Python 2.6, EOL since 2013-10-2.
```
giampaolo, line 14: please `git pull` your branch; there is an entry for this a couple of lines earlier:

```
- 1053_: dropped support for Python 2.6.  (patch by Matthieu Darbois)
- ```

Feel free to add your name (also to CREDITS file).

File: psutil/tests/__init__.py [hunk 2]
```diff
@@ -57,11 +58,6 @@
 from psutil._compat import which
 
 
-if PY3:
-    import unittest
-else:
-    import unittest2 as unittest  # requires "pip install unittest2"
```
giampaolo, line 63: I'm so happy to see this go! Other test files (`test_linux.py`) should stop doing:

```
from psutil.tests import unittest
```

...and instead do:

```
import unittest
```

Could you please update those files as well?

giampaolo (changes requested):
Thanks a lot for this. I added a couple of inline comments.

HISTORY.rst
hugovk: I don't see this at https://github.com/giampaolo/psutil/blob/master/HISTORY.rst, maybe you need to push it?

giampaolo: @hugovk sorry you right; I forgot to push indeed hehel. I pushed now.

psutil/tests/__init__.py
hugovk: This file still has a monkey patch of `unittest.TestCase` for a modified `__str__`, `assertRaisesRegex` compat code for 2.7 and handling for `multiprocessing.Pool`.

giampaolo: I think the monkey patch should still work because `test/__init__.py` is imported before other `test/*.py` files.
When they do `import unittest` the should automatically inherit the monkey-patched version of `unittest.TestCase`.

hugovk: Hmm, https://github.com/giampaolo/psutil/pull/2099/commits/4bc28e515b5ba344a657ea302eea281415c0543f passed on GHA but failed on AppVeyor:

```
psutil.tests.test_windows.TestSystemAPIs.test_nic_names ... ok
psutil.tests.test_windows.TestSystemAPIs.test_pids ... skipped 'test not relieable on appveyor'
psutil.tests.test_windows.TestSystemAPIs.test_total_phymem ... ok
psutil.tests.test_windows.WindowsTestCase.runTest ... ok
ERROR: psutil.tests.test_process.TestProcess.test_nice
  File "C:\projects\psutil\psutil\tests\test_process.py", line 812, in test_nice
    with self.subTest(prio=prio):
AttributeError: 'TestProcess' object has no attribute 'subTest'
Ran 575 tests in 12.912s
FAILED (errors=1, skipped=263)
FAILED
Command exited with code 1
```

giampaolo: I guess `unittest2` package provided `subTest` method on Python 2.7, a unittest method which only appeared later in python 3.X. Not a big deal. I prefer to get rid of unittest2 dependency and fake `subTest` method existence in our `unittest.TestCase` subclass. I think something like this should work:

```python
if not PY3:
    @contextlib.contextmanager
    def subTest(self, *args, **kw):
        # fake it for python 2.7
        yield
```

hugovk: That looks better, thanks!

(Another option is dropping 2.7 ;)

giampaolo: I think 2.7 will stay for a very long time (I'd put a but to at least one decade).

PR #2102, opened 2022-05-01
feature: use ABI3 for cp36+

### LINKED ISSUE #1954: [M1 macOS BigSur] no suitable image found
## Summary

* OS: macOS Big Sur
* Architecture: ARM
* Psutil version: 5.8.0
* Python version: 3.9.4

## Description

I am trying to use the python library [neovim-remote](https://github.com/mhinz/neovim-remote), but I am running into this psutil error while on my M1 mac. I opened an [issue](https://github.com/mhinz/neovim-remote/issues/165) there, but the author suggested I open one here.

I installed neovim remote with `pip3 install --user neovim-remote`

Thanks!

<details>
<summary>This is the stacktrace I get when try to use that tool</summary>
<pre>
  File "/Users/mitchell/Library/Python/3.9/bin/nvr", line 5, in <module>
    from nvr.nvr import main
  File "/Users/mitchell/Library/Python/3.9/lib/python/site-packages/nvr/__init__.py", line 1, in <module>
    from .nvr import main
  File "/Users/mitchell/Library/Python/3.9/lib/python/site-packages/nvr/nvr.py", line 33, in <module>
    import psutil
  File "/Users/mitchell/Library/Python/3.9/lib/python/site-packages/psutil/__init__.py", line 121, in <module>
    from . import _psosx as _psplatform
  File "/Users/mitchell/Library/Python/3.9/lib/python/site-packages/psutil/_psosx.py", line 15, in <module>
    from . import _psutil_osx as cext
ImportError: dlopen(/Users/mitchell/Library/Python/3.9/lib/python/site-packages/psutil/_psutil_osx.cpython-39-darwin.so, 2): no suitable image found.  Did find:
        /Users/mitchell/Library/Python/3.9/lib/python/site-packages/psutil/_psutil_osx.cpython-39-darwin.so: mach-o, but wrong architecture
        /Users/mitchell/Library/Python/3.9/lib/python/site-packages/psutil/_psutil_osx.cpython-39-darwin.so: mach-o, but wrong architecture
</pre>
</summary>

### LINKED ISSUE #2089: Generate Wheels for CPython 3.11 (or Use Limited API)
The faster-cpython project uses pyperformance to run benchmarks, which in turn uses pyperf, which optionally uses psutil.  We run into problems when we try to benchmark CPython's main branch, since there are no corresponding wheels on PyPI.  See https://github.com/python/pyperformance/issues/148.

So it would be helpful to have wheels available on PyPI for the CPython 3.11 releases.  Furthermore, it would likewise help to have a wheel built for the main branch.

One alternative would be to release a build using CPython's limited API.  The resulting extension would be automatically compatible with the newer CPython.

mayeut:
## Summary

* OS: all
* Bug fix: no
* Type: wheels
* Fixes: 
fixes #2089

## Description
Use Limited API when building with CPython 3.6+ on Linux/macOS/Windows.
This allows to use pre-built wheels in all future versions of CPython 3.
Windows builds for CPython 3.6+ are moved from AppVeyor to GitHub Actions.

Diff of wheels produced:
```diff
 psutil-5.9.1-cp27-cp27m-macosx_10_9_x86_64.whl
 psutil-5.9.1-cp27-cp27m-manylinux2010_i686.whl
 psutil-5.9.1-cp27-cp27m-manylinux2010_x86_64.whl
 psutil-5.9.1-cp27-cp27m-win_amd64.whl
 psutil-5.9.1-cp27-cp27m-win32.whl
 psutil-5.9.1-cp27-cp27mu-manylinux2010_i686.whl
 psutil-5.9.1-cp27-cp27mu-manylinux2010_x86_64.whl
-psutil-5.9.1-cp36-cp36m-macosx_10_9_x86_64.whl
+psutil-5.9.1-cp36-abi3-macosx_10_9_x86_64.whl
-psutil-5.9.1-cp36-cp36m-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
+psutil-5.9.1-cp36-abi3-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
-psutil-5.9.1-cp36-cp36m-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
+psutil-5.9.1-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-psutil-5.9.1-cp36-cp36m-win_amd64.whl
+psutil-5.9.1-cp36-abi3-win_amd64.whl
-psutil-5.9.1-cp36-cp36m-win32.whl
+psutil-5.9.1-cp36-abi3-win32.whl
-psutil-5.9.1-cp37-cp37m-macosx_10_9_x86_64.whl
-psutil-5.9.1-cp37-cp37m-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
-psutil-5.9.1-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-psutil-5.9.1-cp37-cp37m-win_amd64.whl
-psutil-5.9.1-cp37-cp37m-win32.whl
-psutil-5.9.1-cp38-cp38-macosx_10_9_x86_64.whl
+psutil-5.9.1-cp38-abi3-macosx_11_0_arm64.whl
-psutil-5.9.1-cp38-cp38-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
-psutil-5.9.1-cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-psutil-5.9.1-cp38-cp38-win_amd64.whl
-psutil-5.9.1-cp38-cp38-win32.whl
-psutil-5.9.1-cp39-cp39-macosx_10_9_x86_64.whl
-psutil-5.9.1-cp39-cp39-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
-psutil-5.9.1-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-psutil-5.9.1-cp39-cp39-win_amd64.whl
-psutil-5.9.1-cp39-cp39-win32.whl
-psutil-5.9.1-cp310-cp310-macosx_10_9_x86_64.whl
-psutil-5.9.1-cp310-cp310-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
-psutil-5.9.1-cp310-cp310-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-psutil-5.9.1-cp310-cp310-win_amd64.whl
-psutil-5.9.1-cp310-cp310-win32.whl
```


setup.py
```diff
@@ -31,6 +31,10 @@
         setuptools = None
         from distutils.core import Extension
         from distutils.core import setup
+    try:
+        from wheel.bdist_wheel import bdist_wheel
+    except ImportError:
+        bdist_wheel = None
```
giampaolo, line 37: I wonder: is it possible (or worth it) to raise in case we're explicitly running cbuildwheels tool? Something like (not tested, just to illustrate what I mean):

```python
    try:
        from wheel.bdist_wheel import bdist_wheel
    except ImportError:
        if "RUNNING_CIBUILDWHEEL_CLI" in os.environ:
            raise
```
....or perhaps `cbuildwheel` command already fails if `bdist_wheel` package is not installed?

setup.py
```diff
@@ -341,11 +362,25 @@ def get_sunos_update():
...
+    class bdist_wheel_abi3(bdist_wheel):
+        def finalize_options(self):
+            bdist_wheel.finalize_options(self)
+            self.root_is_pure = False
+
+        def get_tag(self):
+            python, abi, plat = bdist_wheel.get_tag(self)
+            return python, "abi3", plat
+
+    cmdclass["bdist_wheel"] = bdist_wheel_abi3
```
giampaolo, line 376: This looks like an hack. I guess this means bdist_wheel does not support Py_LIMITED_API natively? Should we keep an eye on this in the future? (e.g. the original class changes and this no longer works)

.ci/appveyor/install.ps1
```diff
@@ -1,85 +0,0 @@
...
-    & $pip_path install $pkg
-}
-
-function main () {
-    InstallPython $env:PYTHON_VERSION $env:PYTHON_ARCH $env:PYTHON
-    InstallPip $env:PYTHON
-    InstallPackage $env:PYTHON wheel
-}
-
-main
```
giampaolo, line 85: I added this file years ago so I don't even remember what it's for. Was this necessary for Python 3 only?

.github/workflows/build.yml
```diff
@@ -60,6 +60,49 @@ jobs:
...
+      run: |
+        mkdir .tests
+        cd .tests
+        pip install $(find ../wheelhouse -name '*-cp36-abi3-${{ matrix.architecture == 'x86' && 'win32' || 'win_amd64'}}.whl')[test]
+        export PYTHONWARNINGS=always
+        export PYTHONUNBUFFERED=1
+        export PSUTIL_DEBUG=1
+        python ../psutil/tests/runner.py
+        python ../psutil/tests/test_memleaks.py
+      shell: bash
```
giampaolo, line 104: Before I was against stop using appveyor, but with pyproject.toml, wheels and now `Py_LIMITED_API` I see why this is necessary. One question though: with this new addition GitHub will test setup.py using `Py_LIMITED_API`, while appveyor CI will test python 2.7 + setup.py **without** `Py_LIMITED_API`. Am I correct?  Do you still see value in keep using appveyor or do you think we should abandon it? 

Some background on why we're using appveyor:

* historical reason: initially (years ago) GitHub did not support windows
* to split work on 2 CI services and therefore complete the whole CI test/build sooner (but perhaps I'm wrong on this)

giampaolo (changes requested):
This truly is an amazing work. Yet another thing I vaguely read about but never had the chance to delve into. Thanks.

I'm reading https://docs.python.org/3/c-api/stable.html. It poses some important considerations and suggests to test all minor Python versions. According to psutil's `setup.py` the most minor version we support is Python 3.4:

```python
    if setuptools is not None:
        kwargs.update(
            python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
            extras_require=extras_require,
            zip_safe=False,
        )
```

As such I wonder: 
* would it make sense for `build.yml` to test 3.4? 
* what python versions are we testing right now?  I see that windows has `["3.7", "3.8", "3.9", "3.10", "~3.11.0-0"]` but I'm unclear about the other platforms.

Again, thanks for this great work!

giampaolo:
Another consideration. https://docs.python.org/3/c-api/stable.html mentions `Py_LIMITED_API` may have some impact on performances. It would be good to make some benchmarks before merging this. Linux makes little use of the C extension, so I think we could test this on macOS or Windows (better).
I guess [scripts/internal/print_api_speed.py](https://github.com/giampaolo/psutil/blob/master/scripts/internal/print_api_speed.py) could be used for this task - or probably a slightly modified version which executes each API a certain number of times.

giampaolo:
CC @ericsnowcurrently in case he has some comment about performances (see https://github.com/giampaolo/psutil/issues/2089).

psutil/_psutil_linux.c
```diff
@@ -298,29 +298,25 @@ psutil_proc_cpu_affinity_set(PyObject *self, PyObject *args) {
...
 
-    py_cpu_seq = PySequence_Fast(py_cpu_set, "expected a sequence or integer");
-    if (!py_cpu_seq)
-        goto error;
-    seq_len = PySequence_Fast_GET_SIZE(py_cpu_seq);
+    seq_len = PySequence_Size(py_cpu_set);
     CPU_ZERO(&cpu_set);
     for (i = 0; i < seq_len; i++) {
-        PyObject *item = PySequence_Fast_GET_ITEM(py_cpu_seq, i);
+        PyObject *item = PySequence_GetItem(py_cpu_set, i);
```
encukou, line 313:
```suggestion
        PyObject *item = PySequence_GetItem(py_cpu_set, i);
        if (!item) {
            goto error;
        }
```
(Previously, `PySequence_Fast` checked for errors while getting items.)

encukou:
> Another consideration. https://docs.python.org/3/c-api/stable.html mentions Py_LIMITED_API may have some impact on performances.

Hi! I wrote that page.

Limited API is a bit slower, but it's still on the C level. If the code is primarily in Python, you're not likely to notice a slowdown.

Tthe biggest threat of slowdown I know of is that common operations (`PyINCREF` / `Py_XDECREF` etc.) can become C function calls, rather than being inlined.
If you are worried about C function call overhead, you might look into building version-specific (fast) wheels *in addition* to a Stable ABI one (which would support e.g. unreleased versions of CPython).
But if you're [not worried even about *Python* call overhead](https://github.com/giampaolo/psutil/blob/master/psutil/_psosx.py#L161), don't worry about Stable ABI being slow :)

Still, if you run the benchmarks and they prove me wrong, I'd love to hear about that!

setup.py
ben9923: Here's the example linked from the cibuildwheel docs:

.github/workflows/build.yml
ben9923: I personally find using 2 CI services a bit confusing, and GHA's better integration with the repo/PRs is a nice advantage.

I believe there's no speed benefit to this split. GHA's concurrency limits are (much?) higher than the current jobs amount.

setup.py
```diff
@@ -100,6 +104,14 @@ def get_version():
 VERSION = get_version()
 macros.append(('PSUTIL_VERSION', int(VERSION.replace('.', ''))))
 
+PY36_PLUS = sys.version_info[:2] >= (3, 6)
+CP36_PLUS = PY36_PLUS and sys.implementation.name == "cpython"
```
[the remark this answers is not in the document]
mayeut: We can't use `CPYTHON  = sys.implementation.name == "cpython"` because [`sys.implementation`](https://docs.python.org/3/library/sys.html#sys.implementation) was introduced in Python 3.3 with [PEP 421](https://peps.python.org/pep-0421/).
I'll move the definitions below `PYPY = ...`.

setup.py
mayeut: > I guess this means bdist_wheel does not support Py_LIMITED_API natively? Should we keep an eye on this in the future? (e.g. the original class changes and this no longer works)

`wheel` only has static support for Py_LIMITED_API through config file as far as I know.
It's probably worth keeping an eye on this in the future.

> which you probably want to do too

Well, I can't remember where I saw a recommendation against building cp3x abi3 with a more recent CPython than 3.x but I'm pretty sure I saw it and there was a rationale behind it. It might not be important in this case but there's probably no harm letting this as implemented now.

> Is manually setting root_is_pure necessary?

Indeed it's not.

.ci/appveyor/install.ps1
mayeut: It might have been necessary to install some missing python versions on AppVeyor at some point.
It's definitely not needed for Python 2.7

.github/workflows/build.yml
mayeut: > Am I correct? 

Yes you are.

> Do you still see value in keep using appveyor or do you think we should abandon it?

If you want to keep Python 2.7 wheels on Windows, then I'd say it shall remain. AppVeyor still keeps old images around  and that includes some build tools "[required](https://cibuildwheel.readthedocs.io/en/1.x/faq/#windows-and-python-27)" for Python 2.7 that are not available for download anymore and not available on GHA.

Given AppVeyor has 1 concurrent job and GHA has higher concurrency limits than the current jobs amount, it will be faster with GHA. The current limiter on `master` to get a green CI is AppVeyor with an average of 30 minutes. This PR cuts that at least in half.

setup.py
ben9923: > Well, I can't remember where I saw a recommendation against building cp3x abi3 with a more recent CPython than 3.x but I'm pretty sure I saw it and there was a rationale behind it. It might not be important in this case but there's probably no harm letting this as implemented now.

I meant checking the implementation before setting the tag as `abi3` (i.e. `if python.startswith("cp"):`).
Now I see it's all wrapped with `if py_limited_api:` which is CPython-only, so this part of my comment is irrelevant. Whoops 😛

Indeed I think the official docs advise building cp3x ABI3 wheels using cp3x :)
Not sure what's the correct thing to do with the `bdist_wheel` class given the explicit setting of `Py_LIMITED_API = 0x03060000`, but I think either way should be fine here haha

mayeut:
@giampaolo,

> what python versions are we testing right now? I see that windows has ["3.7", "3.8", "3.9", "3.10", "~3.11.0-0"] but I'm unclear about the other platforms.

This PR adds Python 3.11 to the versions tested compared to `master` so Linux/macOS/Windows Python 2.7, 3.6->3.11
This PR splits Windows testing from the build job for Python 3.7->3.11 because the process tests do not like being run from a virtual environment starting with 3.7 (there are some comments about this which can be seen in the diff).

It does not add Python 3.4 or 3.5 tests.

> would it make sense for build.yml to test 3.4?

Up to you (but in another PR) depending on the level of support you want to give 3.4 (or 3.5)
```
pypinfo --percent 'psutil==5.9.*' pyversion 
Served from cache: False
Data processed: 2.54 GiB
Data billed: 2.54 GiB
Estimated cost: $0.02

| python_version | percent | download_count |
| -------------- | ------- | -------------- |
| 3.8            |  47.33% |     25,811,595 |
| 3.7            |  30.64% |     16,709,870 |
| 3.9            |   8.98% |      4,898,637 |
| 3.6            |   6.18% |      3,370,019 |
| 3.10           |   4.34% |      2,368,841 |
| 2.7            |   2.40% |      1,310,887 |
| 3.5            |   0.09% |         47,475 |
| 3.11           |   0.03% |         15,462 |
| 3.4            |   0.00% |            989 |
| 3.12           |   0.00% |            685 |
| Total          |         |     54,534,460 |
```

giampaolo:
@encukou wrote:
> Hi! I wrote that page.

It's nicely written. :) 

@encukou  wrote:
> Limited API is a bit slower, but it's still on the C level. If the code is primarily in Python, you're not likely to notice a slowdown.
> The biggest threat of slowdown I know of is that common operations (PyINCREF / Py_XDECREF etc.) can become C function calls, rather than being inlined.

Fair enough. Then I don't think we have a problem. The most "critical" parts I can think of are for/while loops where the C extension populates a list or dict in C, but nothing more. I assume that even with `Py_LIMITED_API`, this is still faster than doing the same in Python.

@ericsnowcurrently wrote:
> If you have a little bit of time and performance regressions are important for this project to avoid then one thing you might consider is writing some meaningful benchmarks for it. It's relatively easy to write benchmarks that you can run using pyperformance: https://pyperformance.readthedocs.io/custom_benchmarks.html.

Interesting. I don't think something that advanced is necessary here, but incidentally I'm working on another project where I may use something like that. Thanks.

giampaolo:
@mayeut given the comments above related to performance I don't think it's necessary to do further benchmarks, as I assume this work doesn't introduce any significant regression. Anyway, I have still improved the benchmark script a bit in https://github.com/giampaolo/psutil/commit/085691ac46b060aeb6d17a202fcc33062f61d9b5 (which eventually inspired https://github.com/giampaolo/psutil/issues/2157).

Also, completely unrelated, since I see you're from Paris: I will be in Paris (first time) from Saturday to Tuesday for vacation. If you want to meet for a coffee shoot me an email to g.rodola@gmail.com. =)

Merging.

giampaolo:
@mayeut I uploaded a new release on PYPI and got the following error when uploading Windows wheels on PYPI:

```
python3 -m twine upload dist/*  # upload tar.gz and Windows wheels on PyPI
Uploading distributions to https://upload.pypi.org/legacy/
Uploading psutil-5.9.4-cp27-cp27m-macosx_10_9_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 270.8/270.8 kB • 00:00 • 753.9 kB/s
Uploading psutil-5.9.4-cp27-cp27m-manylinux2010_i686.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 321.9/321.9 kB • 00:00 • 35.1 MB/s
Uploading psutil-5.9.4-cp27-cp27m-manylinux2010_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 326.2/326.2 kB • 00:00 • 69.5 MB/s
Uploading psutil-5.9.4-cp27-cp27m-win32.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 270.3/270.3 kB • 00:00 • 69.8 MB/s
Uploading psutil-5.9.4-cp27-cp27m-win_amd64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 273.6/273.6 kB • 00:00 • 27.1 MB/s
Uploading psutil-5.9.4-cp27-cp27mu-manylinux2010_i686.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 321.9/321.9 kB • 00:00 • 91.5 MB/s
Uploading psutil-5.9.4-cp27-cp27mu-manylinux2010_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 326.2/326.2 kB • 00:00 • 83.8 MB/s
Uploading psutil-5.9.4-cp36-abi3-macosx_10_9_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 271.3/271.3 kB • 00:00 • 39.1 MB/s
Uploading psutil-5.9.4-cp36-abi3-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 305.4/305.4 kB • 00:00 • 55.4 MB/s
Uploading psutil-5.9.4-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 308.1/308.1 kB • 00:00 • 28.4 MB/s
Uploading psutil-5.9.4-cp36-abi3-win32.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 276.1/276.1 kB • 00:00 • 32.0 MB/s
WARNING  Error during upload. Retry with the --verbose option for more details.                                               
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/                                                      
         The description failed to render for 'text/x-rst'. See https://pypi.org/help/#description-content-type for more      
         information.                                                                                                         
```

I'm not sure why this happens only with Windows wheels. I suspect it has to do with the package `long_description`, which is take from `README.rst`, even though it's converted by this script before uploading it on PYPI: https://github.com/giampaolo/psutil/blob/master/scripts/internal/convert_readme.py.
Do you have any idea what may be the problem?

giampaolo:
This is the problem:

```
~/svn/psutil {master}$ python3 -m twine  check dist/*
Checking dist/psutil-5.9.4-cp27-cp27m-macosx_10_9_x86_64.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27m-manylinux2010_i686.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27m-manylinux2010_x86_64.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27mu-manylinux2010_i686.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27mu-manylinux2010_x86_64.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27m-win32.whl: PASSED
Checking dist/psutil-5.9.4-cp27-cp27m-win_amd64.whl: PASSED
Checking dist/psutil-5.9.4-cp36-abi3-macosx_10_9_x86_64.whl: PASSED
Checking dist/psutil-5.9.4-cp36-abi3-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl: PASSED
Checking 
dist/psutil-5.9.4-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl: PASSED
Checking dist/psutil-5.9.4-cp36-abi3-win32.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.                                    
         line 9: Error: Error in "image" directive:                                                                           
         no content permitted.                                                                                                
                                                                                                                              
         image:: https://img.shields.io/pypi/dm/psutil.svg                                                                    
                                                                                                                              
             :target: https://pepy.tech/project/psutil                                                                        
                                                                                                                              
             :alt: Downloads                                                                                                  
Checking dist/psutil-5.9.4-cp36-abi3-win_amd64.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.                                    
         line 9: Error: Error in "image" directive:                                                                           
         no content permitted.                                                                                                
                                                                                                                              
         image:: https://img.shields.io/pypi/dm/psutil.svg                                                                    
                                                                                                                              
             :target: https://pepy.tech/project/psutil                                                                        
                                                                                                                              
             :alt: Downloads                                                                                                  
Checking dist/psutil-5.9.4-cp38-abi3-macosx_11_0_arm64.whl: PASSED
Checking dist/psutil-5.9.4.tar.gz: PASSED
```

Not sure why only affects windows ABI3 wheels.

mayeut:
@giampaolo,  I'll take a look at this.

PR #2114, opened 2022-06-23
Implicitly include <sys/param.h> so that __FreeBSD_version checks wor…

Torsten-B:
Implicitly include <sys/param.h> so that __FreeBSD_version checks work properly

## Summary

* OS: FreeBSD
* Bug fix: yes
* Type: core
* Fixes: 2113

## Description
Implicitly include <sys/param.h> on all code files in psutil/arc/freebsd that make use of  __FreeBSD_version.
While some code already includes it directly and others pull it in indirectly via other includes, mem.c was not and made __FreeBSD_version checks silently fail on some common FreeBSD installations, including default setups with FreeBSD on arm platforms.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -786,3 +786,6 @@ I: 2039
 N: Hugo van Kemenade
 W: https://github.com/hugovk
 I: 2099
+
+N: Torsten Blum
+I: 2114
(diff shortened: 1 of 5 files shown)


giampaolo:
Looks good. Could you please also update HISTORY and CREDITS files?

PR #2140, opened 2022-09-15
Fix macOS Python 2.7 build

ben9923:
## Summary

* OS: macOS
* Bug fix: no
* Type: tests, wheels

## Description

Messing around with the CI build workflows I noticed something weird with the macOS Python 2.7 builds, being affected from Linux test issues.
Looking deeper I saw the macOS job is actually building & testing Linux wheels... 😝 

I've set the macOS version to 10.15 like the Python 3 jobs, as apparently macOS 11+ breaks `environ` APIs (There's an open issue for it in `psutil`, should probably be documented).


giampaolo:
Good catch. Merged.

PR #2146, opened 2022-09-25
chore: skip test_cpu_freq on macOS arm64

mayeut:
## Summary

* OS: macOS
* Bug fix: no
* Type: tests
* Fixes:

## Description

macOS arm64 does not support cpu_freq: issue #1892

Signed-off-by: mayeut <mayeut@users.noreply.github.com>

### DIFF
--- a/psutil/tests/test_osx.py
+++ b/psutil/tests/test_osx.py
@@ -6,6 +6,7 @@
 
 """macOS specific tests."""
 
+import platform
 import re
 import time
 import unittest
@@ -144,6 +145,8 @@ class TestSystemAPIs(PsutilTestCase):
         num = sysctl("sysctl hw.physicalcpu")
         self.assertEqual(num, psutil.cpu_count(logical=False))
 
+    # TODO: remove this once 1892 is fixed
+    @unittest.skipIf(platform.machine() == 'arm64', "skipped due to #1892")
     def test_cpu_freq(self):
         freq = psutil.cpu_freq()
         self.assertEqual(
--- a/psutil/tests/test_contracts.py
+++ b/psutil/tests/test_contracts.py
@@ -12,6 +12,7 @@ Some of these are duplicates of tests test_system.py and test_process.py
 import errno
 import multiprocessing
 import os
+import platform
 import signal
 import stat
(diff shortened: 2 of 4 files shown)


psutil/tests/__init__.py
```diff
@@ -130,6 +131,7 @@
 CI_TESTING = APPVEYOR or GITHUB_ACTIONS
 # are we a 64 bit process?
 IS_64BIT = sys.maxsize > 2 ** 32
+MACOS_ARM64 = MACOS and platform.machine() == 'arm64'
```
giampaolo, line 134: I would define this as a generic `ARM64` instead of making it specific for macOS.
Regardless, since it's only being used once across the entire code base, I think you can simply use `platform.machine()` later, in the test itself.

psutil/tests/__init__.py
```diff
@@ -901,6 +903,18 @@ def subTest(self, *args, **kw):
...
+def expectedFailureIf(condition):
+    """The test is marked as an expectedFailure if the condition is
+    satisfied.
+    """
+    def wrapper(func):
+        if condition:
+            return unittest.expectedFailure(func)
+        else:
+            return func
+    return wrapper
```
giampaolo, line 915: No need for this in my opinion.

psutil/tests/test_osx.py
```diff
@@ -144,6 +146,7 @@ def test_cpu_count_cores(self):
         num = sysctl("sysctl hw.physicalcpu")
         self.assertEqual(num, psutil.cpu_count(logical=False))
 
+    @expectedFailureIf(MACOS_ARM64)  # macOS arm64 not supported: issue #1892
```
giampaolo, line 149: See my comment above: there's no need of an additional decorator. Please define this as:

```python
# TODO: remove this once 1892 is fixed
@unittes.skipIf(MACOS and platform.machine() == 'arm64', "skipped due to #1892")
def test_cpu_freq(self):
    ...
```

giampaolo (changes requested):

giampaolo:
If this is still valid, there are some comments I added that should be addressed.

PR #2151, opened 2022-09-29
Resolve race condition in Process.threads()

### LINKED ISSUE #2150: [Linux] Race condition in Process.threads()
## Summary

* OS: Linux
* Architecture: 64bit
* Psutil version: 5.9.3
* Python version: 3.10.4
* Type: core

## Description

`Process.threads()` has a race condition in this code: https://github.com/giampaolo/psutil/blob/69b572ef62ff349495d7884e231ed9faec7775c6/psutil/_pslinux.py#L2061-L2068

A thread may exit after the `open_binary()` call and before the `read()` call. In this case `open_binary()` will succeed but `read()` will raise `ProcessLookupError`.

I think this should be treated the same as `FileNotFoundError` from `open_binary()`, since in both cases it means the thread disappeared on us.

Here's some code I used to trigger the race condition:

```python
import psutil
import random
import threading
import time

def noop():
    time.sleep(random.random() * 10)

def generate_threads():
    for _ in range(10):
        threads = []
        for _ in range(100):
            thread = threading.Thread(target=noop)
            thread.start()
            threads.append(thread)
        for thread in threads:
            thread.join()

t = threading.Thread(target=generate_threads)
t.start()

while t.is_alive():
    psutil.Process().threads()
```

li-dan:
Process.threads() has a race condition triggered when a thread exits after the open_binary() call and before the read() call. When this happens, the read() call raises ProcessLookupError.

Handle the race condition by catching ProcessLookupError from read() and treating the same as a FileNotFoundError from open_binary(). This is the same approach used in ppid_map().

Signed-off-by: Daniel Li <daniel.li@deshaw.com>

## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: #2150

## Description

Fix the race condition in `Process.threads()` using the same approach as `ppid_map()`.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -798,3 +798,6 @@ N: Bernhard Urban-Forster
 C: Austria
 W: https://github.com/lewurm
 I: 2135
+
+N: Daniel Li
+I: 2150
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,8 @@ XXXX-XX-XX
   undefined ``ethtool_cmd_speed`` symbol.
 - 2142_, [POSIX]: `net_if_stats()`_ 's ``flags`` on Python 2 returned unicode
   instead of str.  (patch by Matthieu Darbois)
+- 2150_, [Linux] `Process.threads()`_ may raise ``NoSuchProcess``. Fix race
+  condition.  (patch by Daniel Li)
 
 5.9.2
 =====
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -2217,7 +2217,7 @@ class Process(object):
                         with open_binary(file) as f:
                             pos = int(f.readline().split()[1])
                             flags = int(f.readline().split()[1], 8)
-                    except FileNotFoundError:
+                    except (FileNotFoundError, ProcessLookupError):
                         # fd gone in the meantime; process may
                         # still be alive
                         hit_enoent = True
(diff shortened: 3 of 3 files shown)


giampaolo:
I see there is the same problem in `open_files()` method. While you're at it, could you please fix that one as well?
Thanks a lot

li-dan:
Are you referring to these lines? https://github.com/giampaolo/psutil/blob/69b572ef62ff349495d7884e231ed9faec7775c6/psutil/_pslinux.py#L2213-L2220

If the file descriptor disappears between the `open_binary()` and `readline()` calls, then `readline()` raises `FileNotFoundError`. So I believe that `except FileNotFoundError` is correct in `open_files()`.

giampaolo:
Yes I mean those lines. How do you know `readline()` can only raise `FileNotFoundError`? I think calling `readline()` should be like calling `read()`.

li-dan:
In my testing, it seems that `.../fdinfo/...` behaves differently from `.../task/...`. For example:

```python
f1 = open("/etc/motd")
with open(f"/proc/{os.getpid()}/fdinfo/{f1.fileno()}", "rb") as f2:
    f1.close()
    f2.readline() # Raises FileNotFoundError
```

It's possible that some other Linux kernel will do something different, and it won't hurt either way, so I changed `open_files()` too.

giampaolo:
Yeah let's just add it to be sure.  Thanks for your PR.

PR #2154, opened 2022-10-10
Add support for force thermal_zone info in sensors_temperatures().

dargor0:
## Summary

* OS: Linux armv7l 5.15.25
* Bug fix: yes
* Type: core
* Fixes: 1873

## Description

Add an optional argument to `sensors_temperatures()`: `force_thermalzone` (default: `False`), if `True`, it will include temperatures from `/sys/class/thermal/thermal_zone*` even if there is temperatures from `hwmon`. This option only has sense for Linux OS, the argument is ignored on other OS.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1298,7 +1298,7 @@ def disk_partitions(all=False):
 # =====================================================================
 
 
-def sensors_temperatures():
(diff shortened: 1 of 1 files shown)


giampaolo:
Doesn't this mean that this way we may get duplicate entries?
That aside, the PR as-is is not OK since it adds a new function parameter.

PR #2162, opened 2022-10-22
chore: build / test Python 3.5 on Linux

mayeut:
## Summary

* OS: Linux
* Bug fix: no
* Type: tests, wheels
* Fixes:

## Description

As mentioned in https://github.com/giampaolo/psutil/pull/2102#issuecomment-1284617987, given psutil is still supported for python 3.4/3.5 it probably makes sense to test them.

Python 3.4 has a very low download count for psutil 5.9 but 3.5 still has some (also low but does not report as 0.00% by `pypinfo`). This PR adds wheel building/testing on Linux for CPython 3.5

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,9 +103,9 @@ jobs:
         python ../psutil/tests/test_memleaks.py
       shell: bash
 
-  # Linux + macOS + Python 2
-  linux-macos-py2:
-    name: ${{ matrix.os }}-py2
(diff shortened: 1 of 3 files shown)


giampaolo:
Hi Matthew. I'm gonna close this PR because Github just informed me that:

>  You've used 75% of included services for GitHub Actions. To continue using Actions & Packages uninterrupted, update your spending limit.

As such is better to limit the amount of CI test runners.

PR #2168, opened 2022-11-07
fix: long-description on Windows

mayeut:
## Summary

* OS: Windows
* Bug fix: yes
* Type: wheels
* Fixes: 

## Description
Not sure why the Popen communication is failing on GHA but it inserts a bunch of newlines that shouldn't be there.
This PR uses a temporary file rather than communicating through Popen.

### DIFF
--- a/scripts/internal/convert_readme.py
+++ b/scripts/internal/convert_readme.py
@@ -44,7 +44,11 @@ def main():
         data = f.read()
     data = re.sub(r".. raw:: html\n+\s+<div align[\s\S]*?/div>", summary, data)
     data = re.sub(r"Sponsors\n========[\s\S]*?Example usages", funding, data)
-    print(data)
+    if len(sys.argv) > 2:
+        with open(sys.argv[2], "wb") as f:
+            f.write(data.encode("utf8"))
+    else:
+        print(data)
 
 
 if __name__ == '__main__':
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -201,3 +201,14 @@ jobs:
         python3 -m isort .
         # clinter
         find . -type f \( -iname "*.c" -o -iname "*.h" \) | xargs python3 scripts/internal/clinter.py
+
+  check_dist:
+    name: Check dist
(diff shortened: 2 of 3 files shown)


mayeut:
@giampaolo, it's probably not as clean is I'd want but thought I'd share this anyway as it's getting late and I probably won't be able to work again on this before the week-end.

scripts/internal/convert_readme.py
```diff
@@ -44,7 +44,11 @@ def main():
         data = f.read()
     data = re.sub(r".. raw:: html\n+\s+<div align[\s\S]*?/div>", summary, data)
     data = re.sub(r"Sponsors\n========[\s\S]*?Example usages", funding, data)
-    print(data)
+    if len(sys.argv) > 2:
+        with open(sys.argv[2], "wb") as f:
+            f.write(data.encode("utf8"))
```
giampaolo, line 49: Couldn't you simply do `data.encode("utf8")` from setup.py and avoid editing this file?

giampaolo:
Merging as-is. I will try to fix the `encoding()` stuff in a separate commit.

scripts/internal/convert_readme.py
ddelange: @mayeut this one is [missing](https://github.com/ddelange/psutil/actions/runs/3442949410/jobs/5744010265#step:6:178) `import sys`

mayeut: It wasn't missing in my commit (since `sys` was already used to read the first argument) so I don't know what went wrong on master.

ddelange: https://github.com/giampaolo/psutil/commit/ad8426c463b24b4b39b722b4d28949f2080182fc

giampaolo: That's my fault actually. I manually edited @mayeut's PR to get rid of some parts, then merged it.

giampaolo:
Ok, hopefully https://github.com/giampaolo/psutil/commit/c0aadb18038f2ce29dbdd9cc11be072b15fd34eb should fix this once and for all.

PR #2179, opened 2022-11-28
Use GitHub Actions concurrency control

altendky:
## Summary

* OS: n/a
* Bug fix: no
* Type: wheels
* Fixes:

## Description

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

GitHub Actions offers their own built in concurrency control.
This is just an offer in case you are interested in switching over.
It includes the commit hash in the 'concurrency group' if the run is for the default branch so that the default branch builds do not get cancelled.


giampaolo:
Interesting, thank you.  Just one question. What does this line do? I can't understand just git reading GH doc.

```
group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }}
```

altendky:
Sure, it's not the...  clearest mud ever.  Basically, you need to in some way specify what other workflow runs should get cancelled.  This is intended to create a group for each PR, for each branch, and then a unique "group" for each commit to the default branch so on `master` you never cancel a run.

The never-cancel-`master` bit is `${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }}`.  That says that if the ref that triggered this is equal to `refs/heads/master` then insert the SHA of the triggering commit so that every commit is unique.  Otherwise insert nothing (empty string) so that in other cases, such as PRs, old runs for the PR will be in the same group and get cancelled.

I could write more if it isn't clear quite yet, let me know.

Side note, I personally also set the triggers under `on:` to only build the default branch for `push` and then just require a PR for all other cases.  That would 1) avoid 'duplicate' `push`/`pull_request` runs but 2) require that you create a PR to get builds instead of just pushing to a branch.  Though, `push` and `pull_request` triggers aren't exactly the same.  You have to build on `pull_request` to process external PRs.  For your own that you can push to the repo though, a `push` trigger builds just exactly that commit while a `pull_request` trigger first creates a merge of the PR branch with the target branch and then runs the workflow for that.

giampaolo:
> The never-cancel-master bit is ${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }}. That says that if the ref that triggered this is equal to refs/heads/master then insert the SHA of the triggering commit so that every commit is unique. Otherwise insert nothing (empty string) so that in other cases, such as PRs, old runs for the PR will be in the same group and get cancelled.
> I could write more if it isn't clear quite yet, let me know.

OK, this is clear, thanks for the detailed explanation. Could you just add a comment on top? Something like:

```
# -${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && github.sha || '' }} == "never cancel builds on master"
```

...?


> Side note, I personally also set the triggers under on: to only build the default branch for push and then just require a PR for all other cases. 

Clear. I think I don't want this though. I often want to trigger a build while I'm working in a branch (which is not yet a PR) because I want to see how tests run on all platforms.

altendky:
Yeah, I like to have a PR for everything more or less.  I just use the GitHub 'draft PR' feature until it is ready.  Anyways...  Adding the comment now.  I'll use a few more words, but not too many.  `:]`

giampaolo:
https://github.com/giampaolo/psutil/actions/runs/3874621010

```
[Invalid workflow file: .github/workflows/build.yml#L29](https://github.com/giampaolo/psutil/actions/runs/3874621010/workflow)
You have an error in your yaml syntax on line 29
```

giampaolo:
Closing out as outdated. If you want to give this another shot feel free to update the PR.

Issue #79, opened 2014-05-23
Get files opened by a process


giampaolo:
_From [billiej...@gmail.com](https://code.google.com/u/107489592441517285058/) on March 08, 2010 20:49:57_

```
Considering that there's nothing like lsof implemented in python yet, it
would be very cool if psutil could provide a method to retrieve all the
files opened by a process as a list of path names.

Example:

>>> import psutil, os
>>> p = psutil.Process(os.getpid())
>>> p.get_open_files()
['/etc/samba.conf', '/home/user/file.txt', ...]


I've found a working code for Windows, thanks to wj32 (we should really
thank the guy): http://forum.sysinternals.com/forum_posts.asp?TID=18892 ...which I managed to adapt to psutil in some way.

This is still not fully reliable as it lacks the routine to convert
"\Device\HarddiskVolume1\" into actual drive letters and a proper exception
handling, but I'd say it looks promising.
```

_Original issue: http://code.google.com/p/psutil/issues/detail?id=79_

PR #2190, opened 2023-01-07
Windows: Use GetFinalPathNameByHandle for Process.open_files

### LINKED ISSUE #1967: [OS] `Process.open_files()` unsafe API usage can lead to deadlock
## Summary

* OS: Windows
* Architecture: All
* Psutil version: 5.8.0
* Python version: 3.8.10
* Type: performance

## Description

The design of `Process.open_files()` has a serious design flaw that can lead to a deadlock.  Per [the psutil documentation](https://psutil.readthedocs.io/en/latest/#psutil.Process.open_files):

> Warning on Windows this method is not reliable due to some limitations of the underlying Windows API which may hang when retrieving certain file handles. In order to work around that psutil spawns a thread to determine the file handle name and kills it if it’s not responding after 100ms. That implies that this method on Windows is not guaranteed to enumerate all regular file handles (see issue 597). Tools like ProcessHacker has the same limitation.

Beyond this risk of missing data, this API also carries a deadlock risk.

The thread monitoring/termination logic is implemented here: https://github.com/giampaolo/psutil/blob/ac898c7c8cd4d232c602b70d58fc53474e8b9154/psutil/arch/windows/process_handles.c#L166-L177

The danger is that if the target thread is terminated while it is holding a lock, it can deadlock the entire process.  MSDN warns about this [here](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminatethread):

> TerminateThread is a dangerous function that should only be used in the most extreme cases. You should call TerminateThread only if you know exactly what the target thread is doing, and you control all of the code that the target thread could possibly be running at the time of the termination. For example, TerminateThread can result in the following problems:
> 
> - If the target thread owns a critical section, the critical section will not be released.
> - If the target thread is allocating memory from the heap, the heap lock will not be released.

In our environment, we were seeing deadlocks in python.exe where something had taken the loader lock then died.  Unfortunately the thread was gone by the time we could inspect it with a debugger, leading to cryptic deadlocks, but [`!locks`](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/debugging-a-deadlock) indicated that the loader lock was owned by a non-existent thread.

```
0:000> !locks

CritSec ntdll!LdrpLoaderLock+0 at 00000000772580d8
WaiterWoken        No
LockCount          203
RecursionCount     1
OwningThread       4c4
EntryCount         0
ContentionCount    d0
*** Locked

Scanned 687 critical sections
```
There was no thread `4c4` when we were able to dump that process.

Note that the loader lock is not the only lock that can cause a deadlock.  In fact, many times when this deadlock occurs, `!locks` will not indicate anything abnormal.

## Reproduction

To reproduce this, do the following on Windows (any version, really):

1. Start the process that will hang.
```
start /low py -c "import os,psutil; print('Wait for PID %u CPU to drop to 0%% in Task Manager' % os.getpid()); [[p.as_dict() for p in psutil.process_iter()] for x in range(99999)]"
```
2. Open Task Manager and find that process.  It should be using most of one CPU core.  For example, it uses approximately 8% on my 6C/12T machine.
3. Now we need to slow down that process to increase the chances of the race condition occurring.  Run the following to max out all your CPU cores:
```
py -c "import os,multiprocessing; [os.system('start python -c \"while True: pass\"') for i in range(multiprocessing.cpu_count())]"
```
4. Wait a few seconds.  The process from step 1 should drop to 0% CPU.  You can now kill all the processes started in step 3.
5. If the process from step 1 resumes using CPU, repeat steps 3 & 4 a few times.  The reproduction is probabilistic but is very repeatable on my machine.

Using [PyExt](https://github.com/SeanCline/PyExt) for symbols, we can see the call stack hung in `open_files`:
```
0:000> !pyext.pysymfix
Current symbol path: SRV*C:\Symbols*https://msdl.microsoft.com/download/symbols;srv*http://pythonsymbols.sdcline.com/symbols
Loading symbols...

0:000> .shell -ci "~* kP" findstr "qualname"
			struct _object * qualname = 0x000001d6`63ee9670Type: str 'Process.open_files')+0x228 [C:\A\34\s\Python\ceval.c @ 4298] 
			struct _object * qualname = 0x000001d6`63c92270Type: str 'Process.as_dict')+0x228 [C:\A\34\s\Python\ceval.c @ 4298] 
			struct _object * qualname = 0x000001d6`63a980d0Type: str '<listcomp>.<listcomp>')+0x228 [C:\A\34\s\Python\ceval.c @ 4298] 
			struct _object * qualname = 0x00000000`00000000)+0x228 [C:\A\34\s\Python\ceval.c @ 4298] 
.shell: Process exited
```

jschwartzentruber:
## Summary

* OS: Windows
* Bug fix: yes
* Type: core
* Fixes: #1967 

## Description

This uses `GetFinalPathNameByHandle` which was added in Vista/Server 2008. This fixes #1967 based on the steps to reproduce there (only tested on Windows 11).

### DIFF
--- a/psutil/arch/windows/process_handles.c
+++ b/psutil/arch/windows/process_handles.c
@@ -193,6 +193,7 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
     PSYSTEM_HANDLE_INFORMATION_EX       handlesList = NULL;
     PSYSTEM_HANDLE_TABLE_ENTRY_INFO_EX  hHandle = NULL;
     HANDLE                              hFile = NULL;
+    HANDLE                              hSafeFile = NULL;
     ULONG                               i = 0;
     BOOLEAN                             errorOccurred = FALSE;
     PyObject*                           py_path = NULL;
@@ -222,14 +223,17 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
             continue;
         }
 
-        if (GetFileType(hFile) != FILE_TYPE_DISK) {
-            SetLastError(0);
-            CloseHandle(hFile);
-            hFile = NULL;
-            continue;
-        }
+        hSafeFile = ReOpenFile(
+            hFile,
+            0,
+            FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+            0);
+        CloseHandle(hFile);
 
-        fileName = psutil_threaded_get_filename(hFile);
+        if (hSafeFile == INVALID_HANDLE_VALUE)
+            continue;
+
+        fileName = psutil_threaded_get_filename(hSafeFile);
         if (fileName == NULL)
             goto error;
 
@@ -248,8 +252,8 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
             FREE(fileName);
             fileName = NULL;
         }
-        CloseHandle(hFile);
-        hFile = NULL;
+        CloseHandle(hSafeFile);
+        hSafeFile = NULL;
     }
 
     goto exit;
@@ -260,8 +264,8 @@ error:
     goto exit;
 
 exit:
-    if (hFile != NULL)
-        CloseHandle(hFile);
+    if (hSafeFile != NULL)
+        CloseHandle(hSafeFile);
     if (fileName != NULL)
         FREE(fileName);
     if (py_path != NULL)
--- a/psutil/arch/windows/process_handles.c
+++ b/psutil/arch/windows/process_handles.c
@@ -27,8 +27,15 @@
 
 
 #define THREAD_TIMEOUT 100  // ms
-// Global object shared between the 2 threads.
-PUNICODE_STRING globalFileName = NULL;
+#define STATUS_UNSUCCESSFUL ((DWORD)0xC0000001L)
+
+typedef struct
+{
+  NTSTATUS status;
+  HANDLE hFile;
+  PUNICODE_STRING FileName;
+  ULONG Length;
+} psutil_query_thread;
 
 
 static int
@@ -80,129 +87,104 @@ psutil_enum_handles(PSYSTEM_HANDLE_INFORMATION_EX *handles) {
     return 0;
 }
 
-
-static int
+static DWORD
 psutil_get_filename(LPVOID lpvParam) {
-    HANDLE hFile = *((HANDLE*)lpvParam);
-    NTSTATUS status;
-    ULONG bufferSize;
+    psutil_query_thread* context = lpvParam;
+
+    context->status = NtQueryObject(
+        context->hFile,
+        ObjectNameInformation,
+        context->FileName,
+        context->Length,
+        &context->Length);
+    return 0;
+}
+
+
+static PUNICODE_STRING
+psutil_threaded_get_filename(HANDLE hFile) {
+    DWORD dwWait;
+    HANDLE hThread;
+    DWORD result = 0;
     ULONG attempts = 8;
+    psutil_query_thread threadContext;
 
-    bufferSize = 0x200;
-    globalFileName = MALLOC_ZERO(bufferSize);
-    if (globalFileName == NULL) {
+    threadContext.status = STATUS_UNSUCCESSFUL;
+    threadContext.hFile = hFile;
+    threadContext.FileName = NULL;
+    threadContext.Length = 0x200;
+    threadContext.FileName = MALLOC_ZERO(threadContext.Length);
+    if (threadContext.FileName == NULL) {
         PyErr_NoMemory();
-        goto error;
-    }
-
-
-    // Note: also this is supposed to hang, hence why we do it in here.
-    if (GetFileType(hFile) != FILE_TYPE_DISK) {
-        SetLastError(0);
-        globalFileName->Length = 0;
-        return 0;
+        return NULL;
     }
 
     // A loop is needed because the I/O subsystem likes to give us the
     // wrong return lengths...
     do {
-        status = NtQueryObject(
-            hFile,
-            ObjectNameInformation,
-            globalFileName,
-            bufferSize,
-            &bufferSize
-            );
-        if (status == STATUS_BUFFER_OVERFLOW ||
-                status == STATUS_INFO_LENGTH_MISMATCH ||
-                status == STATUS_BUFFER_TOO_SMALL)
-        {
-            FREE(globalFileName);
-            globalFileName = MALLOC_ZERO(bufferSize);
-            if (globalFileName == NULL) {
-                PyErr_NoMemory();
-                goto error;
-            }
+        hThread = CreateThread(
+            NULL, 0, (LPTHREAD_START_ROUTINE)psutil_get_filename,
+            &threadContext, 0, NULL);
+        if (hThread == NULL) {
+            PyErr_SetFromOSErrnoWithSyscall("CreateThread");
+            FREE(threadContext.FileName);
+            return NULL;
         }
-        else {
+
+        // Wait for the worker thread to finish.
+        dwWait = WaitForSingleObject(hThread, THREAD_TIMEOUT);
+
+        // If the thread hangs, kill it and cleanup.
+        if (dwWait == WAIT_TIMEOUT) {
+            psutil_debug(
+                "get handle name thread timed out after %i ms", THREAD_TIMEOUT);
+            if (TerminateThread(hThread, 0) == 0) {
+                PyErr_SetFromOSErrnoWithSyscall("TerminateThread");
+                result = 1;
+            }
+        } else if (dwWait == WAIT_FAILED) {
+            psutil_debug("WaitForSingleObject -> WAIT_FAILED");
+            result = 1;
+            if (TerminateThread(hThread, 0) == 0) {
+                PyErr_SetFromOSErrnoWithSyscall(
+                    "WaitForSingleObject -> WAIT_FAILED -> TerminateThread");
+            } else {
+                PyErr_SetFromOSErrnoWithSyscall("WaitForSingleObject");
+            }
+        } else if (GetExitCodeThread(hThread, &result) == 0) {
+            result = 1;
+            PyErr_SetFromOSErrnoWithSyscall("GetExitCodeThread");
+        } else if (threadContext.status == STATUS_BUFFER_OVERFLOW ||
+                threadContext.status == STATUS_INFO_LENGTH_MISMATCH ||
+                threadContext.status == STATUS_BUFFER_TOO_SMALL)
+        {
+            FREE(threadContext.FileName);
+            threadContext.FileName = MALLOC_ZERO(threadContext.Length);
+            if (threadContext.FileName == NULL) {
+                PyErr_NoMemory();
+                CloseHandle(hThread);
+                return NULL;
+            }
+        } else {
+            CloseHandle(hThread);
             break;
         }
-    } while (--attempts);
-
-    if (! NT_SUCCESS(status)) {
-        psutil_SetFromNTStatusErr(status, "NtQuerySystemInformation");
-        FREE(globalFileName);
-        globalFileName = NULL;
-        return 1;
-    }
-
-    return 0;
-
-error:
-    if (globalFileName != NULL) {
-        FREE(globalFileName);
-        globalFileName = NULL;
-    }
-    return 1;
-}
-
-
-static DWORD
-psutil_threaded_get_filename(HANDLE hFile) {
-    DWORD dwWait;
-    HANDLE hThread;
-    DWORD threadRetValue;
-
-    hThread = CreateThread(
-        NULL, 0, (LPTHREAD_START_ROUTINE)psutil_get_filename, &hFile, 0, NULL);
-    if (hThread == NULL) {
-        PyErr_SetFromOSErrnoWithSyscall("CreateThread");
-        return 1;
-    }
-
-    // Wait for the worker thread to finish.
-    dwWait = WaitForSingleObject(hThread, THREAD_TIMEOUT);
-
-    // If the thread hangs, kill it and cleanup.
-    if (dwWait == WAIT_TIMEOUT) {
-        psutil_debug(
-            "get handle name thread timed out after %i ms", THREAD_TIMEOUT);
-        if (TerminateThread(hThread, 0) == 0) {
-            PyErr_SetFromOSErrnoWithSyscall("TerminateThread");
-            CloseHandle(hThread);
-            return 1;
-        }
         CloseHandle(hThread);
-        return 0;
+
+    } while (--attempts && !result);
+
+    if (result) {
+        FREE(threadContext.FileName);
+        return NULL;
     }
 
-    if (dwWait == WAIT_FAILED) {
-        psutil_debug("WaitForSingleObject -> WAIT_FAILED");
-        if (TerminateThread(hThread, 0) == 0) {
-            PyErr_SetFromOSErrnoWithSyscall(
-                "WaitForSingleObject -> WAIT_FAILED -> TerminateThread");
-            CloseHandle(hThread);
-            return 1;
-        }
-        PyErr_SetFromOSErrnoWithSyscall("WaitForSingleObject");
-        CloseHandle(hThread);
-        return 1;
+    if (! NT_SUCCESS(threadContext.status)) {
+        psutil_SetFromNTStatusErr(threadContext.status, "NtQuerySystemInformation");
+        FREE(threadContext.FileName);
+        return NULL;
     }
 
-    if (GetExitCodeThread(hThread, &threadRetValue) == 0) {
-        if (TerminateThread(hThread, 0) == 0) {
-            PyErr_SetFromOSErrnoWithSyscall(
-                "GetExitCodeThread (failed) -> TerminateThread");
-            CloseHandle(hThread);
-            return 1;
-        }
-
-        CloseHandle(hThread);
-        PyErr_SetFromOSErrnoWithSyscall("GetExitCodeThread");
-        return 1;
-    }
-    CloseHandle(hThread);
-    return threadRetValue;
+    return threadContext.FileName;
 }
 
 
@@ -214,14 +196,11 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
     ULONG                               i = 0;
     BOOLEAN                             errorOccurred = FALSE;
     PyObject*                           py_path = NULL;
-    PyObject*                           py_retlist = PyList_New(0);;
+    PyObject*                           py_retlist = PyList_New(0);
+    PUNICODE_STRING                     fileName = NULL;
 
     if (!py_retlist)
         return NULL;
-
-    // Due to the use of global variables, ensure only 1 call
-    // to psutil_get_open_files() is running.
-    EnterCriticalSection(&PSUTIL_CRITICAL_SECTION);
 
     if (psutil_enum_handles(&handlesList) != 0)
         goto error;
@@ -243,13 +222,20 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
             continue;
         }
 
-        // This will set *globalFileName* global variable.
-        if (psutil_threaded_get_filename(hFile) != 0)
+        if (GetFileType(hFile) != FILE_TYPE_DISK) {
+            SetLastError(0);
+            CloseHandle(hFile);
+            hFile = NULL;
+            continue;
+        }
+
+        fileName = psutil_threaded_get_filename(hFile);
+        if (fileName == NULL)
             goto error;
 
-        if ((globalFileName != NULL) && (globalFileName->Length > 0)) {
-            py_path = PyUnicode_FromWideChar(globalFileName->Buffer,
-                                             wcslen(globalFileName->Buffer));
+        if ((fileName != NULL) && (fileName->Length > 0)) {
+            py_path = PyUnicode_FromWideChar(fileName->Buffer,
+                                             wcslen(fileName->Buffer));
             if (! py_path)
                 goto error;
             if (PyList_Append(py_retlist, py_path))
@@ -258,9 +244,9 @@ psutil_get_open_files(DWORD dwPid, HANDLE hProcess) {
         }
 
         // Loop cleanup section.
-        if (globalFileName != NULL) {
-            FREE(globalFileName);
-            globalFileName = NULL;
+        if (fileName != NULL) {
+            FREE(fileName);
+            fileName = NULL;
         }
         CloseHandle(hFile);
         hFile = NULL;
@@ -276,16 +262,13 @@ error:
 exit:
     if (hFile != NULL)
         CloseHandle(hFile);
-    if (globalFileName != NULL) {
-        FREE(globalFileName);
-        globalFileName = NULL;
-    }
+    if (fileName != NULL)
+        FREE(fileName);
     if (py_path != NULL)
         Py_DECREF(py_path);
     if (handlesList != NULL)
         FREE(handlesList);
 
-    LeaveCriticalSection(&PSUTIL_CRITICAL_SECTION);
     if (errorOccurred == TRUE)
         return NULL;
     return py_retlist;


jschwartzentruber:
I'll try to reproduce the CI failures on other Windows versions.

giampaolo:
Interesting. FYI, `open_files()` on Windows is a bug which goes back to [9 years ago](https://github.com/giampaolo/psutil/issues/79) and never got fully resolved. Searching for [NtQueryObject](https://github.com/giampaolo/psutil/search?q=NtQueryObject&type=issues) also gives an idea on the amount troubles caused by this API over the years. Some considerations:

1) according to this [old comment](https://github.com/giampaolo/psutil/issues/340#issuecomment-59365752), which is of course unverified,  `GetFileInformationByHandle` would still block, and you should use `GetFileInformationByHandleEx` instead. Also I see there are some interesting comments to the original [blog post](http://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html), which of course are also unverified.
2) ProcessHacker is probably the best source of truth here. I looked at ProcessHacker [source code](https://github.com/winsiderss/systeminformer/blob/38e95922b84d5c52ee433b1fbd154192e3e33cab/SystemInformer/hndlprv.c) and unfortunately it doesn't use `GetFinalPathNameByHandle/Ex`, it uses `NtQueryObject`, like we do, so that makes me skeptical. CC @dmex / @wj32 in case they have some comment on this (and I hope you don't mind my CC). 

Just to make things clear, a proper solution would have to:

1) NOT hang, and to my understanding the current code should not because we spawn threads
2) fix this bug https://github.com/giampaolo/psutil/issues/1967, which currently affects us

Pololot64:
I have been looking into this as well. I normally do pure python but does the deadlock refer to the computer hanging for about half a second every time open_files() is used? It seems to me that NtQuerySystemInformation is what causes the hang even with this fixed code.

jschwartzentruber:
> according to this [old comment](https://github.com/giampaolo/psutil/issues/340#issuecomment-59365752), which is of course unverified,  `GetFileInformationByHandle` would still block, and you should use `GetFileInformationByHandleEx` instead. Also I see there are some interesting comments to the original [blog post](http://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html), which of course are also unverified.

I looked into this, including the upstream [bug report](https://web.archive.org/web/20170301214039/https://connect.microsoft.com/VisualStudio/feedback/details/362169/getfinalpathnamebyhandle-api-hungs), and it was closed by MS as "by design", the description "for certain type of handles", and one of the comments says to use `GetFileType` to check only disk files. This all leads me to believe the author was not filtering for pipes or directory handles.

> it uses `NtQueryObject`, like we do

I'm also curious about this. Is it possible that `NtQueryObject` is fine, and psutil should just filter handles differently? Even so, it would be better to use a supported API as long as it doesn't regress functionality or add bugs.

> NOT hang, and to my understanding the current code should not because we spawn threads

I will add back the thread wrapper. I removed it optimistically, but I'm sure there would be situations where even a "fixed" API could hang (slow/broken SMB connections, etc.).

jschwartzentruber:
> does the deadlock refer to the computer hanging for about half a second every time open_files() is used? It seems to me that NtQuerySystemInformation is what causes the hang even with this fixed code.

This is where the call never completes, and is unrecoverable. That's the issue described in #1967 and what is affecting me.

If you're trying my fix, I prematurely removed the threading code, so that may re-introduce #340. I'm going to add the thread back but around the newer API and test that on older platforms.

giampaolo:
>  Is it possible that NtQueryObject is fine, and psutil should just filter handles differently? 

Yes, it is possible that psutil's use of [GetFileType](https://github.com/giampaolo/psutil/blob/2da99502238852f18f74db05304e67a01ee77005/psutil/arch/windows/process_handles.c#L99-L100), is the problem (or **a** problem). If I'm reading the code right, ProcessHacker uses [NtQueryObject with ObjectTypesInformation](https://github.com/winsiderss/systeminformer/blob/38e95922b84d5c52ee433b1fbd154192e3e33cab/phlib/hndlinfo.c#L1981-L1988) instead, so I **definitively** think this is very well worth a try. 

Also, there's an interesting comment in PH [source code](https://github.com/winsiderss/systeminformer/blob/38e95922b84d5c52ee433b1fbd154192e3e33cab/SystemInformer/hndlprv.c#L303-L306):

```c
    // There are three ways of enumerating handles:
    // * On Windows 8 and later, NtQueryInformationProcess with ProcessHandleInformation is the most efficient method.
    // * On Windows XP and later, NtQuerySystemInformation with SystemExtendedHandleInformation.
    // * Otherwise, NtQuerySystemInformation with SystemHandleInformation can be used.
```

We use [NtQuerySystemInformation with SystemExtendedHandleInformation](https://github.com/giampaolo/psutil/blob/2da99502238852f18f74db05304e67a01ee77005/psutil/arch/windows/process_handles.c#L48-L49) (second method) whereas ProcessHacker uses [NtQueryInformationProcess with ProcessHandleInformation](https://github.com/winsiderss/systeminformer/blob/38e95922b84d5c52ee433b1fbd154192e3e33cab/phlib/native.c#L6502-L6504) on Windows 8+ (first method). Probably this will only make a difference in terms of performance, but may also be worth a try.

dmex:
> I looked at ProcessHacker source code and unfortunately it doesn't use GetFinalPathNameByHandle/Ex, it uses NtQueryObject, like we do

@giampaolo

Process Hacker uses the Native API instead of the Win32 API. GetFinalPathNameByHandle/Ex are both Win32 APIs exposing a limited number of information classes from the [NtQueryInformationFile](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntqueryinformationfile) and [NtQueryDirectoryFile](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntquerydirectoryfile) Native APIs.

GetFileInformationByHandleEx is used by Process Hacker but you need to search for NtQueryInformationFile 😉

> Yes, it is possible that psutil's use of [GetFileType](https://github.com/giampaolo/psutil/blob/2da99502238852f18f74db05304e67a01ee77005/psutil/arch/windows/process_handles.c#L99-L100) is the problem (or a problem). ProcessHacker doesn't use it (which is already a red flag).

GetFileType is another Win32 API and it's internally calling the [NtQueryVolumeInformationFile](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntqueryvolumeinformationfile) Native API with the [FileFsDeviceInformation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_file_fs_device_information) information class.

Process Hacker does use GetFileType but you need to search for NtQueryVolumeInformationFile and FileFsDeviceInformation 😉

We've never run into issues with GetFileType/NtQueryVolumeInformationFile(FileFsDeviceInformation) and it's always successfully returned the DeviceType for the handle without deadlocking unlike NtQueryObject and NtQueryInformationFile: https://github.com/winsiderss/systeminformer/blob/f17d0dc7c770fb2453bb604c39a43508bd692e07/SystemInformer/hndlprp.c#L1337-L1374

> it uses NtQueryObject, like we do, so that makes me skeptical
> This blog post (...) suggests using GetFileInformationByHandleEx

NtQueryInformationFile (GetFinalPathNameByHandle/Ex) deadlocks just like NtQueryObject does especially on Windows 10/11 for named pipes (FILE_DEVICE_NAMED_PIPE), console handles (FILE_DEVICE_CONSOLE) and some other device types.

You can see here how we've had to use the same workaround for NtQueryObject with NtQueryInformationFile here:
https://github.com/winsiderss/systeminformer/blob/f17d0dc7c770fb2453bb604c39a43508bd692e07/phlib/hndlinfo.c#L2314-L2323
and here:
https://github.com/winsiderss/systeminformer/blob/f17d0dc7c770fb2453bb604c39a43508bd692e07/phlib/hndlinfo.c#L2425-L2442
and here:
https://github.com/winsiderss/systeminformer/blob/f17d0dc7c770fb2453bb604c39a43508bd692e07/SystemInformer/hndlprp.c#L1169-L1172

Note: These days we're mostly avoiding issues with NtQueryObject by using our kernel driver to query the name and all this code is only used as fallback when the driver isn't available.

> fix this bug [OS] Process.open_files() unsafe API usage can lead to deadlock #1967, which currently affects us
> The design of Process.open_files() has a serious design flaw that can lead to a deadlock.

`ntdll!LdrpLoaderLock` is only used when loading a DLL with functions like [LoadLibrary](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryw) and LdrLoadDll. The `psutil_get_filename` thread does not call any loader functions so it doesn't make sense how terminating the thread leaks the LdrpLoaderLock? 

I tried debugging the psutil code and I think it's very likely the calls to `psutil_SetFromNTStatusErr()`, `PyErr_NoMemory()` and `PyObject_CallFunction` inside the `psutil_get_filename` function causing issues with thread termination. If these functions attempt to convert the error code to strings (or otherwise use FormatMessageW) then it's using the loader to locate the kernel32/ntdll images and copy the RT_MESSAGETABLE resource and the thread was terminated

If it's not those functions then another likely reason is because the `psutil_get_filename` function can be terminated while calling HeapAlloc/HeapFree and leaking memory but more importantly leaking the lock for the process default heap. Leaking the heap lock would block the entire process and most system functions from allocating/freeing memory and manifest as unrelated failures elsewhere in the application.

For example: The LoadLibrary function uses worker threads when loading DLLs on Windows 8/10/11 to improve the performance of import table snaps. A single worker thread terminating because it couldn't access the process heap would leak the LdrpLoaderLock and explain the exited thread (worker factory) referencing the LdrpLoaderLock in that other github issue (and obfuscate the real issue which would be using heap functions while terminating the `psutil_get_filename` thread).

There are also some other issues with the `psutil_get_filename` function. For reference here is the code for psutil_get_filename:
https://github.com/giampaolo/psutil/blob/2da99502238852f18f74db05304e67a01ee77005/psutil/arch/windows/process_handles.c#L84-L147

Compare `psutil_get_filename` to our `PhpCommonQueryObjectRoutine`:
https://github.com/winsiderss/systeminformer/blob/f17d0dc7c770fb2453bb604c39a43508bd692e07/phlib/hndlinfo.c#L2280C29-L2296 

Note how our function doesn't allocate memory for the string and doesn't do anything except call NtQueryObject and pass values directly into a structure? It has to be done this way to safely terminate the thread. The `psutil_get_filename` function needs to stripped down the exact same way otherwise you'll keep running into problems after terminating the thread.

* You need to remove `HeapAlloc`, `HeapFree`, `psutil_SetFromNTStatusErr`, `PyErr_NoMemory` from `psutil_get_filename` and into `psutil_threaded_get_filename` to make the `psutil_get_filename` function safe to terminate. 
* Also move the `do {  } while (--attempts)` loop from `psutil_get_filename` into `psutil_threaded_get_filename` (Yes, you need to call CreateThread multiple times for each different memory allocation)
* Remove SetLastError() from psutil_get_filename. There's no point calling SetLastError and immediately exiting the thread.
* Remove `globalFileName` from the global scope and use a structure. You can pass multiple objects via a structure with CreateThread and access those same objects once the thread exits or terminates.
* Delete `EnterCriticalSection(&PSUTIL_CRITICAL_SECTION);`... The structure you pass into CreateThread correctly shares the data between those two threads and this critical section is no longer required.

Here's a breakdown of the changes required with examples:
______________

1) Create a structure for the FileHandle, NTSTATUS, FileName and FileNameSize:
```
typedef struct _psutil_query_thread
{
  NTSTATUS status;
  HANDLE FileHandle;
  PUNICODE_STRING globalFileName;
  SIZE_T FileNameSize;
} psutil_query_thread;
```

This is what the `psutil_get_filename` function should be doing and nothing else:

```
static int psutil_get_filename(LPVOID lpvParam) 
{
    psutil_query_thread* context = lpvParam;

     context->status = NtQueryObject(
            context->FileHandle,
            ObjectNameInformation,
            context->globalFileName,
            context->FileNameSize,
            &context->FileNameSize
            );

    return 0;
}
```


Now the `psutil_threaded_get_filename` function should be written like this:


* (optional) First check if the handle is a file. Fail if the handle isn't a file.

```
 if (GetFileType(hFile) != FILE_TYPE_DISK)
{
        return 1;
}
```


* Start our query loop...

```
    do
    {
```

* Allocate the structure before calling CreateThread.

```
psutil_query_thread* threadcontext = malloc()
threadcontext->status = STATUS_UNSUCCESSFUL; // initialize with a failure status
threadcontext->FileHandle = hFile;
threadcontext->globalFileName = malloc(1234);
threadcontext->FileNameSize = 1234;
```

* Pass the structure into the `psutil_get_filename` function using CreateThread:

```
HANDLE threadHandle = CreateThread((LPTHREAD_START_ROUTINE)psutil_get_filename, threadcontext);
```

* Wait for the thread to terminate or timeout:

```
WaitForSingleObject(threadHandle, THREAD_TIMEOUT);
```

// Optionally check the thread exit code. It will only return 0 after NtQueryObject completes otherwise STILL_ACTIVE.
GetExitCodeThread() == STILL_ACTIVE


* Terminate the thread with a failure status (ensures we can safely access the threadcontext)
```
TerminateThread(threadHandle, 1);
```


* The thread was terminated. We can safely access the threadcontext and check the status. If the buffer wasn't large enough then we realloc and call CreateThread again:
```
        if (threadcontext->status == STATUS_BUFFER_OVERFLOW | threadcontext->status == STATUS_INFO_LENGTH_MISMATCH ||
            threadcontext->status== STATUS_BUFFER_TOO_SMALL)
        {
            free();
            buffer = malloc(); // Increase the buffer and call CreateThread again.
        }
        else
        {
            break; // Success... Break out of our loop.
        }
    } while (--attempts);
```

If we reach this point we've either made 8 attempts at querying the handle and still failed or we've succeeded and have a valid string.

```
if (threadcontext->status == STATUS_SUCCESS)
{
    printf("%s", threadcontext->globalFileName->Buffer)
}

free(threadcontext);
```
____________

> use GetFileType to check only disk files. This all leads me to believe the author was not filtering for pipes or directory handles.

We've never run into deadlocks with GetFileType/NtQueryVolumeInformationFile and you won't deadlock querying files or directories only when you query handles for devices like NPFS (FILE_DEVICE_NAMED_PIPE), ConDrv (FILE_DEVICE_CONSOLE) or VolMgrControl.

Process Hacker is designed to show handles so we don't have the luxury of limiting the query. If psutil `Process.open_files` is only supposed to return `files` then you should limit the filename query to FILE_DEVICE_DISK and have a separate method like `Process.open_handles` for handles. 


> CC @dmex in case they have some comment on this (I hope you won't mind my CC).

Feel free to @ me if you get stuck fixing psutil_get_filename or need more info about what Process Hacker is doing (Btw Process Hacker was renamed System Informer so I'll be referring to SI instead of PH going forward).

There are some other methods that I haven't looked into to avoid deadlocks like process reflection. These processes can't execute and we've been using reflection to avoid some other types of deadlocks with running processes and querying the reflected handle might also avoid the issue.

-dmex

giampaolo:
Thanks for the amazing write up @dmex. It's an honor to meet you here. Needless to say, PH / SI inspired a lot of the psutil development on Windows, so I want to take this opportunity to finally say hi and thank you! I believe @wj32 even contributed some code to psutil many (>10) years ago. 

> Feel free to @ me if you get stuck fixing psutil_get_filename or need more info about what Process Hacker is doing 

I appreciate it. It's a lot of info to digest so it may take a while, but the overall suggestions appear clear.

> Btw Process Hacker was renamed System Informer so I'll be referring to SI instead of PH going forward

That surprised me. PH is so etched into my brain that it will take me a while to get used to the new name. :)

giampaolo:
> If psutil Process.open_files is only supposed to return files then you should limit the filename query to FILE_DEVICE_DISK and have a separate method like Process.open_handles for handles.

Yes, we only return files (paths). No handles.

jschwartzentruber:
I've implemented the ideas from @dmex, but I still see deadlock. Although less often, it's still fairly easy to reproduce using the steps in #1967 (using 32-bit Python 3.7 on Windows Vista). I can't reproduce on 64-bit Python 3.10 Windows 11, but I couldn't using `GetFinalPathNameByHandle` either.

`!locks` still shows `LdrpLoaderLock`.

dmex:
@jschwartzentruber 

I was able completly resolve deadlock issues with NtQueryObject by using the [ReOpenFile function](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-reopenfile) to clone/recreate the internal file_object referenced by handle.

Try using this function to query the handles with psutil (It's not 100% complete but should be straightforward enough for testing):

```
BOOLEAN QueryHandleFileName(
    _In_ HANDLE ProcessHandle,
    _In_ HANDLE Handle
    )
{
    BOOLEAN handleNeedsFree = FALSE;
    HANDLE dupHandle = NULL;
    HANDLE safeHandle = NULL;

    if (ProcessHandle == GetCurrentProcess())
    {
        dupHandle = Handle;
    }
    else
    {
        if (!DuplicateHandle(
            ProcessHandle,
            Handle,
            GetCurrentProcess(),
            &dupHandle,
            0,
            FALSE,
            0
            ))
        {
            return FALSE;
        }

        handleNeedsFree = TRUE;
    }

    safeHandle = ReOpenFile(dupHandle,
        0,
        FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
        0 // FILE_FLAG_OVERLAPPED
        );

    if (safeHandle != INVALID_HANDLE_VALUE) // && (GetLastError() != ERROR_INVALID_PARAMETER)
    {
        ULONG bufferSize = sizeof(OBJECT_NAME_INFORMATION) + (MAXIMUM_FILENAME_LENGTH * sizeof(WCHAR));
        POBJECT_NAME_INFORMATION buffer = malloc(bufferSize); memset(buffer, 0, bufferSize);

        if (NT_SUCCESS(NtQueryObject(
            safeHandle,
            ObjectNameInformation,
            buffer,
            bufferSize,
            &bufferSize
            )))
        {
            printf("Success: %wZ\n", buffer);
            //return TRUE;
        }

        CloseHandle(safeHandle);
        free(buffer);
    }

    if (handleNeedsFree)
    {
        CloseHandle(dupHandle);
    }

    return FALSE;
}
```

jschwartzentruber:
> I was able to resolve the deadlock issues with NtQueryObject

@dmex thanks again! I don't see deadlock either with your change.

jschwartzentruber:
On further testing, this doesn't filter for pipes. I'm not seeing it hang/deadlock, but I do see `OSError` originated from `NtQueryObject`.

On Vista I see `[WinError 233] No process is on the other end of the pipe` and on Windows 11 I see `[WinError 50] The request is not supported`.

jschwartzentruber:
Above is visible by running `python -c "import psutil; [proc.as_dict() for proc in psutil.process_iter()]"` as administrator. Note that the current commit will say `NtQuerySystemInformation` in the exception because of copy-paste [here](https://github.com/giampaolo/psutil/pull/2190/commits/3eba2d3c6304d3bf003f0e5406c5871996a98d7a#diff-07535636fefdc3f3a7e37d05e9a9160c85c8c35eac3ef81f44a6beef723a0bbeR182), it's actually from `NtQueryObject`.

dmex:
> On further testing, this doesn't filter for pipes. I'm not seeing it hang/deadlock, but I do see OSError originated from NtQueryObject.

@jschwartzentruber That would be the expected behavior. Named pipes are created using NtCreateNamedPipeFile (npfs.sys) instead of NtCreateFile and are 'connected' by chaining each `RootDirectory` to the new handle. The ReOpenFile function creates an entirely new FILE_OBJECT that doesn't have the same RelatedFileObject/FileObjectExtension required for named pipes:

```
ntdll!_FILE_OBJECT
   +0x000 Type             : Int2B
   +0x002 Size             : Int2B
   +0x008 DeviceObject     : Ptr64 _DEVICE_OBJECT
   +0x010 Vpb              : Ptr64 _VPB
   +0x018 FsContext        : Ptr64 Void
   +0x020 FsContext2       : Ptr64 Void
   +0x028 SectionObjectPointer : Ptr64 _SECTION_OBJECT_POINTERS
   +0x030 PrivateCacheMap  : Ptr64 Void
   +0x038 FinalStatus      : Int4B
   +0x040 RelatedFileObject : Ptr64 _FILE_OBJECT
   +0x048 LockOperation    : UChar
   +0x049 DeletePending    : UChar
   +0x04a ReadAccess       : UChar
   +0x04b WriteAccess      : UChar
   +0x04c DeleteAccess     : UChar
   +0x04d SharedRead       : UChar
   +0x04e SharedWrite      : UChar
   +0x04f SharedDelete     : UChar
   +0x050 Flags            : Uint4B
   +0x058 FileName         : _UNICODE_STRING
   +0x068 CurrentByteOffset : _LARGE_INTEGER
   +0x070 Waiters          : Uint4B
   +0x074 Busy             : Uint4B
   +0x078 LastLock         : Ptr64 Void
   +0x080 Lock             : _KEVENT
   +0x098 Event            : _KEVENT
   +0x0b0 CompletionContext : Ptr64 _IO_COMPLETION_CONTEXT
   +0x0b8 IrpListLock      : Uint8B
   +0x0c0 IrpList          : _LIST_ENTRY
   +0x0d0 FileObjectExtension : Ptr64 Void
```


> If psutil Process.open_files is only supposed to return files then you should limit the filename query to FILE_DEVICE_DISK and have a separate method like Process.open_handles for handles.
> Yes, we only return files (paths). No handles.

This is why I asked if `Process.open_files` was only supposed to return files because the ReOpenFile method won't work for named pipes. There's two options:

1) GetFileType and exclude pipes (ReOpenFile is doing this already)
2) NtQueryInformationFile with FilePipeLocalInformation and check NamedPipeEnd==FILE_PIPE_CLIENT_END or FILE_PIPE_SERVER_END. The server side of the pipe blocks when using NtQueryObject and requires using a thread.

Remember that DuplicateHandle continues referencing the existing FILE_OBJECT and NtQueryObject deadlocks because the `Lock` and `Event` of the FILE_OBJECT is already inuse by the original process owning the handle. ReOpenFile creates a completely new FILE_OBJECT with a new Lock and Event that isn't shared (or held/inuse) which is why it avoids deadlocking when calling NtQueryObject.

> I've implemented the ideas from @dmex, but I still see deadlock. Although less often, 

Can you enable loader snaps and find the DLL responsible for the lock? 

1) Create a IFEO key for python/psutil. For example:
```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\processhacker.exe]
"GlobalFlag"=dword:00000002
```

The debugger will show the loader output. DebugView from Sysinternals can help if python doesn't:

image

Looking at you changes here: https://github.com/giampaolo/psutil/pull/2190/commits/3eba2d3c6304d3bf003f0e5406c5871996a98d7a

The psutil_get_filename thread is doing absolutely nothing except calling NtQueryObject just like we're doing... NtQueryObject simply does a syscall instruction and exits so there's nothing there that can load a DLL onto that thread and terminating it shouldn't be an issue. There has to be something else going on here to be causing the problem unrelated to psutil_get_filename?

xenu:
(I don't use psutil, I found this thread in Google)

We had a [similar problem](https://github.com/Perl/perl5/issues/19963) in Perl, we needed a way to tell if a `HANDLE` is a named pipe without deadlocking. We had no luck with almost all of `NtQueryInformationFile` and `NtQueryVolumeInformationFile` classes, but we found one that works: `FileVolumeNameInformation`. It's a new API, it was added in Win10 and it's undocumented.

https://github.com/Perl/perl5/blob/caef5a7dffddb63ccd397171f834d8f5122356ee/win32/win32.c#L5279

dmex:
There's also a third option using the built-in Windows rundown support?

Does not require a kernel driver, opening handles to processes or handle duplication so it's able to support protected processes that would otherwise require a driver. Rundown is included with all versions of Windows and can be used to enumerate handles, threads, images (dlls), processes, heap etc... 

image

Also supports stack traces (even stack traces from protected processes/objects/threads):

image

PolarGoose:
@dmex,
What is rundown?
Could you please provide more information and a reference code?

PolarGoose:
@giampaolo,

I'm an author of [Handle2](https://github.com/PolarGoose/Handle2). Thus, I had to research this topic extensively to find a way to avoid the deadlock.

### Quick answer
It is impossible to avoid deadlock. You have to use `TerminateThread` or a kernel driver like `PROCEXP152.SYS` (requires admin rights).

### Detailed answer

#### Why `NtQueryObject` deadlocks
Some handles are pipes. When you call `NtQueryObject` on them, it tries to acquire a lock of this handle and dealocks without any ability to cancel.
There is no way to know if a handle is a pipe or not unless you call `NtQueryObject` on it :) Thus, there is no way to avoid the deadlock.

#### Workarounds
* If you only want to enumerate file handles, then you can use `ReOpenFile`. It doesn't deadlock.
* If you want to enumerate all handles, there is no way to avoid `NtQueryObject` deadlocking. You have to use `TerminateThread`. Microsoft itself uses this approach in their [File Locksmith](https://learn.microsoft.com/en-us/windows/powertoys/file-locksmith) and [Sysinternals Handle](https://learn.microsoft.com/en-us/sysinternals/downloads/handle).
* Requires admin rights: use `PROCEXP152.SYS` driver. It is part of the [Sysinternals Process Explorer](https://learn.microsoft.com/en-us/sysinternals/downloads/process-explorer). It is why it doesn't hang. The driver doesn't have a public interface. But using pieces from various reverse engineering efforts, I created [ProcExp152SysDotNetProxy](https://github.com/PolarGoose/ProcExp152SysDotNetProxy) for C#. Also, [Backstab](https://github.com/Yaxser/Backstab) is a good reference for using this driver.

### Answering your questions

>  it seems we should use ReOpenFile() and fall back to the threaded NtQueryObject if ReOpenFile fails (named pipes, etc.)

It doesn't help. `ReOpenFile` only tells you whether this handle points to a file on disk. If it is a file on disk, then `NtQueryObject` will not hang anyway. Thus, there is no need to add an extra `ReOpenFile` check if you want to go through all handles.

> Has anyone tried combining those two lately on recent Windows versions (>= Win 10)? Or is the deadlock frequency still high enough that the timeout + TerminateThread is effectively mandatory?

Nothing has changed since `Windows XP`, at least. The frequency of hanging depends on how many pipes are used by the processes.

giampaolo:
Hi. This was finally fixed in https://github.com/giampaolo/psutil/pull/2894. The discussion in here was crucial, so thank you all.

PR #2197, opened 2023-01-26
Fix license: extra open quote.

snooppr:
The problem is that the open quote messes up the formatting of the entire document for other developers, which includes mentioning the author and his permission to freely use psutil.


snooppr:
cop
Here's an example where the quote breaks formatting in my license where psutil is mentioned.

PR #2198, opened 2023-02-04
support PyPy in setup.py

dand-oss:
pypy
fix TypeError: 'module' object is not subscriptable (key 'exec')
    __builtins__["exec"]("""def raise_from(value, from_value):

## Summary

* OS: { type-or-version }
* Bug fix: { yes/no }
* Type: { core, doc, performance, scripts, tests, wheels, new-api }
* Fixes: { comma-separated list of issues fixed by this PR, if any }

## Description

{{{
  A clear explanation of your bugfix or enhancement. Please read the contributing guidelines before submit:
  https://github.com/giampaolo/psutil/blob/master/CONTRIBUTING.md
}}}

### DIFF
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -369,7 +369,15 @@ class TimeoutExpired(Error):
 # This should be in _compat.py rather than here, but does not work well
 # with setup.py importing this module via a sys.path trick.
 if PY3:
-    __builtins__["exec"]("""def raise_from(value, from_value):
+    import platform
+    impl = platform.python_implementation()
+    if impl == "CPython":
+        f_exec = __builtins__["exec"]  # cpython dict
+    elif impl == "PyPy":
+        f_exec = getattr(__builtins__, "exec")  # pypy module
+    else:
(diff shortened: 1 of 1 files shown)


giampaolo:
I don't understand. These 2:

```python
        f_exec = __builtins__["exec"]  # cpython dict
        f_exec = getattr(__builtins__, "exec")  # pypy module
```

...should act exactly the same. 
What if we do this instead:

```
~/svn/psutil {master}$ git diff
diff --git a/psutil/_common.py b/psutil/_common.py
index da8a96e5..94fe9bb1 100644
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -369,7 +369,7 @@ class TimeoutExpired(Error):
 # This should be in _compat.py rather than here, but does not work well
 # with setup.py importing this module via a sys.path trick.
 if PY3:
-    __builtins__["exec"]("""def raise_from(value, from_value):
+    getattr(__builtins__, "exec")("""def raise_from(value, from_value):
     try:
         raise value from from_value
     finally:
```

Does it work on PyPy?

PR #2209, opened 2023-03-11
fix(windows): running tests in a virtual environment

mayeut:
## Summary

* OS: Windows
* Bug fix: yes
* Type: tests
* Fixes:
* close #2215
* close #2216

## Description

On windows, starting with python 3.7, virtual environments use a venvlauncher startup process This does not play well when counting spawned processes or when relying on the pid of the spawned process to do some checks e.g. connection check per pid

### DIFF
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -41,7 +41,6 @@ from socket import SOCK_STREAM
 
 import psutil
 from psutil import AIX
-from psutil import FREEBSD
 from psutil import LINUX
 from psutil import MACOS
 from psutil import POSIX
@@ -80,8 +79,8 @@ if POSIX:
 __all__ = [
     # constants
     'APPVEYOR', 'DEVNULL', 'GLOBAL_TIMEOUT', 'TOLERANCE_SYS_MEM', 'NO_RETRIES',
-    'PYPY', 'PYTHON_EXE', 'ROOT_DIR', 'SCRIPTS_DIR', 'TESTFN_PREFIX',
-    'UNICODE_SUFFIX', 'INVALID_UNICODE_SUFFIX',
+    'PYPY', 'PYTHON_BASE_EXE', 'PYTHON_EXE', 'ROOT_DIR', 'SCRIPTS_DIR',
+    'TESTFN_PREFIX', 'UNICODE_SUFFIX', 'INVALID_UNICODE_SUFFIX',
     'CI_TESTING', 'VALID_PROC_STATUSES', 'TOLERANCE_DISK_USAGE', 'IS_64BIT',
     "HAS_CPU_AFFINITY", "HAS_CPU_FREQ", "HAS_ENVIRON", "HAS_PROC_IO_COUNTERS",
     "HAS_IONICE", "HAS_MEMORY_MAPS", "HAS_PROC_CPU_NUM", "HAS_RLIMIT",
@@ -241,13 +240,8 @@ def _get_py_exe():
             return exe
 
     if GITHUB_ACTIONS:
-        if PYPY:
-            return which("pypy3") if PY3 else which("pypy")
-        elif FREEBSD:
-            return os.path.realpath(sys.executable)
-        else:
-            return which('python')
-    elif MACOS:
+        return sys.executable
+    if MACOS:
         exe = \
             attempt(sys.executable) or \
             attempt(os.path.realpath(sys.executable)) or \
@@ -262,7 +256,25 @@ def _get_py_exe():
         return exe
 
 
+# PYTHON_EXE always has psutil installed
 PYTHON_EXE = _get_py_exe()
+
+
+def _get_py_base_exe():
+    # On windows, starting with python 3.7,
+    # virtual environments use a venvlauncher startup process
+    # This does not play well when counting spawned processes or
+    # when relying on the pid of the spawned process to do some checks
+    # e.g. connection check per pid
+    # let's use the base python in this case.
+    base = getattr(sys, "_base_executable", None)
+    if WINDOWS and sys.version_info >= (3, 7) and base is not None:
+        return base
+    return PYTHON_EXE
+
+
+# PYTHON_BASE_EXE might not have psutil (or anything) installed
+PYTHON_BASE_EXE = _get_py_base_exe()
 DEVNULL = open(os.devnull, 'r+')
 atexit.register(DEVNULL.close)
 
@@ -365,7 +377,7 @@ def spawn_testproc(cmd=None, **kwds):
             pyline = "from time import sleep;" \
                      "open(r'%s', 'w').close();" \
                      "sleep(60);" % testfn
-            cmd = [PYTHON_EXE, "-c", pyline]
+            cmd = [PYTHON_BASE_EXE, "-c", pyline]
             sproc = subprocess.Popen(cmd, **kwds)
             _subprocesses_started.add(sproc)
             wait_for_file(testfn, delete=True, empty=True)
@@ -398,7 +410,7 @@ def spawn_children_pair():
             s += "time.sleep(60);"
             p = subprocess.Popen([r'%s', '-c', s])
             p.wait()
-            """ % (os.path.basename(testfn), PYTHON_EXE))
+            """ % (os.path.basename(testfn), PYTHON_BASE_EXE))
         # On Windows if we create a subprocess with CREATE_NO_WINDOW flag
         # set (which is the default) a "conhost.exe" extra process will be
         # spawned as a child. We don't want that.
@@ -471,10 +483,17 @@ def pyrun(src, **kwds):
     kwds.setdefault("stdout", None)
     kwds.setdefault("stderr", None)
     srcfile = get_testfn()
+    if PYTHON_BASE_EXE != PYTHON_EXE:
+        # We're in a Windows python 3.7+ virtual environment
+        # we need to set __PYVENV_LAUNCHER__ to sys.executable
+        # for the base python executable to know about the environment
+        env = kwds.get("env", os.environ).copy()
+        env["__PYVENV_LAUNCHER__"] = sys.executable
+        kwds["env"] = env
     try:
         with open(srcfile, 'wt') as f:
             f.write(src)
-        subp = spawn_testproc([PYTHON_EXE, f.name], **kwds)
+        subp = spawn_testproc([PYTHON_BASE_EXE, f.name], **kwds)
         wait_for_pid(subp.pid)
         return (subp, srcfile)
     except Exception:
@@ -851,7 +870,7 @@ def create_exe(outpath, c_code=None):
             safe_rmpath(f.name)
     else:
         # copy python executable
-        shutil.copyfile(PYTHON_EXE, outpath)
+        shutil.copyfile(PYTHON_BASE_EXE, outpath)
         if POSIX:
             st = os.stat(outpath)
             os.chmod(outpath, st.st_mode | stat.S_IEXEC)
@@ -1175,7 +1194,7 @@ def print_sysinfo():
     info['user'] = getpass.getuser()
     info['home'] = os.path.expanduser("~")
     info['cwd'] = os.getcwd()
-    info['pyexe'] = PYTHON_EXE
+    info['pyexe'] = PYTHON_BASE_EXE
     info['hostname'] = platform.node()
     info['PID'] = os.getpid()
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,21 @@ name: build
 jobs:
   # Linux + macOS + Windows Python 3
   py3:
-    name: py3-${{ matrix.os }}
+    name: py3-${{ matrix.os }}-${{ startsWith(matrix.os, 'windows') && matrix.archs || 'all' }}
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12, windows-2019]
-        # python: ["3.6", "3.11"]
+        include:
+        - os: ubuntu-latest
+          archs: "x86_64 i686"
+        - os: macos-12
+          archs: "x86_64 arm64"
+        - os: windows-2019
+          archs: "AMD64"
+        - os: windows-2019
+          archs: "x86"
 
     steps:
     - name: Cancel previous runs
@@ -44,6 +51,8 @@ jobs:
 
     - name: Create wheels + run tests
       uses: pypa/cibuildwheel@v2.11.2
+      env:
+        CIBW_ARCHS: "${{ matrix.archs }}"
 
     - name: Upload wheels
       uses: actions/upload-artifact@v3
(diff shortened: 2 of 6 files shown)


psutil/tests/__init__.py
```diff
@@ -260,6 +262,27 @@ def attempt(exe):
...
+        WINDOWS and \
+        sys.version_info >= (3, 7) and \
+        IS_VIRTUALENV and \
+        base is not None
+    if needs_base:
+        return base
+    return PYTHON_EXE
+
+
+PYTHON_BASE_EXE = _get_base_py_exe()
```
giampaolo, line 285: Unless there are circumstances in which it is necessary to use one instead of the other, I think it's simpler for the caller to simply think in terms of a single "PYTHON_EXE", without introducing a new public constant. Why not integrate this logic into `_get_py_exe()` instead? Something like (not tested):

```diff
+++ b/psutil/tests/__init__.py
@@ -196,6 +196,21 @@ SKIP_SYSCONS = (MACOS or AIX) and os.getuid() != 0
 # --- misc
 
 
+def _get_win_venv_py_exe():
+    # On windows, starting with python 3.7, virtual environments use a
+    # venvlauncher startup process This does not play well when counting
+    # spawned processes or when relying on the pid of the spawned
+    # process to do some checks e.g. connection check per pid let's use
+    # the base python in this case.
+    if (
+        WINDOWS and
+        sys.version_info >= (3, 7) and
+        IS_VIRTUALENV and
+        getattr(sys, "_base_executable", None)
+    ):
+        return sys._base_executable
+
+
 def _get_py_exe():
     def attempt(exe):
         try:
@@ -206,7 +221,9 @@ def _get_py_exe():
         else:
             return exe
 
-    if GITHUB_ACTIONS:
+    if WINDOWS and _get_win_venv_py_exe():
+        return _get_win_venv_py_exe()
+    elif GITHUB_ACTIONS:
         if PYPY:
             return which("pypy3") if PY3 else which("pypy")
         elif FREEBSD:
```

If possible, I would prefer a unified solution, mostly because I think there currently isn't a specific build for Windows that uses a virtualenv, so I'm afraid we may accidentally break tests again in the future.

psutil/tests/__init__.py
```diff
@@ -131,6 +131,8 @@
 COVERAGE = 'COVERAGE_RUN' in os.environ
 # are we a 64 bit process?
 IS_64BIT = sys.maxsize > 2 ** 32
+# PEP 405 virtual environment
+IS_VIRTUALENV = sys.prefix != getattr(sys, "base_prefix", sys.prefix)
```
giampaolo, line 135: I'm afraid this code won't work on Python 2.7 (`base_prefix` was added in [3.3](https://docs.python.org/3/library/sys.html#sys.base_prefix)).

A similar check is present in the Makefile, but we do something different (I'm not sure this works on 3.X though):
https://github.com/giampaolo/psutil/blob/2974941c179c1637b5102b0cebd3d332c2847678/Makefile#L40-L41

I'm surprised detecting a venv is so difficult. FWIW, I bumped into this:
https://stackoverflow.com/a/1883251
...which seems to work on both Python 2 and 3.

giampaolo (changes requested):

psutil/tests/__init__.py
mayeut: > Unless there are circumstances in which it is necessary to use one instead of the other, I think it's simpler for the caller to simply think in terms of a single "PYTHON_EXE", without introducing a new public constant

Agreed, however, using the base executable might require setting a specific environment variable to get things working properly (c.f. `__PYVENV_LAUNCHER__ ` in `pyrun`).
I'll check if I can get things unified.

> I think there currently isn't a specific build for Windows that uses a virtualenv, so I'm afraid we may accidentally break tests again in the future.

All windows tests are moved to cibuildwheel which always uses a virtual environment. This simplifies CI setup a bit and would allow to get a dev workflow using `nox` or `tox`, or just a virtual environment easily.

psutil/tests/__init__.py
mayeut: We only need to detect this properly on python 3.7+ on windows.
If python 2.7 does not respect PEP 405, we won't detect it's running in a virtual environment but the test outcome will be correct nonetheless.

psutil/tests/__init__.py
giampaolo: > All windows tests are moved to cibuildwheel which always uses a virtual environment.

I'm curious: was there a specific test failure that led to this PR?

giampaolo: Wow! That's a lot! Glad you're fixing it.

psutil/tests/__init__.py
mayeut: I removed the variable IS_VIRTUALENV to lessen confusion that this could trigger.

psutil/tests/__init__.py
mayeut: I reworked the PR to simplify all of this.
The only test where I expect things to fail is test_misc.TestScripts when ran from a virtual environment but those tests are just skipped in CI.
Either I try to get them running in CI here or leave it as-is for now ?

mayeut: I opened #2211 for `test_misc.TestScripts`, maybe good to get it merged as a pre-requisite for this PR ?

psutil/tests/__init__.py
```diff
@@ -228,35 +227,16 @@ def macos_version():
...
-            attempt(os.path.realpath(sys.executable)) or \
-            attempt(which("python%s.%s" % sys.version_info[:2])) or \
-            attempt(psutil.Process().exe())
-        if not exe:
-            raise ValueError("can't find python exe real abspath")
-        return exe
-    else:
-        exe = os.path.realpath(sys.executable)
-        assert os.path.exists(exe), exe
-        return exe
```
giampaolo, line 259: Mmm no, you're removing too much. All this part was gradually built over the years because relying on `sys.executable` alone was not enough. 

I suggest the approach I recommended here:
https://github.com/giampaolo/psutil/pull/2209#discussion_r1134446053
...meaning leaving the existing code as-is, and integrate the new logic on top of it.

psutil/tests/__init__.py
```diff
@@ -468,6 +448,10 @@ def pyrun(src, **kwds):
     kwds.setdefault("stdout", None)
     kwds.setdefault("stderr", None)
     srcfile = get_testfn()
+    if sys.executable != PYTHON_EXE:
+        env = kwds.get("env", os.environ).copy()
+        env["__PYVENV_LAUNCHER__"] = sys.executable
+        kwds["env"] = env
```
giampaolo, line 454: Please add a comment explaining why this is necessary. I am sure that I will look a this in X months and will be clueless. :)

psutil/tests/__init__.py
mayeut: > All this part was gradually built over the years 

At least the part mentioning GHA can be removed (it either passes or fails CI).

For the rest, it seems to be coming from mostly one commit https://github.com/giampaolo/psutil/commit/d108baf4a27f0f3b115aae5e6109eaf109302bc3, and without an issue or PR discussion around that, it seems mostly a preemptive action to something rather than fixing something, or maybe, it was fixing something at the time which might not be required anymore as the tests would suggest. Do you have any reference other than this commit ?

If keeping this, we might need to reintroduce 2 pythons for test_misc if you're not fine with the usage of sys.executable I'm about to push.

giampaolo: > For the rest, it seems to be coming from mostly one commit https://github.com/giampaolo/psutil/commit/d108baf4a27f0f3b115aae5e6109eaf109302bc3, and without an issue or PR discussion around that, it seems mostly a preemptive action to something rather than fixing something, or maybe, it was fixing something at the time which might not be required anymore as the tests would suggest. Do you have any reference other than this commit ?

I did that because on my virtualized OSX box the python exe was a symlink, and that created issues with tests using PYTHON_EXE (most likely they complained about missing path). I haven't turned on VirtualBox in a while, but I think the OSX workaround is still relevant today.

Note: I think you run tests via cibuildwheel and/or virtual env. I use the Makefile on all platforms (including windows via winmake.py), so some workarounds that you see in the code are because of my setup / way of doing things. 

> At least the part mentioning GHA can be removed (it either passes or fails CI).

If it doesn't break tests then sure, but does it? I can't remember why I added logic for PYPI and FREEBSD specifically, but most likely it was because some tests failed on GITHUB. 
The part about PYPY is not used because right now we don't have a PYPY build configured.
Anyway, I'm fine with removing the GITHUB_ACTIONS part as long as it doesn't break tests.

mayeut: Thanks for the background, I will reintroduce that.

> Note: I think you run tests via cibuildwheel and/or virtual env. I use the Makefile on all platforms (including windows via winmake.py), so some workarounds that you see in the code are because of my setup / way of doing things.

Indeed, I always use virtual environment, be it through cibuildwheel, nox, tox, manual setup, ...
I do really think that any development workflow shall not interfere with the system/user installation but that's just my opinion.

giampaolo: > I do really think that any development workflow shall not interfere with the system/user installation but that's just my opinion.

I certainly agree. It's an old bad habit I have. FWIW, also the Makefile can be used with venv: 
`make install test PYTHON=/path/to/your/venv`

mayeut: > Anyway, I'm fine with removing the GITHUB_ACTIONS part as long as it doesn't break tests.

I kept it but simplified it.

It seems running linux tests from a virtual environment is failing when just using `exe = os.path.realpath(sys.executable)`. Let's defer this to another PR if needed.

The symptom is `ModuleNotFoundError: No module named 'psutil'`
Using `os.path.realpath(sys.executable)` resolves python outside the virtual env which explains the symptom (also tested with python 3.6).
```
[root@b0438d6ef25a ~]# python3.11 -c 'import sys; import os; print(sys.executable); print(os.path.realpath(sys.executable))'
/usr/local/bin/python3.11
/opt/_internal/cpython-3.11.2/bin/python3.11
[root@b0438d6ef25a ~]# venv-311/bin/python -c 'import sys; import os; print(sys.executable); print(os.path.realpath(sys.executable))'
/root/venv-311/bin/python
/opt/_internal/cpython-3.11.2/bin/python3.11
[root@b0438d6ef25a ~]# virtualenv-311/bin/python -c 'import sys; import os; print(sys.executable); print(os.path.realpath(sys.executable))'
/root/virtualenv-311/bin/python
/opt/_internal/cpython-3.11.2/bin/python3.11
```

giampaolo: Wait. You introduced the 2 PYTHONs again (PYTHON_EXE and PYTHON_BASE_EXE). I don't understand why you can't get away with just using one. I'd like to see the problem you have specifically and try debugging it. Can you tell me how I can I reproduce the problem you're experiencing (I'm on linux but I can fire up VMs with other OSes)?

giampaolo: OK, I was able to reproduce the errors and fix them with:

```diff
diff --git a/psutil/tests/__init__.py b/psutil/tests/__init__.py
index 61a06e21..28a0c941 100644
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -240,7 +240,10 @@ def _get_py_exe():
         else:
             return exe

-    if GITHUB_ACTIONS:
+    base_executable = getattr(sys, "_base_executable", None)
+    if WINDOWS and sys.version_info >= (3, 7) and base_executable is not None:
+        return base_executable
+    elif GITHUB_ACTIONS:
         if PYPY:
             return which("pypy3") if PY3 else which("pypy")
         elif FREEBSD:
```

The only one left I have is this, but I'm not sure if it's related with the issue at hand:

```
FAIL: psutil.tests.test_process.TestProcess.test_nice (prio=256)
  File "C:\Users\user\Desktop\psutil\psutil\tests\test_process.py", line 825, in test_nice
    self.assertEqual(new_prio, prio)
AssertionError: <Priority.HIGH_PRIORITY_CLASS: 128> != 256
```

What errors do you get if you apply the same patch?

giampaolo: And this is how I run tests btw:

```
C:\Users\user\Desktop\psutil>set PYTHON=venv\Scripts\python.exe & make test
```

mayeut: let me try to setup something similar on GHA and share the results on a branch in my fork.

mayeut: This patch is failing (failing to find psutil). It's what I'd expect. https://github.com/mayeut/psutil/actions/runs/4461399430/jobs/7835201234

I'm still trying to adapt the "set PYTHON=venv\Scripts\python.exe & make test" workflow to GHA as I don't have access to a Windows VM otherwise.

mayeut: I tried to reproduce the dev workflow in https://github.com/mayeut/psutil/actions/runs/4461656326 which is also failing with venv.

giampaolo: Mmm failed to `"import psutil"` is a different problem than the previous tests that we saw failing, meaning the ones you pasted here (and that were also my failures):
https://github.com/giampaolo/psutil/pull/2209#discussion_r1134563313

It should mean that psutil was installed with one python executable path, and `PYTHON_EXE` var is pointing to another python executable path. At the begging of the build output we can see the value of `PYTHON_EXE`:

```
2023-03-19T16:36:12.5353340Z pyexe:            C:\hostedtoolcache\windows\Python\3.11.2\x64\python.exe
```

That line is being printed by this:
https://github.com/giampaolo/psutil/blob/fb0d5d3325efeb56dd0cf48c0f0d2379baca20d5/psutil/tests/__init__.py#L1178

I suppose one thing you can do is adding more debugging "prints" in there to try to understand the differences between the different paths. E.g.:

```
    info['PYTHON_EXE'] = PYTHON_EXE
    info['sys.executable'] = sys.executable
    info['sys."_base_executable'] =  getattr(sys, "_base_executable", None)
```

Considering in the past I was bitten by symlinks, maybe another thing you can try is `os.path.realpath(sys._base_executable)`.

giampaolo: BTW, I added `PYTHON_EXE` / `_get_py_exe()` precisely for this sort of problems. That's what I meant when I said relying on `sys.executable` was not enough. Workarounds are needed depending on the platform / setup / environment.

mayeut: The patch in this PR also works well in the reproduced dev workflow: https://github.com/mayeut/psutil/actions/runs/4465692219/jobs/7843003771

As mentioned in comments & in the code, PYTHON_BASE_EXE / sys._base_executable / or even os.path.realpath(sys._base_executable) won't have psutil installed, by design (it's the global python interpreter, not the venv) & that's why we need 2. In this PR, only `pyrun` reinjects the `__PYVENV_LAUNCHER__` to get psutil importable when using PYTHON_BASE_EXE.
If we don't use the global python interpreter, we can't count processes (or we'd need a specific workaround) or rely on the pid of the launched interpreter to count connections. If we use the global interpreter, we can't import psutil.
My conclusion, getting 2 is the easiest, otherwise, we'd need 1 + an environment variable or rework the failing tests but I feel the 2 interpreters is easier. We can't get away from those issues just by modifying PYTHON_EXE.

giampaolo: > My conclusion, getting 2 is the easiest, otherwise, we'd need 1 + an environment variable.

This seems more viable to me. My point is the user of these constants should not think "which one should I use?".  It's an implementation detail that is difficult to understand (at least I still don't fully understand), and which only affects Windows + venv running GITHUB_CI.

I forgot how your previous changes of 1 PYTHON_EXE + env var looked like, but I would try to rehash that approach instead and see how we can arrange it.

giampaolo: > I forgot how your previous changes of 1 PYTHON_EXE + env var looked like, but I would try to rehash that approach instead and see how we can arrange it.

Maybe you can do it in a separate PR so that we stop switching between approaches in the same PR.

mayeut: > I forgot how your previous changes of 1 PYTHON_EXE + env var looked like, but I would try to rehash that approach instead and see how we can arrange it.

I've never pushed this one (or even did it in my fork). I will try to open a new PR with this approach this week.

mayeut: I opened 2 other PR with alternative implementations.
The 3 PRs pass the "dev like workflow" on my fork (side-note, it seems the just released pyperf 2.6.0 breaks that workflow, at least the way I implemented it in GHA, so I pinned that to 2.5.0 in my tests of the dev workflow)

PR #2215, opened 2023-03-25
fix(windows): running tests in a virtual environment (alt 1)

mayeut:
## Summary

* OS: Windows
* Bug fix: yes
* Type: tests
* Fixes:
* close #2209
* close #2216

## Description

On windows, starting with python 3.7,
virtual environments use a venvlauncher startup process This does not play well when counting spawned processes or when relying on the pid of the spawned process to do some checks e.g. connection check per pid

1st alternative to #2209, substitutes PYTHON_EXE with the correct one in `psutil.tests.spawn_testproc`/`psutil.tests. spawn_children_pair`
No comments added yet.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,21 @@ name: build
 jobs:
   # Linux + macOS + Windows Python 3
   py3:
-    name: py3-${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


mayeut:
The windows x86 build seems to be a random error, shall `@retry_on_failure()` be added to that test ?

PR #2216, opened 2023-03-25
fix(windows): running tests in a virtual environment (alt 2)

mayeut:
## Summary

* OS: Windows
* Bug fix: yes
* Type: tests
* Fixes:
* close #2209
* close #2215

## Description

On windows, starting with python 3.7,
virtual environments use a venvlauncher startup process This does not play well when counting spawned processes or when relying on the pid of the spawned process to do some checks e.g. connection check per pid

2nd alternative to #2209, exposes PYTHON_EXE_ENV to be set as the base environment when starting a process using PYTHON_EXE.

The 1st commit always uses `env= PYTHON_EXE_ENV`.
The 2nd commit removes occurrences when calling `psutil.tests.spawn_testproc` (either directly or indirectly) which uses this as its default environment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,21 @@ name: build
 jobs:
   # Linux + macOS + Windows Python 3
   py3:
-    name: py3-${{ matrix.os }}
+    name: py3-${{ matrix.os }}-${{ startsWith(matrix.os, 'windows') && matrix.archs || 'all' }}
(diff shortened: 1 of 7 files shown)


giampaolo (changes requested):
I think this is the cleanest approach. I just added a couple of minor comments.

giampaolo:
Merged. Thank you Matthew.

PR #2220, opened 2023-04-02
[Linux] More tolerant battery detection

### LINKED ISSUE #2219: [Linux] battery detection in `psutil.sensors_battery()` relies on an assumption which is not true of all systems.
## Summary

* OS: Linux debian 5.10.0-21-686-pae
* Architecture: i686
* Psutil version: 5.8.0
* Python version: 3.9.2
* Type: core? (not sure)

## Description

Running `psutil.sensors_battery()` on a HP 520 laptop returns `None`. The `/sys/class/power_supply` directory contains `C1AB  C1AC` but no directory that begins in BAT, out of these I think the C1AC is the right directory.

sujaldev:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: #2219

## Description

Implements a more tolerant method for detecting battery.

- [x] Updated HISTORY.rst
- [x] Updated CREDITS

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1548,17 +1548,45 @@ def sensors_battery():
                     return ret.strip()
         return None
 
+    def detect_battery_path(*expected_files):
+        """Find the device path having the maximum number of files
+        that are to be excepted for a battery. If no path has the
+        expected files, return None. Used as a fallback for when
+        the usual prefixes don't find a match.
+        """
+        highest_score, final_battery = 0, None
+        for battery in os.listdir(POWER_SUPPLY_PATH):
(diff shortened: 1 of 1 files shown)


sujaldev:
If this wasn't being merged due to the merge conflicts, I have resolved those.

Also I've rewritten everything such that my proposed method of detecting the battery device path (which is to look at the contents of the directory rather than it's name) only comes into play if the current way of doing things wasn't able to detect a battery.

sujaldev:
@giampaolo don't know if you noticed the changes, bump.

sujaldev:
I'm deleting the changes to `CREDITS` and `HISTORY.rst` because they keep accumulating merge conflicts and I sense this is not being merged any time soon (assuming there even are plans to merge this).

Issue #2189, opened 2023-01-05
Cannot get cmdline attribute for terminal process


tamruta:
Recreated the problem I was facing:

``` python
process_hash = "login"  
for proc in psutil.process_iter():
    if process_hash in proc.name():
        server_process_running = True
        print(proc)
        print(str(proc.as_dict(['cmdline'])))  # this is where the problem rises 
```

``` console
psutil.Process(pid=3629, name='login', status='running', started='09:16:35')
Traceback (most recent call last):
  File "/opt/anaconda3/envs/trust/lib/python3.10/site-packages/psutil/_psosx.py", line 346, in wrapper
    return fun(self, *args, **kwargs)
  File "/opt/anaconda3/envs/trust/lib/python3.10/site-packages/psutil/_psosx.py", line 404, in cmdline
    return cext.proc_cmdline(self.pid)
ProcessLookupError: [Errno 3] assume no such process (originated from sysctl(KERN_PROCARGS2) -> EINVAL)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/test.py", line 14, in <module>
    print(str(proc.as_dict(['cmdline'])))
  File "/opt/anaconda3/envs/trust/lib/python3.10/site-packages/psutil/__init__.py", line 528, in as_dict
    ret = meth()
  File "/opt/anaconda3/envs/trust/lib/python3.10/site-packages/psutil/__init__.py", line 681, in cmdline
    return self._proc.cmdline()
  File "/opt/anaconda3/envs/trust/lib/python3.10/site-packages/psutil/_psosx.py", line 351, in wrapper
    raise NoSuchProcess(self.pid, self._name)
psutil.NoSuchProcess: process no longer exists (pid=3629, name='login')
```

When I looked online about similar errors, some suggested it was caused by a zombie process. So I tried to kill the process and the terminal session closed. 

I'm not sure if this is a bug or a feature, but this only happened when I updated psutil from 5.6.7 to 5.9.4

giampaolo:
That's a race condition: between the `for` loop iteration and the `proc.as_dict()` call the process disappears. You have to ways to solve this problem. You either specify what process attributes you want to access up front:

```python
for proc in psutil.process_iter(attrs=["name", "cmdline"]):
      print(proc.info["name"])
      print(proc.info["cmdline"])
```

...or you catch the exception later:

```python
for proc in psutil.process_iter():
    try:
        print(proc.name())
        print(proc.cmdline())
[...]

PR #2222, opened 2023-04-08
Fix cpu_freq for Apple silicon

### LINKED ISSUE #1892: [MacOS] psutil.cpu_freq() broken on Apple M1

snOm3ad:
## Summary

* OS: macOS Ventura 13.3
* Bug fix: yes
* Type: core
* Fixes: #1892 

## Description

Apple SoC returns empty string when querying the cpu frequency using `sysctl`, however, the `min` and `max` frequency values can still be extracted from the IOKit registry. This PR adds a new implementation of `cpu_freq` for arm64 architecture that fetches these values. It does *not* try to retrieve the current frequency as this metric is not available in IOKit registry and can only be found using a more involved process.


psutil/arch/osx/cpu.c
```diff
@@ -109,6 +112,48 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+        if (strncmp(name, "pmgr", 4) == 0) {
+            break;
+        }
+    }
+    IOObjectRelease(iter);
+    CFRelease(matching);
+    CFTypeRef pCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states5-sram"), kCFAllocatorDefault, 0);
+
+    size_t length = CFDataGetLength(pCoreRef);
+    for (size_t i = 0; i < length - 3; i += 4) {
```
dbwiddis, line 139: While this iteration works for this particular use case, it's inaccurate.  The array you are iterating over contains pairs of (4-byte) integers; the first one is the frequency you're looking for (little-endian).   For example, here's the array you're parsing from an [online M1 dump](https://github.com/xzhih/one-key-hidpi/files/5612845/ioreg_output_m1mac.txt), split into the pairs of numbers:
```
0046c323 0d030000
00475a31 0d030000
0048f13e 0d030000
0049884c 20030000
002f6859 2c030000
0030ff66 3f030000
00319674 61030000
00322d82 8d030000
00fd558e b9030000
00c87e9a eb030000
005d39a5 1d040000
004319b2 39040000
003689b8 39040000
00a265bb 39040000
0029f9be 39040000
```
The min value you want is 0x23c34600, the first integer in the array.
The max value you want is 0xbef92900, the second-to-last integer in the array.

Here's a screenshot of my M2 Max E-core values:
Screenshot 2023-04-08 at 11 08 05 AM

My min E-core freq is 0x365c0400 (912 MHz).
My max E-core freq is 0x907b4e00 (2.424 GHz).

And the P-core values:
Screenshot 2023-04-08 at 11 16 48 AM

My min P-core freq is 0x29d7ab80 (702 MHz).
My max P-core freq is 0xdc4c7c00 (3.696 GHz).

So you could (probably) grab from index 0 and index (length - 8) and avoid iteration.  However, I'm only guessing that the array is always sorted in increasing order, as I haven't found that documented anywhere but have confirmed it with every sample output I could get my hands on.  If you do iterate, I would increment by 8 rather than 4.  That will save you the need to do your `curr_freq > 1e6` test below.

dbwiddis (commented):
Some leaking references you need to handle, and an iteration suggestion.

giampaolo (changes requested):
Instead of defining 2 separate C functions and catching `FileNotFoundError` in python I suggest to split the logic in 2 separate functions, and use the C pre-processor to check whether you're on ARM. Something like (pseudo code):

```c
#ifdef SOMETHING_DEFINING_ARM
PyObject *
psutil_cpu_freq(PyObject *self, PyObject *args) {
    ...
}
#else
PyObject *
psutil_cpu_freq(PyObject *self, PyObject *args) {
    ...
}
#endif
```

psutil/arch/osx/cpu.c
giampaolo: Hey Daniel

> Compare to powermetrics output with my CPUs idling (mostly) at min:

It's interesting to note that `powermetrics` output shows it's able to return per-CPU frequencies. Current psutil's [cpu_freq() implementation](https://github.com/giampaolo/psutil/blob/7eadee31db2f038763a3a6f978db1ea76bbc4674/psutil/arch/osx/cpu.c#L113-L140) on OSX doesn't do that (same on Windows and OpenBSD).

dbwiddis: > It's interesting to note that `powermetrics` output shows it's able to return per-CPU frequencies. Current psutil's [cpu_freq() implementation](https://github.com/giampaolo/psutil/blob/7eadee31db2f038763a3a6f978db1ea76bbc4674/psutil/arch/osx/cpu.c#L113-L140) on OSX doesn't do that (same on Windows and OpenBSD).

Yeah, I've been chasing the "current frequency" nut for years.  I've always thought that CPU frequency was something like Heisenberg's uncertainty principle.  You change it by measuring it. :)

`powermetrics` is useful but is also closed-source so I don't know how it works, but it isn't really suitable for command-line parsing:
 - it polls every 5 seconds 
 - it requires sudo
 - it has differing output per-core.

That said, I will show you how powermetrics gets its "current frequency" and try to amplify your comment on your review,
> I am not sure I understand what is going on here. It seems it's based too much on guessing.

I'd prefer to use the term "reverse engineering". :) There are plenty of ARM CPU IORegistry dumps out there to validate these particular calls from iPhones, and I've personally tested them on both M1 and M2.

The pairs of numbers are most definitely frequency values.  I just wrote a quick program to print out the pairs on my M2:
```
voltage-states5-sram
702000000 / 790
948000000 / 790
1188000000 / 790
1452000000 / 806
1704000000 / 834
1968000000 / 859
2208000000 / 896
2400000000 / 915
2568000000 / 931
2724000000 / 978
2868000000 / 1009
3000000000 / 1062
3132000000 / 1087
3264000000 / 1140
3360000000 / 1140
3408000000 / 1140
3504000000 / 1140
3528000000 / 1196
3696000000 / 1196

voltage-states1-sram
912000000 / 590
1284000000 / 625
1752000000 / 678
2004000000 / 706
2256000000 / 778
2424000000 / 831
```
As I mentioned before I'm not sure what the second number is, but the first one is definitely frequency, as the more detailed `powermetrics` output shows. It gives percentage of time (probably over that 5-second polling interval) that the CPU was operating at each of its possible speeds.

CPU 0 (an E-core) spent 100% of its time in the lowest 912 MHz setting; but you can see the other values exactly line up with the voltage states IOReg value.
```
CPU 0 frequency: 912 MHz
CPU 0 idle residency:   0.00%
CPU 0 active residency: 100.00% (912 MHz: 100% 1284 MHz: .00% 1752 MHz:   0% 2004 MHz:   0% 2256 MHz:   0% 2424 MHz:   0%)
```
CPU 4 (a P-core) is a bit more insteresting, the 751 MHz "current frequency" is a weighted average of the possible frequencies by the amount of time spent in them:
```
CPU 4 frequency: 751 MHz
CPU 4 idle residency:  15.76%
CPU 4 active residency:  84.24% (702 MHz:  68% 948 MHz:  16% 1188 MHz:   0% 1452 MHz: .03% 1704 MHz: .00% 1968 MHz: .13% 2208 MHz: .03% 2400 MHz: .01% 2568 MHz:   0% 2724 MHz:   0% 2868 MHz:   0% 3000 MHz:   0% 3132 MHz:   0% 3264 MHz:   0% 3360 MHz: .00% 3408 MHz:   0% 3504 MHz:   0% 3528 MHz:   0% 3696 MHz:   0%)
```
Note that 751 MHz is only an average for the 84.24% of time that it was active.  What was its frequency when idle?  Zero?  So perhaps the average should really be 629 MHz?   

TLDR: "Current Frequency" on any CPU is too poorly defined to really measure.  However, these outputs show these voltage states as a reliable source for the possible settings.

Is it documented these outputs are always in ascending order?  No, I'm guessing. But it would make no sense for outputs like the above to have to sort each time, so it's a reasonable guess that every other 4 bytes is a possible frequency, so index 0 is the min frequency and index (length-8) is the max frequency for that core.

dbwiddis: This is an interesting search: https://github.com/search?q=voltage-states5-sram&type=code

It includes:
 - My project (oshi) and a project BUS which seems to copy my (MIT) code :) 
 - The go project @shoenig mentioned in the issue
 - qemu which hardcodes those voltage states arrays when emulating a t8030 chip (A13)
 - an `osquery` project with [code that looks a lot like this PR](https://github.com/cdprf/osquery/blob/aea0d6ef30a5e38f66cca252ebfcebe80d1f231a/osquery/tables/system/darwin/cpu_info.cpp#L145-L155)

dbwiddis: My current working theory is the second number in each pair is the voltage (in millivolts) that is set when the corresponding frequency is in use (e.g., 1.2V at highest frequencies).  Would make sense given the name :)

psutil/arch/osx/cpu.c
```diff
@@ -109,6 +112,48 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+        uint32_t curr_freq = 0;
+        CFDataGetBytes(pCoreRef, CFRangeMake(i, sizeof(uint32_t)), (UInt8 *) &curr_freq);
+        if (curr_freq > 1e6 && curr_freq < min) {
+            min = curr_freq;
+        }
+        if (curr_freq > max) {
+            max = curr_freq;
+        }
+    }
+    curr = max;
```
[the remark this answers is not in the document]
snOm3ad: I will refactor this to match what Daniel mentioned above. It's reasonable to assume that the array is always sorted. Though this _not_ documented anywhere, the evidence so far seems to suggest this is the case. This will simplify this logic to something like:

```c
CFDataGetBytes(pCoreRef, CFRangeMake(0, sizeof(uint32_t)), (UInt8 *) &min_freq);
CFDataGetBytes(pCoreRef, CFRangeMake(length - sizeof(uint32_t), sizeof(uint32_t)), (UInt8 *) &max_freq);
```

This way we no longer have to loop through the entire byte array.

psutil/arch/osx/cpu.c
snOm3ad: > Also I notice you're only iterating the P-core values and not the E-core ones. Is this intentional?

I was under the impression that E-core values have a higher minimum frequency and lower max frequency than P-core ones. However, I think only the latter statement is true. So to find the minimum frequency you would also need to query the E-core values as well and compare them against the others.

snOm3ad:
> Instead of defining 2 separate C functions and catching `FileNotFoundError` in python I suggest to split the logic in 2 separate functions, and use the C pre-processor to check whether you're on ARM.

Is there a flag that is currently being used? Or would I need to create one to achieve this. I am assuming this would need to be incorporated in the `setup.py` script?

dbwiddis:
> Is there a flag that is currently being used? Or would I need to create one to achieve this.

I'm pretty sure you can do an ifdef on `__aarch64__`.  See https://opensource.apple.com/source/WTF/WTF-7601.1.46.42/wtf/Platform.h.auto.html
```c
/* CPU(X86_64) - AMD64 / Intel64 / x86_64 64-bit */
#if   defined(__x86_64__) \
    || defined(_M_X64)
#define WTF_CPU_X86_64 1
#endif

/* CPU(ARM64) - Apple */
#if (defined(__arm64__) && defined(__APPLE__)) || defined(__aarch64__)
#define WTF_CPU_ARM64 1
#endif
```

snOm3ad:
Thanks Daniel! The only pending question would be what to do about the current frequency? At the moment we don't know if (let alone where) this information is accessible from IOReg. As you mentioned above `powermetrics` is closed source this makes it especially hard to know where the numbers are coming from.

psutil/arch/osx/cpu.c
dbwiddis: > I was under the impression that E-core values have a higher minimum frequency and lower max frequency than P-core ones. 

That's true now, but I'm not sure it will always be true.

dbwiddis:
> Thanks Daniel! The only pending question would be what to do about the current frequency? At the moment we don't know if (let alone where) this information is accessible from IOReg. 

Actually, we do, or at least @BitesPotatoBacks does, see https://github.com/BitesPotatoBacks/SocPowerBuddy

It doesn't yet work on M2 and thus is probably brittle.  What about M3, etc.?  

Anyway, the powermetrics "current" value is just a "current average over the last 5 seconds" which isn't really current, is it?   

I'm of the opinion that we really shouldn't care about current frequency, based on the logic in this answer: https://apple.stackexchange.com/a/329537/388596

snOm3ad:
>  see https://github.com/BitesPotatoBacks/SocPowerBuddy

The sampler code is relying on `IOReport` functions. This technique is explained on these [slides](https://www.slideshare.net/kstan2/exploring-your-apple-m1-devices-with-open-source-tools) by @freedomtan. The [ioreport](https://github.com/freedomtan/test-ioreport) repository seems particularly interesting.

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    while ((entry = IOIteratorNext(iter))) {
+        io_name_t name;
+        IORegistryEntryGetName(entry, name);
+
+        if (strncmp(name, "pmgr", 4) == 0) {
+            break;
+        }
+    }
+
+    CFTypeRef pCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states5-sram"), kCFAllocatorDefault, 0);
```
dbwiddis, line 139: In the case you don't find "pmgr" in the iteration, `entry` will be null here.

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    CFDictionaryRef matching = IOServiceMatching("AppleARMIODevice");
+    io_iterator_t iter;
+
+    IOServiceGetMatchingServices(kIOMainPortDefault, matching, &iter);
+    io_registry_entry_t entry;
+    while ((entry = IOIteratorNext(iter))) {
+        io_name_t name;
+        IORegistryEntryGetName(entry, name);
+
+        if (strncmp(name, "pmgr", 4) == 0) {
```
dbwiddis, line 134: This works, but why not use `strcmp()` instead?

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    CFTypeRef eCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states1-sram"), kCFAllocatorDefault, 0);
+
+    IOObjectRelease(iter);
+    IOObjectRelease(entry);
 
+    size_t pCoreLength = CFDataGetLength(pCoreRef);
+
+    CFDataGetBytes(pCoreRef, CFRangeMake(0, 4), (UInt8 *) &pMin);
+    CFDataGetBytes(eCoreRef, CFRangeMake(0, 4), (UInt8 *) &eMin);
+    CFDataGetBytes(pCoreRef, CFRangeMake(pCoreLength - 8, 4), (UInt8 *) &max);
```
dbwiddis, line 149: You should probably do a check to ensure `pCoreLength >= 8` before this line.

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+
+    min = pMin < eMin ? pMin : eMin;
+
+    CFRelease(pCoreRef);
+    CFRelease(eCoreRef);
+
+    return Py_BuildValue(
+            "sKK",
+            NULL,
```
dbwiddis, line 158: In the x86 world, "current" just returns "max".  That may be preferable here.

Also I'm not a python coder but I would think you'd want the return types identical to the x86 option, e.g., IKK.

dbwiddis (changes requested):

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
     );
 }
 
+#if (defined(__arm64__) && defined(__APPLE__)) || defined(__aarch64__)
```
giampaolo, line 117: defined(__APPLE__) seems redundant. This file is only loaded on macOS so it shouldn't be needed.

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,52 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+psutil_cpu_freq(PyObject *self, PyObject *args) {
+    uint32_t min;
+    uint32_t pMin;
+    uint32_t eMin;
+    uint32_t max;
+
+    CFDictionaryRef matching = IOServiceMatching("AppleARMIODevice");
+    io_iterator_t iter;
+
+    IOServiceGetMatchingServices(kIOMainPortDefault, matching, &iter);
```
giampaolo, line 128: You should always check the return value of each call. E.g. in here:
```c
kern_return_t   kern_ret;

kern_ret = IOServiceGetMatchingServices(kIOMainPortDefault, matching, &iter);
if (kern_ret != KERN_SUCCESS) {
    PyErr_SetString(PyExc_RuntimeError, "IOServiceGetMatchingServices failed");
    return NULL;
}

...but also subsequent calls should have similar checks.

**Review by giampaolo — CHANGES_REQUESTED** [review]

**Comment by freedomtan** [comment]
> > see https://github.com/BitesPotatoBacks/SocPowerBuddy
> 
> The sampler code is relying on `IOReport` functions. This technique is explained on these [slides](https://www.slideshare.net/kstan2/exploring-your-apple-m1-devices-with-open-source-tools) by @freedomtan. The [ioreport](https://github.com/freedomtan/test-ioreport) repository seems particularly interesting.

As far as I can tell, those frequency information are not "real" ones. They are data Apple used for power/energy modeling (which in turn is used for scheduling and others)

File: psutil/arch/osx/cpu.c [hunk 9]
```diff
@@ -109,7 +114,78 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    io_registry_entry_t entry;
+    while ((entry = IOIteratorNext(iter))) {
+        io_name_t name;
+        status = IORegistryEntryGetName(entry, name);
+        if (status != KERN_SUCCESS) {
+            continue;
+        }
+        if (strcmp(name, "pmgr") == 0) {
+            break;
+        }
```
dbwiddis, line 147: You still need to release entry here during the iteration before you advance to the next (in the case where you don't break).

dbwiddis (changes requested):

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,78 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    }
+
+    CFTypeRef pCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states5-sram"), kCFAllocatorDefault, 0);
+    if (!pCoreRef) {
+        return PyErr_Format(PyExc_RuntimeError, "'voltage-states5-sram' property not found");
+    }
+
+    CFTypeRef eCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states1-sram"), kCFAllocatorDefault, 0);
+    if (!eCoreRef) {
+        return PyErr_Format(PyExc_RuntimeError, "'voltage-states1-sram' property not found");
```
giampaolo, line 161: You need to release `pCoreRef` before returning. More in general, I suggest to centralize the cleanup logic in one place, and use a goto.
This construct is used pretty much everywhere in psutil, so you can search for "error:" in the C files to know what I mean, but long story short what I'm suggesting is something like this (not tested).
Every time there is an error, anywhere, just set exception and then do `goto error;`.

```c
PyObject *
psutil_cpu_freq(PyObject *self, PyObject *args) {
    CFTypeRef pCoreRef = NULL;
    CFTypeRef eCoreRef = NULL;
    io_iterator_t iter = NULL;
    io_registry_entry_t entry = NULL;

    ...

    eCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states1-sram"), kCFAllocatorDefault, 0);
    if (!eCoreRef) {
        PyErr_Format(PyExc_RuntimeError, "'voltage-states1-sram' property not found");
        goto error;
    }

    ...

    error:
        if (pCoreRef != NULL):
            CFRelease(pCoreRef);
        if (eCoreRef != NULL):
            CFRelease(eCoreRef);
        if (iter != NULL)
            IOObjectRelease(iter);
        if (entry != NULL)
            IOObjectRelease(entry);
        return NULL;
}
```

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,94 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+        PyErr_Format(PyExc_RuntimeError, "'voltage-states5-sram' property not found");
+        goto error;
+    }
+
+    eCoreRef = IORegistryEntryCreateCFProperty(entry, CFSTR("voltage-states1-sram"), kCFAllocatorDefault, 0);
+    if (!eCoreRef) {
+        PyErr_Format(PyExc_RuntimeError, "'voltage-states1-sram' property not found");
+        goto error;
+    }
+
```
dbwiddis, line 173: You've released all the `entry` instances in the loop and in error messages, but you need one more here. :-)

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,94 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    size_t pCoreLength = CFDataGetLength(pCoreRef);
+    if (pCoreLength < 8) {
```
dbwiddis, line 175: I think you also need a check for `eCoreRef` having at least 4 bytes (line 181)

dbwiddis (approved):
Almost there!

snOm3ad:
Thanks for the help Daniel! Still have to wait for Giampaolo's comments, but I think it should be good now.

psutil/arch/osx/cpu.c
```diff
@@ -109,6 +112,48 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    unsigned int curr;
+    uint32_t min = UINT32_MAX;
+    uint32_t max = 0;
+
+    CFDictionaryRef matching = IOServiceMatching("AppleARMIODevice");
+    io_iterator_t iter;
+
+    IOServiceGetMatchingServices(kIOMainPortDefault, matching, &iter);
+    io_registry_entry_t entry;
+    while ((entry = IOIteratorNext(iter))) {
```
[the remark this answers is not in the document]
dbwiddis: I'm not sure if these are redundant parens or not.

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,107 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    }
 
+    CFDataGetBytes(pCoreRef, CFRangeMake(0, 4), (UInt8 *) &pMin);
+    CFDataGetBytes(eCoreRef, CFRangeMake(0, 4), (UInt8 *) &eMin);
+    CFDataGetBytes(pCoreRef, CFRangeMake(pCoreLength - 8, 4), (UInt8 *) &max);
+
+    min = pMin < eMin ? pMin : eMin;
+    curr = max;
+
+    if (pCoreRef != NULL)
```
dbwiddis, line 192: I realized you copy-pasted from the error condition, but I'm pretty sure we know this isn't null here (line 163).

Similarly for `eCoreRef` tested on line 160 and `entry` line 157.   And I don't think it'll hurt to call a release on a 0 `iter` (since you already iterated on it).

dbwiddis (approved):
LGTM with a few more improvements noted. :)

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,103 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+        return PyErr_Format(PyExc_RuntimeError, "IOServiceMatching call failed, 'AppleARMIODevice' not found");
+    }
+
+    status = IOServiceGetMatchingServices(kIOMainPortDefault, matching, &iter);
+    if (status != KERN_SUCCESS) {
+        PyErr_Format(PyExc_RuntimeError, "IOServiceGetMatchingServices call failed");
+        goto error;
+    }
+
+    while ((entry = IOIteratorNext(iter))) {
```
giampaolo, line 143: Isn't this line supposed to be like this (assignment + conditional)?

```c
while ((entry = IOIteratorNext(iter) != 0)) {
     ...
}
```

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,103 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+            min / 1000 / 1000,
+            max / 1000 / 1000);
+error:
+    if (pCoreRef != NULL)
+        CFRelease(pCoreRef);
+    if (eCoreRef != NULL)
+        CFRelease(eCoreRef);
+    if (iter)
+        IOObjectRelease(iter);
+    if (entry)
```
giampaolo, line 209: Should this be `if (entry != 0)`?

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,103 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+            "IKK",
+            curr / 1000 / 1000,
+            min / 1000 / 1000,
+            max / 1000 / 1000);
+error:
+    if (pCoreRef != NULL)
+        CFRelease(pCoreRef);
+    if (eCoreRef != NULL)
+        CFRelease(eCoreRef);
+    if (iter)
```
giampaolo, line 207: Should this be `if (iter != 0)`?

psutil/arch/osx/cpu.c
```diff
@@ -109,7 +114,103 @@ psutil_cpu_stats(PyObject *self, PyObject *args) {
...
+    if (status != KERN_SUCCESS) {
+        PyErr_Format(PyExc_RuntimeError, "IOServiceGetMatchingServices call failed");
+        goto error;
+    }
+
+    while ((entry = IOIteratorNext(iter))) {
+        io_name_t name;
+        status = IORegistryEntryGetName(entry, name);
+        if (status != KERN_SUCCESS) {
+            if (entry)
```
giampaolo, line 147: If you define the while loop like this (see my previous remark):

```c
while ((entry = IOIteratorNext(iter) != 0)) {
     ...
}
```

...this `if (entry)` check should be unnecessary.

giampaolo (changes requested):

snOm3ad:
For the merging, I believe you can squash and merge from Github. Otherwise, I don't know how you want me to handle the various commits? I can rebase and squash them into one, but I don't know if that would rewrite the commit history as there are other commits in between mine. Let me know.

giampaolo:
Also, can you paste the output of ` python3 -c "import psutil; print(psutil.cpu_freq())"`? I'm curious to see what these numbers look like.

As for the merge, don't worry: I always "squash and merge" everything via the Web UI.

snOm3ad:
> Also, can you paste the output of these 2 commands?
> 
> ```
> python3 -c "import psutil; print(psutil.cpu_freq())"
> python3 -c "import psutil; print(psutil.cpu_freq(percpu=1))"
> ```
>

I get the following:

```
scpufreq(current=3504, min=702, max=3504)
[scpufreq(current=3504, min=702, max=3504)]
```

Let me get back to you on the tests, but FYI there are some tests that are failing due to #2189 like `psutil.tests.test_contracts.TestFetchAllProcesses.test_all`.

snOm3ad:
I see 3 failing tests on my machine:

psutil.tests.test_contracts.TestFetchAllProcesses.test_all 

<details>
    <summary> see trace </summary>
    
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 346, in wrapper
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 404, in cmdline
        return cext.proc_cmdline(self.pid)
    ProcessLookupError: [Errno 3] assume no such process (originated from sysctl(KERN_PROCARGS2) -> EINVAL)

      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/tests/test_contracts.py", line 389, in proc_info
        info[fun_name] = fun()
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/__init__.py", line 686, in cmdline
        return self._proc.cmdline()
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 351, in wrapper
        raise NoSuchProcess(self.pid, self._name)
    psutil.NoSuchProcess: process no longer exists (pid=1404, name='login')
    
</details>

psutil.tests.test_unicode.TestFSAPIsWithInvalidPath

<details>
    <summary> see trace </summary>

      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/tests/test_unicode.py", line 179, in setUpClass
        create_exe(cls.funky_name)
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/tests/__init__.py", line 859, in create_exe
        shutil.copyfile(PYTHON_EXE, outpath)
      File "/opt/homebrew/Cellar/python@3.10/3.10.11/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 256, in copyfile
        with open(dst, 'wb') as fdst:
    OSError: [Errno 92] Illegal byte sequence: '/private/var/folders/hm/y1b672p17sv_ww02x03q8hcm0000gn/T/@psutil-61554-icml95ikf\udcc0\udc80'

</details>

psutil.tests.test_posix.TestSystemAPIs.test_users_started

<details>
    <summary> see trace </summary>

      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/tests/__init__.py", line 723, in wrapper
        return fun(*args, **kwargs)
      File "/Users/olivertome/Desktop/dev/projects/oss/psutil/psutil/tests/test_posix.py", line 371, in test_users_started
        raise ValueError(
    ValueError: cannot interpret tstamp in who output
    olivertome       console       9 Apr 12:20  old       566
    olivertome       ttys000       9 Apr 12:21 00:05     1404
    olivertome       ttys001      18 Apr 18:12   .      59842

</details>

snOm3ad:
I see 3 failing tests on my machine:

psutil.tests.test_contracts.TestFetchAllProcesses.test_all 

<details>
    <summary> see trace </summary>
    
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 346, in wrapper
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 404, in cmdline
        return cext.proc_cmdline(self.pid)
    ProcessLookupError: [Errno 3] assume no such process (originated from sysctl(KERN_PROCARGS2) -> EINVAL)

      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/tests/test_contracts.py", line 389, in proc_info
        info[fun_name] = fun()
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/__init__.py", line 686, in cmdline
        return self._proc.cmdline()
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/_psosx.py", line 351, in wrapper
        raise NoSuchProcess(self.pid, self._name)
    psutil.NoSuchProcess: process no longer exists (pid=1404, name='login')
    
</details>

psutil.tests.test_unicode.TestFSAPIsWithInvalidPath

<details>
    <summary> see trace </summary>

      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/tests/test_unicode.py", line 179, in setUpClass
        create_exe(cls.funky_name)
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/tests/__init__.py", line 859, in create_exe
        shutil.copyfile(PYTHON_EXE, outpath)
      File "/opt/homebrew/Cellar/python@3.10/3.10.11/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 256, in copyfile
        with open(dst, 'wb') as fdst:
    OSError: [Errno 92] Illegal byte sequence: '/private/var/folders/hm/y1b672p17sv_ww02x03q8hcm0000gn/T/@psutil-61554-icml95ikf\udcc0\udc80'

</details>

psutil.tests.test_posix.TestSystemAPIs.test_users_started

<details>
    <summary> see trace </summary>

      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/tests/__init__.py", line 723, in wrapper
        return fun(*args, **kwargs)
      File "/Users/*****/Desktop/dev/projects/oss/psutil/psutil/tests/test_posix.py", line 371, in test_users_started
        raise ValueError(
    ValueError: cannot interpret tstamp in who output
    *****       console       9 Apr 12:20  old       566
    *****       ttys000       9 Apr 12:21 00:05     1404
    *****       ttys001      18 Apr 18:12   .      59842

</details>

giampaolo:
Thank you! Test failures are unrelated (and needs separate fixing). `current` and `max` are the same though. Is it just a case, or they keep being the same even between different calls?

giampaolo:
Can you paste the output of this?
```python
import psutil, time
print(psutil.cpu_count())
for x in range(5)
     print(psutil.cpu_freq(1))
     time.sleep(1)
```

dbwiddis:
> Thank you! Test failures are unrelated (and needs separate fixing). `current` and `max` are the same though. Is it just a case, or they keep being the same even between different calls?

I suggested we use "max" for "current" to match the current behavior of the x64 mac chips.

I don't think "current" is really a relevant metric anyway, but everyone wants it. :)

giampaolo:
> I don't think "current" is really a relevant metric anyway, but everyone wants it. 

Why do you think so? In my mind `current` is the "changing" value, and `min` and `max` are the nominal values advertised by CPU vendor and that never change. Or at least this is how it works on Linux, which represents the only platform where we can get a real-time value:

```
$ python3 -c "import psutil, time; [print(psutil.cpu_freq()) for x in range(5) if not time.sleep(1)]"
scpufreq(current=800.063125, min=800.0, max=3500.0)
scpufreq(current=800.0755, min=800.0, max=3500.0)
scpufreq(current=799.9767499999999, min=800.0, max=3500.0)
scpufreq(current=2028.0061249999999, min=800.0, max=3500.0)
scpufreq(current=1250.0946250000002, min=800.0, max=3500.0)
```

From psutil doc:

> On Linux current frequency reports the real-time value, on all other platforms this usually represents the nominal “fixed” value (never changing). 

If possible we should try to replicate the exact same behavior here, so we remain consistent with Linux, which represents the best platform in terms of `cpu_freq()` accuracy. Are we able to get a real-time value on Apple M1? What about "fixed" `min / max` values?

snOm3ad:
> If possible we should try to replicate the same behavior here, so we remain consistent with Linux, which represents the best platform in terms of `cpu_freq()` accuracy. Are we able to get a real-time value on Apple M1? What about "fixed" `min / max` values?

It is possible, but I have not seen this being done anywhere else in `C`. What I have found from my research is that the API seems to only be available in C++ (see [IOKernelReporters.h](https://opensource.apple.com/source/xnu/xnu-4570.41.2/iokit/IOKit/IOKernelReporters.h.auto.html)), so at the very least we know we would have to dynamically link to it. The symbols are defined in `/usr/lib/libIOReport.dylib`.

Additionally, the lack of documentation regarding the properties contained in the outputs makes this a difficult feat. For example, the tool created by @BitesPotatoBacks works well with M1 CPUs but breaks in M2 chips as there are some missing properties in the output.

dbwiddis:
> > I don't think "current" is really a relevant metric anyway, but everyone wants it.
> 
> Why do you think so? In my mind current is the "changing" value, and min and max are the nominal values advertised by the CPU vendor and that never change. 

The simple answer is, "because that's how x86 Macs do it."  The `sysctl` value for `hw.cpufrequency` returns the max.

> Or at least this is how it works on Linux, which is the only platform where we can get a real-time value:

 - "value" not "values"
   - We have multiple cores, each of them potentially operating at a different frequency.  
   - Do we report on the core doing the work or the other ones which are idle? Or do we average them?
 - "real-time"
   - If I report at the time I'm measuring it, I'm doing work on that core, raising its frequency.  It's just like [Heisenberg's Uncertainty Principle](https://scienceexchange.caltech.edu/topics/quantum-science-explained/uncertainty-principle). 
   - If I report an average of some previous amount of time, it's no longer real-time.  That's what `powermetrics` does, it averages over 5 seconds.  Do you want users to wait 5 seconds to get a measure of the frequency? 

In fact, the "real-time" CPU frequency for the core processing this calculation is most likely to be the max.  The way the Apple Silicon chips work is:
 - They idle at the min
 - When doing work, they spin up to higher frequencies; the E-cores have two higher frequencies they go to depending on load; the P-cores just go min to max
 - The spin-up time goes through several intermediate frequencies (see the voltage-states output above) and take about 70 ms to get there.  
 - When done with work, they drop back down to min, in about 5 ms
 - The average human reads 300 words per minute.  That's 200 ms per word.  Even if we ignored the time it takes to output the value, by the time the average human has read the output, it's changed (probably to min).

So the most accurate "average" frequency is really probably a weighted average of cpu usage and min/max frequencies (percent usage * max + percent idle * min).  But we don't have cpu usage handy, we just have the processor ticks that we have to measure across some time interval.  What time interval should we use?  Shouldn't that be something the user does... and in fact can do to get per-core frequencies, and can then do this identical math themselves if they want a changing value.

We do have `loadavg` which is a decent proxy for "# of cores doing work".  We could use that to create a weighted average of min and max and probably get reasonably close.  We'd still be using the shortest (1 minute ) expoentially weighted moving average, which doesn't quite get to "near real time" but gives you a quick, changing, single number that's most likely closer to truth than any other single number we spit out.

giampaolo:
@snOm3ad
> Additionally, the lack of documentation regarding the properties contained in the outputs makes this a difficult feat. 

Fair enough. I know the feeling. Let's just try to just return a reasonable value.

>  For example, the tool created by @BitesPotatoBacks works well with M1 CPUs but breaks in M2 chips as there are some missing properties in the output.

For reference / note to self: 
https://github.com/giampaolo/psutil/issues/1892#issuecomment-989959636
https://github.com/BitesPotatoBacks/SocPowerBuddy
For clarity: what you're doing here tries to fix both M1 and M2? Could you please also update HISTORY explaining what is being fixed and on what platforms / architectures. Thanks.
Also CC @BitesPotatoBacks just in case he wants to comment on this. Judging from https://github.com/BitesPotatoBacks/SocPowerBuddy this is quite a deep rabbit hole.

> I believe the output we get from IORegistry regarding the min and max frequencies is fixed.

Since I cannot test this myself I'm still not clear what's currently "fixed" and "variable". =) Can you please paste the output of?

```
$ python3 -c "import psutil, time; [print(psutil.cpu_freq()) for x in range(5) if not time.sleep(1)]"
```

snOm3ad:
@giampaolo

Here's the output for the command you mentioned:

```
scpufreq(current=3504, min=702, max=3504)
scpufreq(current=3504, min=702, max=3504)
scpufreq(current=3504, min=702, max=3504)
scpufreq(current=3504, min=702, max=3504)
scpufreq(current=3504, min=702, max=3504)
```

As you can see the current frequency is always the `max` frequency, and both `min` and `max` are fixed values. I had originally intended to return `None` for the current frequency. I think this is a more transparent approach as it lets users know that we're not making any effort to calculate this metric for them. 

Let me know your thoughts. I also updated the description in this PR as well as the HISTORY file.

dbwiddis:
> I had originally intended to return None for the current frequency. I think this is a more transparent approach as it lets users know that we're not making any effort to calculate this metric for them.

If we do that I'd say "unknown" rather than "none".  Because there is a frequency, we just don't know it.

However, I stand by my opinion (but its' up to @giampaolo to decide here) that we should either return the max here, or change the x86 code to match whatever we do here.  I think it's more important to return a consistent result to current users who just changed their hardware and may be relying on the fact that "current" has always been max on Apple hardware.

Here's what we return on x86:
https://github.com/giampaolo/psutil/blob/995fa82a5f0b5013ca5535a9f1669c96d07ef4ca/psutil/arch/osx/cpu.c#L122-L124

This is [documented as returning the max](https://opensource.apple.com/source/xnu/xnu-792.12.6/libkern/libkern/sysctl.h):
> The current frequency is the max frequency in the current mode.

giampaolo:
>  I think it's more important to return a consistent result to current users who just changed their hardware and may be relying on the fact that "current" has always been max on Apple hardware.

Yes. If this PR is consistent with the values that we return on x86 then let's merge it.

giampaolo:
I have fixed the formatting (limited lines to 79 chars).

CarterLi:
Why not use `IOServiceNameMatching("pmgr")`? It should be easier and faster.

dbwiddis:
> It should be easier 

Perhaps easier to write at the start, but since the work has been done, the "easier" thing to do is leave it as is.

> and faster.

I'm not so sure. We're limiting the iteration to a particular service, and then matching a name and stopping on success.  Your proposal matches a name across the entire IO Registry may have to check for more matches (and won't stop on success as it needs to collect all matches).

We're probably talking a few nanoseconds here or there but I'm skeptical on it being "faster" without seeing a benchmark.

Then again, I don't know the inner workings of the registry and maybe there's an O(1) lookup somewhere on the name....

giampaolo:
Sorry, I lost track on this. Is it good to go? Should we merge it?

snOm3ad:
From my side it's good to go! I did notice that someone put up an issue saying the problem remains when using Docker (#2318). I believe we'd need to add some runtime checks to address this, so not sure if it should be included in the same PR or treated as a different issue?

giampaolo:
I would say let's treat it as a different issue.

giampaolo:
Can you move the HISTORY.rst entry on top of the file and also add yourself to CREDITS?

snOm3ad:
@giampaolo the DCO is failing cause some of your commits weren't signed. Other than that it should be good now! Also thank you for looking into this once again :)

giampaolo:
This causes a failure on arm64, see https://github.com/giampaolo/psutil/issues/2382.

dbwiddis:
> This causes a failure on arm64, see #2382.

The Github-provided VMs do not provide the same drivers as real hardware.  Specifically, they don't have P-Cores or E-cores, just virtual CPUs.  So grokking the IORegistry for the `pmgr` just won't work.

See https://eclecticlight.co/2023/10/23/how-does-macos-manage-virtual-cores-on-apple-silicon/ for some interesting discussion, and https://github.com/oshi/oshi/issues/2470#issuecomment-1899990202 for my Java-based encounter with this oddness.

PR #2247, opened 2023-04-22
Run `abi3audit` on produced abi3 wheels

mayeut:
## Summary

* OS: All
* Bug fix: more or less
* Type: wheels

## Description

Run `abi3audit` on produced abi3 wheels

This showed a mismatch for Windows abi3 wheels. While I don't think it will actually break something or show up as bug, I moved abi3 to cp37+ on Windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,3 +135,4 @@ jobs:
       - run: |
           python scripts/internal/print_hashes.py wheelhouse/
           pipx run twine check --strict wheelhouse/*
+          pipx run abi3audit --verbose --strict wheelhouse/*-abi3-*.whl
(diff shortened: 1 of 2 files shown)


giampaolo:
Nice. I like this. I think the extra `abi3audit` should be part of `make pre-release` and `make release` targets. They already do a similar check (`python3 -m twine check --strict dist/*`). Can you please add that to the Makefile?

mayeut:
> Can you please add that to the Makefile (also means adding a new dep to PY3_DEPS list)?

abi3audit only works with Python 3.8+, it seems like a lot of extra complexity to be added in the Makefile unless we can assume it will only be run with Python 3.8+.

giampaolo:
Clear. Merging it then.

giampaolo:
I added Makefile integration in https://github.com/giampaolo/psutil/commit/4d5b9e89b5f258b71d8caab0d720e8762a31eb92.

PR #2248, opened 2023-04-22
doc: Reformat the oneshot speedup-groups information without the table layout

ferdnyc:
## Summary

* OS: all
* Bug fix: no
* Type: doc
* Fixes:

## Description

IMHO the table layout used for the list of methods sped up by `oneshot()` was not really working. Horizontal alignment of various methods meant nothing, so having the platforms side-by-side also meant nothing. Blank rows were hard to spot, making the groupings unclear. Basically, it was just a list of groups-of-methods, formatted so it's difficult to read on narrow screens.

This PR folds up each column of the table into a sub-sub-section of a subsection (titled "Method grouping & speedup potential") within the method documentation for `oneshot()`. Each platform section contains an autonumbered ordered list containing from one to three entries, each of which is a comma-separated list of method names.

Using this format also allowed me to notice that the Solaris and AIX columns were identical, and combine them into a `Solaris & AIX` section rather than repeating info.

I'm open to changing any or all of this as requested, and of course you're welcome to reject it outright if you feel it's the wrong approach.

### Example

In the source, the entries are listed vertically, each on a separate line, which makes editing vastly easier. For example:

```rst
    Linux
    """""

    #. :meth:`cpu_num`,
       :meth:`~Process.cpu_percent`,
       :meth:`cpu_times`,
       :meth:`create_time`,
       :meth:`name`,
       :meth:`ppid`,
       :meth:`status`,
       :meth:`terminal`

    #. :meth:`gids`,
       :meth:`num_ctx_switches`,
       :meth:`num_threads`,
       :meth:`uids`,
       :meth:`username`

    #. :meth:`memory_full_info`,
       :meth:`memory_maps`

    *Speedup: +2.6x*
```

Rendered, that becomes:

image

(Or it may wrap, depending on the viewport width.)

### Other changes

Other adjustments I made in the course of doing this:

1. Fixed the alphabetization in BSD (`name` was before `memory_info`), something that would've been tedious to do in the table format, but with this formatting was a simple matter of moving one source line down by two lines.

1. Updated the `.. versionadded` for the method from `5.0.0` to `4.5.0`, matching other mentions elsewhere in the file.

1. Moved the `.. versionadded` above the new subsection, which seemed to be clearer / make more sense, as having it at the very bottom gave the impression that only the last platform was added in 4.5.0.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1113,54 +1113,121 @@ Process class
     ...
     >>>
 
-    Here's a list of methods which can take advantage of the speedup depending
-    on what platform you're on.
-    In the table below horizontal empty rows indicate what process methods can
-    be efficiently grouped together internally.
-    The last column (speedup) shows an approximation of the speedup you can get
+    .. versionadded:: 4.5.0
+
+    Method grouping & speedup potential
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Here's a list of method groups for each platform which can benefit
+    from speedups.
+    Methods in each group can be efficiently grouped together internally.
+    Calling any method in a group will cache the values for all other
+    methods in the same group.
+    "Speedup" is an approximation of the speedup you can get
     if you call all the methods together (best case scenario).
 
-    +------------------------------+-------------------------------+------------------------------+------------------------------+--------------------------+--------------------------+
-    | Linux                        | Windows                       | macOS                        | BSD                          | SunOS                    | AIX                      |
-    +==============================+===============================+==============================+==============================+==========================+==========================+
-    | :meth:`cpu_num`              | :meth:`~Process.cpu_percent`  | :meth:`~Process.cpu_percent` | :meth:`cpu_num`              | :meth:`name`             | :meth:`name`             |
-    +------------------------------+-------------------------------+------------------------------+------------------------------+--------------------------+--------------------------+
-    | :meth:`~Process.cpu_percent` | :meth:`cpu_times`             | :meth:`cpu_times`            | :meth:`~Process.cpu_percent` | :meth:`cmdline`          | :meth:`cmdline`          |
-    +------------------------------+-------------------------------+------------------------------+------------------------------+--------------------------+--------------------------+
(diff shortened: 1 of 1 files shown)


giampaolo:
Closing as outdated.

ferdnyc:
@giampaolo While I agree this PR is very outdated, the basic table structure in question is still present at https://psutil.io/performance/#methods-sped-up-by-oneshot, and still (IMHO) not the best way of presenting that information.

I'm willing to redo this PR for the updated docs, if there's any interest in merging it. If not, then I won't bother.

giampaolo:
Current code is already using a listed table

```
.. list-table::
   :header-rows: 1
   :class: wide-table

   * - Linux
     - Windows
     - macOS
     - BSD
   * - :meth:`~Process.cpu_num`
     - :meth:`~Process.cpu_percent`
     - :meth:`~Process.cpu_percent`
     - :meth:`~Process.cpu_num`
```

...basically it's the same change you did here. I did it some time ago when I rewrote the whole doc (but I forgot you did it already in this PR).

ferdnyc:
Agreed the **source** formatting is cleaner, but IMHO the _rendered_ table presentation is unnecessarily busy. (It forces the user to read _down_ a column, rather than across a row, for relevant information. And each column is completely separate from the adjacent ones. Adjacent methods have no relevance — there's no meaning whatsoever to any of the rows in the table.)

Basically, as I interpreted the data it's just four sets of lists-of-methods (each displayed as a table column), so that's how my changes displayed it. The rendered output was:

image

(Though the fonts and etc. would look different on the live site.)

But if you're happy with the table presentation, that's cool.

giampaolo:
Ah I get what you mean now. You propose to remove the table and use a bullet list, per-platform, showing which methods are grouped. Is that correct?
Indeed it may be better. Want to try making it in a new PR and let's see how it looks? (I honestly don't know, but I'm more keen on thinking that your idea is better compared to the table)

ferdnyc:
Exactly, yes. I'll put together a replacement PR with edits to the current doc.

PR #2260, opened 2023-05-28
Set `Process.info` to `None` by default

sobolevn:
Refs https://github.com/python/typeshed/issues/10195

## Summary

* OS: any
* Bug fix: yes
* Type: core
* Fixes: None

## Description

Right now `.info` attribute is only set during `process_iter`, which is not quite correct.
This results in `AttributeError` when using regular `Process`:

```python
>>> import psutil
>>> psutil.Process(1)
psutil.Process(pid=1, name='launchd', status='running', started='2023-05-27 09:31:50')
>>> psutil.Process(1).info
AttributeError: 'Process' object has no attribute 'info'
```

I propose using `None` default for cases when it is not created using `process_iter`.

This way it will simplify typing of `Process` for us.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -378,8 +378,8 @@ class Process(object):
         # it might refer to a process whose PID has been reused).
         # This will be used later in __eq__() and is_running().
         self._ident = (self.pid, self._create_time)
-        # Is set only in `process_iter()`:
-        self.info = None
+        # Is modified only in `process_iter()`:
+        self.info = {}
 
     def __str__(self):
         info = collections.OrderedDict()
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -378,6 +378,8 @@ class Process(object):
         # it might refer to a process whose PID has been reused).
         # This will be used later in __eq__() and is_running().
         self._ident = (self.pid, self._create_time)
+        # Is set only in `process_iter()`:
+        self.info = None
 
     def __str__(self):
         info = collections.OrderedDict()


giampaolo:
Hmm... debatable. 
I think I'm more keen on agreeing with you though: I suppose it's less surprising for the user and more consistent if `info` attribute is always set. 

I wonder if it's more correct for it to be an empty dict `{}` rather than `None` though.


Uhmm...

sobolevn:
Oh, whatever you like, I am not a user of this feature, so I cannot really decide.

emcek:
> Hmm... debatable. I think I'm more keen on agreeing with you: I suppose it's less surprising for the user if `info` attribute is always set. I wonder if it's more correct for it to be an empty dict `{}` rather than `None` though. Uhmm...

Yes, I think empty dict is fine!

ringohoffman:
I came here to raise a similar issue. This issue from the docs raises type errors since `info` is not defined in the `__init__`:

https://psutil.readthedocs.io/en/latest/#psutil.process_iter:
```python
import psutil
for proc in psutil.process_iter(['pid', 'name', 'username']):
    print(proc.info)  # Cannot access member "info" for type "Process"; Member "info" is unknown
```

We would need to update the type stubs accordingly: https://github.com/python/typeshed/blob/e347af1d59ed46478a788ed60fd02096fce26edd/stubs/psutil/psutil/__init__.pyi#L158

ringohoffman:
```
psutil-debug [psutil/arch/windows\disk.c:134]> DeviceIoControl -> ERROR_INVALID_FUNCTION; ignore PhysicalDrive0
psutil-debug [psutil/arch/windows\socks.c:61]> GetExtendedTcpTable: retry with different bufsize
    main()
    print_sysinfo()
  File "c:\projects\psutil\psutil\tests\__init__.py", line 1195, in print_sysinfo
  File "c:\projects\psutil\psutil\__init__.py", line 540, in as_dict
    ret = meth()
TypeError: 'dict' object is not callable
```

🤔 is this related to these changes? This seems unexpected.

sobolevn:
Oh, closing in favor of https://github.com/giampaolo/psutil/pull/2267
Sorry, I haven't seen your PR! 👍

PR #2267, opened 2023-06-07
add settable info property on psutil.Process, default to empty dict

ringohoffman:
## Summary

* OS: linux, darwin
* Bug fix: no
* Type: new-api (adding an official `property` for `psutil.Process.info`) 
* Related: https://github.com/python/typeshed/issues/10195, https://github.com/giampaolo/psutil/pull/2260

## Motivation
This example from the docs raises type errors since `info` is not defined in the `__init__`:

https://psutil.readthedocs.io/en/latest/#psutil.process_iter:
```python
import psutil
for proc in psutil.process_iter(['pid', 'name', 'username']):
    print(proc.info)  # Cannot access member "info" for type "Process"; Member "info" is unknown
```

## Description

This PR is an alternate implementation of #10195, addressing @giampaolo's comment: https://github.com/giampaolo/psutil/pull/2260#issuecomment-1566821018

> I suppose it's less surprising for the user if `info` attribute is always set. I wonder if it's more correct for it to be an empty dict `{}` rather than `None` though.

Which I agree with.

Here is the typeshed update corresponding to this PR: https://github.com/python/typeshed/pull/10275

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -348,6 +348,7 @@ class Process(object):
         self._pid_reused = False
         self._hash = None
         self._lock = threading.RLock()
+        self._info = {}
         # used for caching on Windows only (on POSIX ppid may change)
         self._ppid = None
         # platform-specific modules define an _psplatform.Process
@@ -424,6 +425,17 @@ class Process(object):
     def pid(self):
         """The process PID."""
         return self._pid
+
+    @property
+    def info(self):
+        """Stored result of proc.as_dict(attrs=attrs) when process_iter() is
+        called with the attrs parameter.
+        """
+        return self._info
+
+    @info.setter
+    def info(self, value):
+        self._info = value
 
     # --- utility methods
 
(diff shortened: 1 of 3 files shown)


ringohoffman:
```
psutil-debug [psutil/arch/windows\disk.c:134]> DeviceIoControl -> ERROR_INVALID_FUNCTION; ignore PhysicalDrive0
psutil-debug [psutil/arch/windows\socks.c:61]> GetExtendedTcpTable: retry with different bufsize
    main()
    print_sysinfo()
  File "c:\projects\psutil\psutil\tests\__init__.py", line 1195, in print_sysinfo
  File "c:\projects\psutil\psutil\__init__.py", line 540, in as_dict
    ret = meth()
TypeError: 'dict' object is not callable
```

🤔 is this related to these changes? This seems unexpected.

ringohoffman:
@giampaolo how’s this looking?

ringohoffman:
@giampaolo have you had a chance to check this out?

qthequartermasterman:
@giampaolo Is there any plan to look at this? I would like to use it in a project of mine.

ringohoffman:
@giampaolo any chance you can look at this?

PR #2297, opened 2023-08-18
Fix process_iter() not running for entire process

chozekun:
## Summary

* OS: Ubuntu 18.04.1 (Bionic Beaver)
* Bug fix: yes
* Type: core
* Fixes:

## Description
There is a bug where pids() very occasionally returns fewer processes than it actually has. Therefore, instead of deleting it right away, check if the actual process is alive before deleting it.
(Sorry for my poor English)


chozekun:
Closed due to test failure.

PR #2305, opened 2023-09-13
Exclude external batteries from sensors_battery

RodoMa92:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core

## Description
The single check for battery in power_supply was causing sometimes to grab external device battery statuses since they have the form of hid_\<mac\>_battery, which match the bat filter. 

However they do not have an ACPI path to the system itself. 

This takes into account that detail and properly exclude them from being included as an internal battery.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1471,9 +1471,13 @@ def sensors_battery():
                 except ValueError:
                     return ret.strip()
         return None
+    # The final check verifies if the reported battery is internal
+    # since the existence of the device/path file represent an available
+    # ACPI path to the battery itself.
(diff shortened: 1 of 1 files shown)


jardon:
@RodoMa92 @giampaolo  if this only requires signing the commit, can we get this pushed through?

RodoMa92:
> @RodoMa92 @giampaolo  if this only requires signing the commit, can we get this pushed through?

I'm not sure if this will not break under other platforms like arm, so it should be properly tested before a merge. Might not behave like on x86 platforms, sadly I do not have such hardware available to me.

PR #2322, opened 2023-10-25
Added pieshell to list of projects using psutil

redhog:
## Summary

* OS: all
* Bug fix: no 
* Type: doc
* Fixes: 

## Description

Added pieshell to the list of projects using psutil. Pieshell is a Python shell environment that combines the expressiveness of shell pipelines with the power of python iterators. It uses psutil to provide detailes on processes it has started, and to provide a "process browsing" functionality for running processes.


giampaolo:
Sorry, I don't consider it relevant enough.

PR #2330, opened 2023-11-29
Update README.rst

shawn-bluce:
fix psutil.cpu_times() output

## Summary

* OS: Linux
* Bug fix: yes
* Type: doc

## Description

README.rst have a bug. On code's output `psutil.cpu_times()`, this output have two `nice`, but the last `nice` should be `guest_nice`.


giampaolo:
Good catch, thanks.

PR #2224, opened 2023-04-10
wheels: add arm64 builds

shawaj:
## Summary

* OS: windows, arm, mac
* Bug fix: no
* Type: wheels
* Fixes: #2206, #1972, #1966 

## Description

- add arm64 builds via qemu
- add macos universal2 build
- add windows arm64 build
- bump cancel workflow action to latest to remove nodejs warnings
- bump cibuildwheel action to latest

PR #2375, opened 2024-02-09
chore: build macOS arm64 wheels on macos-14

mayeut:
## Summary

* OS: macOS
* Bug fix: no
* Type: wheels
* Fixes:

## Description

macos-14 runners are running on arm64 processors.
This allows to test macOS arm64 wheels.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ concurrency:
 jobs:
   # Linux + macOS + Windows Python 3
   py3:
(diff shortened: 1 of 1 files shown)


MrMino:
I believe this duplicates https://github.com/giampaolo/psutil/pull/2224

mayeut:
This is not a duplicate of #2224
This PR does not change the wheels that are built but allows to test macOS arm64 ones. #2224 adds universal2 for macOS which but does not test those wheels on arm64 (neither the universal2 nor the arm64 wheel).

PR #2376, opened 2024-02-22
Linux: skip offline cpu cores in cpu_freq

### LINKED ISSUE #2254: [Linux] offline cpus raise NotImplementedError in cpu_freq
python3-psutil (5.9.2) on fedora 38

turn cpu1 off:
`echo 0 > /sys/devices/system/cpu/cpu1/online`

or every cpu but cpu0:
`for i in $(seq 1 $(nproc --ignore=1)); do echo 0 > /sys/devices/system/cpu/cpu$i/online; done
`
cpu_freq needs one more case to skip offline cpus (or ones without the sysfs files for frequency, which means the same thing)

https://github.com/giampaolo/psutil/blob/7eadee31db2f038763a3a6f978db1ea76bbc4674/psutil/_pslinux.py#L749

shadeyg56:
## Summary
Fixes #2254 

Previously offline CPU cores would return an error in `cpu_freq()` since psutil didn't skip them. 
With this change, any CPU core found to be offline will return `scpufreq(0, 0, 0)`
I figured it would be pretty common for programs to expect the cores to be returned in order which is why I had it append a tuple with all-zero values. If this is unwanted, it could be easily removed to only return online cpus and their frequencies.

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: #2254 

## Description

Runs `cat` on `/sys/devices/system/cpu/cpu{i}/online` to determine if a CPU is offline or not.  If `cat` returns 0, then `scpufreq(0, 0, 0)` is appended and the loop continues to the next iteration. Otherwise, the function runs as normal.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -821,3 +821,7 @@ I: 2222
 N: Ryan Carsten Schmidt
 W: https://github.com/ryandesign
 I: 2361
+
+N: Shade Gladden
+W: https://github.com/shadeyg56
+I: 2376
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,7 @@
 **Bug fixes**
 
 - 2360_, [macOS]: can't compile on macOS < 10.13.  (patch by Ryan Schmidt)
+- 2254_, [Linux]: offline cpus raise NotImplementedError in cpu_freq() (patch by Shade Gladden)
 
 5.9.8
 =====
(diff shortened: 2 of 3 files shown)


giampaolo:
Nice. Some remarks:

* I'd rather put the new logic in [here](https://github.com/giampaolo/psutil/blob/da0b7c6880165fe45de68ee819c68e304958498a/psutil/_pslinux.py#L781-L783), after `if curr is None`, and raise `NotImplementedError` if also the new file is unreadable.
* the `0` value should be a float (`0.0`)
* if you can, also update HISTORY.rst / CREDITS files

Thank you

shadeyg56:
Do I need to remove the f-string in order to support python 2?

giampaolo:
Oh right. Yes.

Issue #2272, opened 2023-06-27
[Linux] psutil.NoSuchProcess exception can't be pickled


anthonyryan1:
## Summary

* OS: Linux
* Architecture: 64bit
* Psutil version: 5.9.5
* Python version: 3.11.4
* Type: core, tests

## Description

`psutil.NoSuchProcess` exceptions can't be pickled and sent through a pipe. Normally, one would expect python exceptions to be pickleable so they can be moved.

```python
#!/bin/env python3
import multiprocessing
import psutil

# err = Exception("A standard exception works")
err = psutil.NoSuchProcess(pid=123)

recv_pipe, send_pipe = multiprocessing.Pipe(duplex=False)

send_pipe.send(err)

output = recv_pipe.recv()
# assert isinstance(output, Exception)
assert isinstance(output, psutil.NoSuchProcess)
```

PR #2380, opened 2024-03-14
Add pickle support to psutil Exceptions

anthonyryan1:
## Summary

* OS: N/A
* Bug fix: yes
* Type: core
* Fixes: 2272

## Description

Add `__reduce__` method to exceptions commonly thrown, and start testing that exceptions can be pickled in test_serialization.

Prior to this, it was not possible to pickle a psutil exception (for error reporting), and this also made it impossible to send them through multiprocessing pipes. Code demonstrating the problem is available in #2272

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -825,3 +825,7 @@ I: 2361
 N: Shade Gladden
 W: https://github.com/shadeyg56
 I: 2376
+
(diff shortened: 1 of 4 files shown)


giampaolo:
Nice change. I like it.

PR #2417, opened 2024-05-18
chore(ci): update actions

mayeut:
## Summary

* OS: N/A
* Bug fix: no
* Type: CI
* Fixes:

## Description

This notably removes all warnings related to Node.js 16 actions being deprecated.

CPython 3.13 is now tested with the cibuildwheel update.

Additional commits:
- fix the linter workflow
- test python 3.8 on macOS arm64

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,7 +108,7 @@ jobs:
         python-version: 3.x
     - name: 'Run linters'
       run: |
-        python3 -m pip install ruff==0.3.4 black rstcheck toml-sort sphinx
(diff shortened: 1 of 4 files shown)


giampaolo:
Hi Matthew.

> test python 3.8 on macOS arm64

This is just running tests or also producing a new .whl for python 3.8 on macOS arm64?

mayeut:
This is just running the tests. The wheel is already there but only tested on Python 3.9+

giampaolo:
Got it, thanks.

PR #2428, opened 2024-06-24
Fix GIL warnings and a few thread-safety issues in free-threaded CPython

### LINKED ISSUE #2427: psutil fails (segfault) on Mac OS in the free-threaded (no GIL) version of Python 3.13
## Summary

* Mac OS 12.7.5
* Architecture: 64bit ARM (Apple M1)
* Psutil version: 6.0.0
* Python version: Python 3.13.0b2
* Type: fails to import

## Description

Transfered here from https://github.com/python/cpython/issues/120717

I used pip to install psutil on python 3.13 0b2, both the regular and the "t" version. The regular one works; the t one fails at import psutil with a segfault.

The Python teams comments: "Any extensions that uses more than Python would need to adapt to free-threading if they want to onboard. It's impossible for CPython to ensure all the C extensions work without GIL. So this should be an issue for psutil ."

colesbury:
## Summary

* OS: multiple
* Bug fix: yes
* Type: core
* Fixes: #2427

## Description

This adds `PyUnstable_Module_SetGIL` calls to allow `psutil` to run with the GIL disabled in the CPython free-threading builds. This also fixes a few thread-safety that could cause problems with the GIL disabled: 

- The temporary `argv` C array is no longer global in OpenBSD's `proc_cmdline`
- The `maxcpus` variable is no longer global in FreeBSD's `per_cpu_times`.

### DIFF
--- a/psutil/arch/openbsd/proc.c
+++ b/psutil/arch/openbsd/proc.c
@@ -193,7 +193,8 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
     return py_retlist;
 
 error:
-    free(argv);
+    if (argv != NULL)
+        free(argv);
     Py_XDECREF(py_arg);
     Py_DECREF(py_retlist);
     return NULL;
--- a/psutil/arch/openbsd/proc.c
+++ b/psutil/arch/openbsd/proc.c
@@ -147,7 +147,7 @@ PyObject *
 psutil_proc_cmdline(PyObject *self, PyObject *args) {
     pid_t pid;
     int mib[4];
-    static char **argv;
+    char **argv = NULL;
     char **p;
     size_t argv_size = 128;
     PyObject *py_retlist = PyList_New(0);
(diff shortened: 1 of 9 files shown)


psutil/arch/openbsd/proc.c
```diff
@@ -189,9 +189,11 @@ psutil_proc_cmdline(PyObject *self, PyObject *args) {
         Py_DECREF(py_arg);
     }
 
+    free(argv);
     return py_retlist;
 
 error:
+    free(argv);
```
giampaolo, line 196: Shouldn't this be?

```c
if (argv != null)
     free(argv);
```

giampaolo (changes requested):

psutil/arch/openbsd/proc.c
colesbury: `free()` handles NULL pointers, but I can add the `if (argv != NULL)` guard if you prefer that style.

https://en.cppreference.com/w/c/memory/free#:~:text=If%20ptr%20is%20a%20null%20pointer%2C%20the%20function%20does%20nothing

giampaolo: Thank you. 
Side story: years ago psutil received a CVE due to a double `free()`, so I believe I started using this style since then.

pkoning2: Double free is a separate issue: to avoid that you can use the coding style:
```
free(p);
p = NULL;
```

PR #2440, opened 2024-09-04
Bug 2439 fix for thermal zones

siena-sam:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core
* Fixes: 2439

## Description

Store and clear values for the high and critical temperatures read from `thermal_zone#/trip_point_#_temp` files. Otherwise, values will be incorrectly divided by 1000 (to account for milidegrees) for each trip point after the first that is of type `critical` or `high`.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -835,3 +835,7 @@ I: 2272
 N: Sam Gross
 W: https://github.com/colesbury
 I: 2401, 2427
+
+N: Siena Richard
+W: https://github.com/siena-sam
+I: 2439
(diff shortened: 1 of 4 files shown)


siena-sam:
For confirmation, here are some screenshots of this fix verified on the ARM system we first noticed this on.

Here is the bug seen in 5.9.8 which is the version we are currently using:
image


Here is the bug fixed when running commit `7f705e4b1db4f0b220f84a17a02669a1e3e02eac`:
image

siena-sam:
For confirmation, here are some screenshots of this fix verified on the ARM system we first noticed this on. 

Here is the bug seen in 5.9.8 which is the version we are currently using:
image


Here is the bug fixed when running commit `7f705e4b1db4f0b220f84a17a02669a1e3e02eac`:
image

PR #2442, opened 2024-09-27
Define Py_GIL_DISABLED macro on Windows

cvijdea-bd:
## Summary

* OS: Windows
* Bug fix: yes
* Type: wheels
* Fixes: bad build for 3.13t (free-threading/nogil) on windows

## Description

https://github.com/python/cpython/issues/111650
CPython doesn't ensure Py_GIL_DISABLED is defined when building against the free-threading interpreter, and setuptools doesn't either.

If it is not defined the resulting module will have the 313t abi tag but will crash on import due to linking to `python313.dll` instead of `python313t.dll`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -224,6 +224,9 @@ if WINDOWS:
         # see: https://github.com/giampaolo/psutil/issues/348
         ('PSAPI_VERSION', 1),
     ])
+    
+    if Py_GIL_DISABLED:
+        macros.append(('Py_GIL_DISABLED', 1))
 
     ext = Extension(
         'psutil._psutil_windows',


giampaolo:
@colesbury is it necessary to do this for all C extensions on Windows? Shouldn't cPython do it automatically?

cvijdea-bd:
Opened https://github.com/pypa/setuptools/issues/4662 since discussion on the linked cpython issue seemed to imply they believe it should be setuptools' job to handle it.

colesbury:
> is it necessary to do this for all C extensions on Windows?

Yes, I think so, at least for now.

> Shouldn't cPython do it automatically?

That would be much better, but I think the way the Python Windows installer works makes that not possible. We should probably work on getting setuptools to define `Py_GIL_DISABLED` on Windows for free-threaded extensions.

giampaolo:
Thank you @colesbury and @cvijdea-bd! I am merging this then.

PR #2450, opened 2024-10-03
use setup() options to pass py_limited_api to bdist_wheel & bump cibuildwheel

mayeut:
## Summary

* OS: all
* Bug fix: no
* Type: wheels
* Fixes: 

## Description
With wheel being merged in setuptools, the trick used to relabel wheels by re-implementing wheel.bdist_wheel might stop to work in the future. Use setup options kwarg to pass the information down to bdist_wheel.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,7 @@ jobs:
       if: matrix.arch == 'aarch64'
 
     - name: Create wheels + run tests
-      uses: pypa/cibuildwheel@v2.19.2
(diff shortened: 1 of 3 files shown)


giampaolo:
Thanks. It's good to have you as I really am detached from these topics related to wheels, their integration with github actions etc. In addition it seems the toolings surrounding them (cibuildwheel, setuptools, etc.) change very rapidly and it's hard to keep up. Life's not easy for maintainers of C extension modules.

Issue #2143, opened 2022-09-19
[QNX] QNX support


large-cat:
## Summary

* OS: QNX 7.0 
* Type: core

## Description
  Whether psutil can be supported on qnx.
  I want to use psutil on "qnx 7.0". Could you help me complete the adaptation?
  Looking forward to your reply. 
  Thank you so much.

giampaolo:
It looks like this is a proprietary platforms, with very few potential downloads. More or less, the same considerations for rejecting [IBM i](https://github.com/giampaolo/psutil/pull/1564#issuecomment-748328481) support apply here as well.

PR #1690, opened 2020-02-13
WIP: preliminary Haiku port

mmuman:
Still a lot to do but enough to get [glances running](https://twitter.com/mmu_man/status/1227746291490467842).

PR #1564, opened 2019-08-05
Add support for new platform (IBM i)

ThePrez:
For reference, see lengthy discussions in #1415 and #1444. I found it eventually much more sensible to have IBM i as a completely new module than sharing a codestream with AIX. 
I'll attach the test output shortly. While there is still plenty of test cases that should be cleaned up, the basic functions seem to work well. This fork has been used by several applications for nearly six months, so there is some assurance of quality. 

This new module relies in an external package, but IBM will distribute the package in RPM form as to ensure that does not pose a problem.

PR #2462, opened 2024-10-16
GNU: Add new platform for GNU/Hurd (based off Linux)

zamaudio:
## Summary

* OS: GNU
* Bug fix: no
* Type: new api

## Description

Adds support for new platform GNU/Hurd based on Linux one.

TODO: Fix NotImplementedError:
/proc/diskstats nor /sys/block filesystem are available on this system

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -27,6 +27,7 @@ include psutil/_common.py
 include psutil/_compat.py
 include psutil/_psaix.py
 include psutil/_psbsd.py
+include psutil/_psgnu.py
 include psutil/_pslinux.py
 include psutil/_psosx.py
 include psutil/_psposix.py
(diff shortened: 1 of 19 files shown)


giampaolo:
Thanks for the effort, but I'm afraid I am not gonna merge this work for the same reasons given for IBMi (https://github.com/giampaolo/psutil/pull/1564#issuecomment-748328481), Haiku (https://github.com/giampaolo/psutil/pull/1690#issuecomment-748222016) and QNX (https://github.com/giampaolo/psutil/issues/2143) platforms. 

At a first glance, GNU/Hurd seems even more niche than those platforms. We can't setup a CI for it, you would be the only maintainer, and the number of downloads would be so low that it's not worth the trouble to add (and maintain) the additional 2800+ lines of code to the current source tree.

I think it makes more sense to make this work develop and mature in your own fork. If your fork is good enough, I can list it somewhere in psutil's README or doc.

With that said, I can give you support reviewing your changes. E.g. the first thing I would recommend is to re-use `_pslinux.py` instead of copying it into `_psgnu.py`. If GNU/Hurd supports the /proc FS, as it seems, adding support for it should be much easier than you think, as you can reuse `_pslinux.py` implementation in full, and just add `if GNUHURD: ...` where necessary.

giampaolo:
Closing this out, but impressed by the effort.

PR #2474, opened 2024-12-03
Add option to retrieve number of performance and efficiency cores

### LINKED ISSUE #2034: psutil.cpu_count:  Add argument(s) to allow differentiating "performance cores" from "efficiency cores"
## Summary

* OS: all
* Type: core

## Description

Many applications need to spawn a number of worker threads or processes, where the number of physical cores is the ideal number of workers to create.  The existing `psutil.cpu_count(logical=False)` has served us well in that regard, however, changes in modern hardware are causing that API to become inadequate.

For example, ARM big.LITTLE (including Apple M1), Intel Alder Lake (and rumor has it future AMD CPUs) will feature a mix of "performance cores" and "efficiency cores", some of which may have SMT, and others not, even within the same CPU.

AFAIK, the use-case for `cpu_count` is almost always going to be the performance core count.  Except in the odd corner case where the performance and efficiency cores can be used at the same time, and the performance delta does not matter because the individual jobs are small and there are many of them.

My proposal is to add something like this:
```
class CPUCoreType(Enum):
    ALL = 0
    PERFORMANCE = 1
    EFFICIENCY = 2

def cpu_count(logical: bool=True, core_type: CPUCoreType=CPUCoreType.ALL):
    ...
```

The use of an enum should keep this feature future proof as as CPU core types become more exotic and diverse than what they are today.

Things to consider:
* ARM has their performance and efficiency cores segmented into families, but some CPUs contain only performance cores, and others only efficiency cores.  Should they be reported as such, or if a CPU only has one type of core, should they all be reported as performance cores?
* Are there reliable  cross platform ways to determine this information?  Or will it require maintaining a database, or asking the OS vendors to provide an API to query this information?

ssbarnea:
Fixes: #2034

## Summary

* OS: { type-or-version }
* Bug fix: { yes/no }
* Type: { core, doc, performance, scripts, tests, wheels, new-api }
* Fixes: { comma-separated list of issues fixed by this PR, if any }

## Description

{{{
  A clear explanation of your bugfix or enhancement. Please read the contributing guidelines before submit:
  https://github.com/giampaolo/psutil/blob/master/CONTRIBUTING.md
}}}

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -25,6 +25,7 @@ from __future__ import division
 import collections
 import contextlib
 import datetime
+import enum
 import functools
 import os
 import signal
@@ -221,6 +222,12 @@ _timer = getattr(time, 'monotonic', time.time)
 _TOTAL_PHYMEM = None
 _LOWEST_PID = None
 _SENTINEL = object()
+
+class CPUCoreType(enum.Flag):
+    PERFORMANCE = enum.auto()
+    EFFICIENCY = enum.auto()
+    ALL = PERFORMANCE | EFFICIENCY
+
 
 # Sanity check in case the user messed up with psutil installation
 # or did something weird with sys.path. In this case we might end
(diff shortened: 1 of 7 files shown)


giampaolo:
This PR does not really implement anything. Also as dicussed in https://github.com/giampaolo/psutil/issues/2034 the API we want to come up with is not clear yet. Closing.

PR #2486, opened 2024-12-21
fix: python_requires>=3.6

mayeut:
## Summary

* OS: All
* Bug fix: yes
* Type: build
* Fixes:

## Description
python_requires still allow python<3
with python 2.7 being dropped, it can be set to `python_requires>=3.6`


giampaolo:
:)

PR #2425, opened 2024-06-15
feat: add aarch64 manylinux wheels

mayeut:
## Summary

* OS: Linux
* Bug fix: no
* Type: wheels
* Fixes:
fixes #1966
fixes #1972
fixes #2206
mitigates #2090
mitigates #2175
mitigates #2293
closes #2178
closes #2224

## Description

This PR continues work from #2103 (I've never removed that branch and rebased on master once in a while so that's the same branch  in this PR).

cherry-picked commits from #2126 for missing test dependency in manylinux.

main differences between this PR and #2224:
- This PR does not build `ppc64le` & `s390x`.
- This PR does not build musllinux wheels.
- This PR adds tests for architectures running QEMU:
While the tests needed some modifications in order to pass, IMHO, it's best to check than what can be tested does work as intended.
Running with QEMU however hides some bugs (thinking about /proc/cpuinfo and cpu related stuff here) that actually do exist or comes with its own bugs (s390x and /proc endianness issue is most likely a qemu issue).
aarch64 will most likely be able to run natively [by the end of the year](https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/).

- This PR does not add Windows ARM64 wheel
It could be added without any tests but I'd rather see Microsoft add Windows ARM64 runners to GitHub if they want to get some traction for their new platform (hopefully [by the end of the year](https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/)).

- This PR does not add macOS universal2 wheel
My personal opinion is that it's not useful as long as you provide both x86_64 & arm64 wheels (which is the case).
They do take more space for the end user, pip will always choose the most specific one so publishing 3 will change nothing for almost all users. Remain people packaging universal2 python apps, and, IMHO, this should be handled by the tools creating such apps.

- This PR feels simpler regarding the change of workflow

Regarding tests that are failing on real arm64 hardware but hidden by QEMU testing:
- Using docker on macOS M1:
```
======================================================================
FAIL: psutil.tests.test_system.TestCpuAPIs.test_cpu_freq
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/project/psutil/tests/test_system.py", line 609, in test_cpu_freq
    assert ls, ls
AssertionError: []

======================================================================
FAIL: psutil.tests.test_linux.TestSystemCPUCountCores.test_method_2
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #2503, opened 2025-02-01
ci: build aarch64 wheel on GHA aarch64 runner

mayeut:
## Summary

* OS: Linux
* Bug fix: no
* Type: wheels
* Fixes:

## Description

Update CI workflows to build and test Linux aarch64 on native aarch64 runners instead of using QEMU.
The build/test time for aarch64 goes from 8 minutes (testing on 2 python versions) to 3 minutes (testing all python versions).

This also fixes testing wheels being built in CI:
https://github.com/giampaolo/psutil/pull/2447  changed the way tests were run and we can see psutil gets rebuilt before testing rather than using the just built wheel in the logs: https://github.com/giampaolo/psutil/actions/runs/13076760479/job/36490726448#step:6:655)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,8 @@ test-coverage:  ## Run test coverage.
 	$(PYTHON) -m coverage html
 	$(PYTHON) -m webbrowser -t htmlcov/index.html
 
-test-ci: install-sysdeps
+test-ci:
+	${MAKE} install-sysdeps
 	mkdir -p .tests
 	cd .tests/ && $(PYTHON) -c "from psutil.tests import print_sysinfo; print_sysinfo()"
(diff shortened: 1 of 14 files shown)


mayeut:
The newly skipped tests were already mentioned in https://github.com/giampaolo/psutil/pull/2425

Makefile
```diff
@@ -162,6 +162,12 @@ test-coverage:  ## Run test coverage.
 	$(PYTHON) -m coverage html
 	$(PYTHON) -m webbrowser -t htmlcov/index.html
 
+test-ci: install-sysdeps
```
giampaolo, line 165: For consistency with the rest of the file, could you please rephrase it as:

```Makefile
test-ci: 
     {MAKE} install-sysdeps
```

giampaolo (commented):
Hi. Thanks.
If no longer used, can we remove `psutil.tests.QEMU_USER`?

Makefile
mayeut: done

mayeut:
> If no longer used, can we remove psutil.tests.QEMU_USER?

Done. It could be used downstream but not here anymore.

PR #2505, opened 2025-02-07
[OSX] proc.c: Fix goo.gl link in comment for source reference

korli:
## Summary

* OS: any
* Bug fix: yes
* Type: doc
* Fixes: 

## Description

goo.gl is going away: https://developers.googleblog.com/en/google-url-shortener-links-will-no-longer-be-available/

Original redirect points to vs/darwinsource/tarballs/other/lsof-28.tar.gz


korli:
> Thanks. I grepped for `goo.gl` and there are other occurrences. Could you change all?

All, but the one in psutil/tests/test_system.py

Issue #1119, opened 2017-08-30
Extend Process.open_files with other FDs


hzpc-joostk:
This is a proposal to query all file descriptors separately.

For now, this only works for Linux.

## code

```python
# module: psutil._pslinux

# ...

class Process(object):
    """Linux process implementation."""

    # ...

    def get_fdinfo(self, fd):
        # Get file position and flags.
        file = "%s/%s/fdinfo/%s" % (
            self._procfs_path, self.pid, fd)
        with open_binary(file) as f:
            pos = int(f.readline().split()[1])
            flags = int(f.readline().split()[1], 8)
        mode = file_flags_to_mode(flags)
        return pos, flags, mode

    @wrap_exceptions
    def open_fds(self, kind=None):
        assert kind in {'file', 'special', 'pipe', None}, kind

        retlist = []
        files = os.listdir("%s/%s/fd" % (self._procfs_path, self.pid))
        hit_enoent = False

        for fd in files:
            file = "%s/%s/fd/%s" % (self._procfs_path, self.pid, fd)
            try:
                path = readlink(file)
            except OSError as err:
                # ENOENT == file which is gone in the meantime
                if err.errno in (errno.ENOENT, errno.ESRCH):
                    hit_enoent = True
                    continue
                elif err.errno == errno.EINVAL:
                    # not a link
                    continue
                else:
                    raise
            else:
                # If path is not an absolute there's no way to tell
                # whether it's a regular file or not, so we skip it.
                # A regular file is always supposed to be have an
                # absolute path though.
                if path.startswith('/'):
                    if isfile_strict(path):
                        if (kind is not None and kind != 'file'):
                            continue
                    else:
                        if (kind is not None and kind != 'special'):
                            continue
                elif path.startswith('pipe'):
                    if (kind is not None and kind != 'pipe'):
                        continue
                    path = int(path[6:-1])
                
                else:
                    continue

                pos, flags, mode = self.get_fdinfo(fd)
                ntuple = popenfile(path, int(fd), pos, mode, flags)
                retlist.append(ntuple)

        if hit_enoent:
            # raise NSP if the process disappeared on us
            os.stat('%s/%s' % (self._procfs_path, self.pid))

        return retlist

    def open_files(self):
        return self.open_fds(kind='file')

    def open_pipes(self):
        return self.open_fds(kind='pipe')

    def open_specials(self):
        return self.open_fds(kind='special')
```

## benchmark

The following `cat | python` pipeline sleeps in background.

[...]

PR #2524, opened 2025-03-18
Add support for getting *all* files a process has open

### LINKED ISSUE #2523: Feature request: add ability to query for all files a process has open, not just the regular files
## Summary

* OS: All
* Type: core

## Description

Currently, `open_files` filters to return only regular files. I'd like to look at the TTYs that a process has open, so I'd like to look at non-regular files as well.

I propose that we add a `only_regular` `bool` parameter to `open_files` that defaults to `True` (the old behavior). If folks want to look at all files regardless of type, they can invoke `open_files(only_regular=False)` to get all the files.

This looks like a pretty simple change to make, I'll send in a PR.

jfly:
## Summary

* OS: All
* Bug fix: no
* Type: core
* Fixes: [{ comma-separated list of issues fixed by this PR, if any }](https://github.com/giampaolo/psutil/issues/2523)

## Description

Add support for getting *all* files a process has open.

To preserve the old behavior, I've changed the `open_files` to take an optional `only_regular` bool that defaults to `True` (the old behavior). If people want to look at all files, they can set that to `False`.

### DIFF
--- a/psutil/_psaix.py
+++ b/psutil/_psaix.py
@@ -511,9 +511,11 @@ class Process:
         # XXX is '?' legit? (we're not supposed to return it anyway)
         return PROC_STATUSES.get(code, '?')
 
-    def open_files(self):
+    def open_files(self, only_regular=True):
         # TODO rewrite without using procfiles (stat /proc/pid/fd/* and then
         # find matching name of the inode)
+        # <<< TODO >>>: does the current implementation ensure we only return
+        #               regular files? or is that missing?
         p = subprocess.Popen(
             ["/usr/bin/procfiles", "-n", str(self.pid)],
             stdout=subprocess.PIPE,
--- a/psutil/_psbsd.py
+++ b/psutil/_psbsd.py
@@ -894,8 +894,10 @@ class Process:
     if HAS_PROC_OPEN_FILES:
 
         @wrap_exceptions
-        def open_files(self):
+        def open_files(self, only_regular=True):
             """Return files opened by process as a list of namedtuples."""
+            # <<< TODO >>>: the regular file filtering is done in the
+            # underlying c code: `psutil/arch/bsd/proc.c`
             rawlist = cext.proc_open_files(self.pid)
             return [_common.popenfile(path, fd) for path, fd in rawlist]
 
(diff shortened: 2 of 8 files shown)


psutil/_psaix.py
```diff
@@ -511,9 +511,11 @@ def status(self):
         # XXX is '?' legit? (we're not supposed to return it anyway)
         return PROC_STATUSES.get(code, '?')
 
-    def open_files(self):
+    def open_files(self, only_regular=True):
         # TODO rewrite without using procfiles (stat /proc/pid/fd/* and then
         # find matching name of the inode)
+        # <<< TODO >>>: does the current implementation ensure we only return
+        #               regular files? or is that missing?
```
jfly, line 518: I could use some help here. Is it possible the existing implementation returns non-regular files already? Reading through the docs of [`procfiles`](https://www.ibm.com/docs/sk/aix/7.2?topic=p-procfiles-command), it mentions non-regular files:

> Regular files have permission based on mode it was opened with. Any non-regular files have 0 access mode.

So possibly we just weren't doing the filtering here and the fix is to add the missing filtering if `only_regular == True`?

psutil/_psbsd.py
```diff
@@ -894,8 +894,10 @@ def _not_implemented(self):
     if HAS_PROC_OPEN_FILES:
 
         @wrap_exceptions
-        def open_files(self):
+        def open_files(self, only_regular=True):
             """Return files opened by process as a list of namedtuples."""
+            # <<< TODO >>>: the regular file filtering is done in the
+            # underlying c code: `psutil/arch/bsd/proc.c`
```
jfly, line 900: I am pretty sure I can make this change. I just don't want to take the time to do it until I've gotten some indication from a maintainer that y'all would be willing to merge this PR.

giampaolo:
This was a topic discussed in the past (e.g. see: https://github.com/giampaolo/psutil/issues/1119#issuecomment-326812591, but I remember there were others). I ended up rejecting the proposal because providing a sane API to return all possible FD types, on all UNIX systems is very hard, because there's a ton of them. Just look at `lsof` output to get an idea. It's much more complex than just the changes you have introduced here.

PR #2527, opened 2025-03-26
Ensure consistent `Process._ident` value on Linux

### LINKED ISSUE #2526: [linux] `Process.create_time()` changes during execution, causing mayhem.
## Summary

* OS: Ubuntu Linux 24.04
* Architecture: 64bit, ARM64
* Psutil version: 7.0.0
* Python version: 3.12
* Type: core

## Description

I've identified a significant bug when running on Linux, related to #2521, that causes the process creation time reported by `Process.create_time()` for any given process to change during execution. Because the creation time is calculated relative to the current system time, if the system time is corrected during execution (for instance due to an NTP time update, or manually), the reported creation time changes under some conditions.

Since `Process` uses the values from `create_time()` along with the PID to uniquely identify a process, this leads to very unexpected results for many methods. For instance, `Process().is_running()` can return False for the currently running process (see example below). Or where `proc.is_running()` returns False for an external process that is very much alive, but `proc.wait()` still blocks indefinitely.

## Expected behavior

Intuitively, I would expect `Process.create_time()` to return a constant value, indicating the system timestamp as recorded when the process actually started, which doesn't change if the system clock is subsequently adjusted.

## Reproducible example

This scenario is tricky to recreate, but the script below is consistently showing this error on my machine. This example attempts to change the system time during execution, so it requires `sudo` access. This can have other consequences on your system, so it might be better run in a VM or container.

```python
from datetime import datetime, timezone
import time

from psutil import boot_time, Process

print(f"Current time: {datetime.now(tz=timezone.utc)}")

boot_time_before = boot_time()
print(f"Boot time: {datetime.fromtimestamp(boot_time_before, tz=timezone.utc)}")

this_proc = Process()
create_time_before = Process().create_time()
print(f"Create time: {datetime.fromtimestamp(create_time_before, tz=timezone.utc)}")
print()

try:
    print('Changing system time by +1 hour')
    real_time = time.clock_gettime(time.CLOCK_REALTIME)
    time.clock_settime(time.CLOCK_REALTIME, real_time + 3600)
    print(f"Current time: {datetime.now(tz=timezone.utc)}")

    cached_create_time = this_proc.create_time()
    new_create_time = Process().create_time()
    print(f"Create time "
          f"cached={datetime.fromtimestamp(cached_create_time, tz=timezone.utc)}, "
          f"new={datetime.fromtimestamp(new_create_time, tz=timezone.utc)}")

    print(f"Is running: {this_proc.is_running()}")
    print()

    print('Calling psutil.boot_time()')
    boot_time_after = boot_time()
    print(f"Boot time: {datetime.fromtimestamp(boot_time_after, tz=timezone.utc)}")

    cached_create_time_after = this_proc.create_time()
    new_create_time_after = Process().create_time()
    print(f"Create time: "
          f"cached={datetime.fromtimestamp(cached_create_time_after, tz=timezone.utc)}, "
          f"new={datetime.fromtimestamp(new_create_time_after, tz=timezone.utc)}")

    print(f"Is running: {this_proc.is_running()}")
    print(f"this_proc == Process(): {this_proc == Process()}")
finally:
    time.clock_settime(time.CLOCK_REALTIME, real_time)
```

## Results

The output of the above script is reproduced below:

```
Current time: 2025-03-21 01:18:56.465400+00:00
Boot time: 2025-03-06 18:06:38+00:00
Create time: 2025-03-21 01:18:55.630000+00:00

Changing system time by +1 hour
Current time: 2025-03-21 02:18:56.465695+00:00
Create time cached=2025-03-21 01:18:55.630000+00:00, new=2025-03-21 01:18:55.630000+00:00
Is running: True

Calling psutil.boot_time()
Boot time: 2025-03-06 19:06:38+00:00
Create time: cached=2025-03-21 01:18:55.630000+00:00, new=2025-03-21 02:18:55.630000+00:00
Is running: False
this_proc == Process(): False

```

## Cause

I've traced the cause as best I can, but there is not a good workaround to avoid the issue.

On linux, `Process.create_time` calculates the process creation time by adding the unix `btime` timestamp from `/proc/stat` to the process starttime from `/proc/<pid>/stat`, which records the system counter time elapsed since startup.

The `btime` value in `/proc/stat` also changes when the system time is updated, and appears to report the current unix timestamp minus the system uptime, rather than the actual system HW time at startup.

However, the `btime` value is cached in `psutil._pslinux.BOOT_TIME` every time `psutil._pslinux.boot_time()` is called, and this cached value is used to calculate `Process.create_time()`, which is also cached. So if the system time changes and the cached values are recalculated (due to explicitly calling `psutil.boot_time()` or creating a new `Process()` instance), different values are returned for `boot_time()` or `create_time()`.

## Partial solution

Because process PID and create_time are used to uniquely identify each running process, it is essential that this identifier remains unique and constant for any given process under any circumstances. Because the linux `btime` value is not actually a constant, it would seem more reliable to use the monotonic counter `starttime` recorded in `/proc/<pid>/stat` together with the PID to uniquely identify each process in `Process._ident`, since the `starttime` value does not change.

kohlerjl:
## Summary

* OS: Linux, Ubuntu 24.04
* Bug fix: yes
* Type: core
* Fixes: #2526 (partial)

## Description

This PR partially fixes the issues described in #2526, by:
* Not caching `boot_time` value in `_pslinux`, so `psutil.boot_time()` always provides a consistent (though not constant) value.
* Using monotonic process 'starttime' parsed from '/proc/<pid>/stat' to form `Process._ident` on linux to ensure a consistent and unique process ID.

The values reported by `Process.create_time()` and `boot_time()` will now immediately change to reflect changes in the system time, but the monotonic starttime used in `Process._ident` will remain unique and consistent.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1563,6 +1563,7 @@ def boot_time():
         for line in f:
             if line.startswith(b'btime'):
                 ret = float(line.strip().split()[1])
+                BOOT_TIME = ret
                 return ret
         msg = f"line 'btime' not found in {path}"
         raise RuntimeError(msg)
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -81,6 +81,7 @@ HAS_CPU_AFFINITY = hasattr(cext, "proc_cpu_affinity_get")
 # Number of clock ticks per second
 CLOCK_TICKS = os.sysconf("SC_CLK_TCK")
 PAGESIZE = cext_posix.getpagesize()
+BOOT_TIME = None  # set later
 LITTLE_ENDIAN = sys.byteorder == 'little'
 UNSET = object()
 
@@ -1556,6 +1557,7 @@ def users():
 
 def boot_time():
     """Return the system boot time expressed in seconds since the epoch."""
+    global BOOT_TIME
     path = f"{get_procfs_path()}/stat"
     with open_binary(path) as f:
         for line in f:
@@ -1894,7 +1896,7 @@ class Process:
         # We first divide it for clock ticks and then add uptime returning
         # seconds since the epoch.
         # Also use cached value if available.
-        bt = boot_time()
+        bt = BOOT_TIME or boot_time()
         return (ctime / CLOCK_TICKS) + bt
 
     @wrap_exceptions
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -81,7 +81,6 @@ HAS_CPU_AFFINITY = hasattr(cext, "proc_cpu_affinity_get")
 # Number of clock ticks per second
 CLOCK_TICKS = os.sysconf("SC_CLK_TCK")
 PAGESIZE = cext_posix.getpagesize()
-BOOT_TIME = None  # set later
 LITTLE_ENDIAN = sys.byteorder == 'little'
 UNSET = object()
 
@@ -1557,13 +1556,11 @@ def users():
 
 def boot_time():
     """Return the system boot time expressed in seconds since the epoch."""
-    global BOOT_TIME
     path = f"{get_procfs_path()}/stat"
     with open_binary(path) as f:
         for line in f:
             if line.startswith(b'btime'):
                 ret = float(line.strip().split()[1])
-                BOOT_TIME = ret
                 return ret
         msg = f"line 'btime' not found in {path}"
         raise RuntimeError(msg)
@@ -1897,7 +1894,7 @@ class Process:
         # We first divide it for clock ticks and then add uptime returning
         # seconds since the epoch.
         # Also use cached value if available.
-        bt = BOOT_TIME or boot_time()
+        bt = boot_time()
         return (ctime / CLOCK_TICKS) + bt
 
     @wrap_exceptions
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1898,6 +1898,11 @@ class Process:
         return (ctime / CLOCK_TICKS) + bt
 
     @wrap_exceptions
+    def create_monotonic(self):
+        ctime = float(self._parse_stat_file()['create_time'])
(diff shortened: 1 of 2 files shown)


giampaolo:
Hi. Thanks for the detailed research you did on this one.
Why do you say this is just a partial solution? What other problem(s) do we still have?

kohlerjl:
This fix should make sure the `Process._ident` only is constant and reproducible. There still is some ambiguity, to my mind, about what `boot_time()` and `create_time()` mean, and the fact that they can change on Linux. 

For my application, I want to store information to uniquely identify the process at a later time, but since the `Process._ident` value is not made available in the stable API, there currently is not away to get at a guaranteed consistent process start time from `psutil`.

psutil/_pslinux.py
```diff
@@ -1897,9 +1894,14 @@ def create_time(self):
         # We first divide it for clock ticks and then add uptime returning
         # seconds since the epoch.
         # Also use cached value if available.
-        bt = BOOT_TIME or boot_time()
+        bt = boot_time()
         return (ctime / CLOCK_TICKS) + bt
 
+    @wrap_exceptions
+    def create_monotonic(self):
```
giampaolo, line 1901: Please rename this to   'create_time_monotonic`.

giampaolo: Also I think you can refactor this a bit and have `create_time` call `create_time_monotonic`.

giampaolo:
I edited and removed the BOOT_TIME change. Let's discuss that in another PR / issue.

giampaolo:
Note: this created a difference in that `Process._creation_time` is no longer cached after `Process()` is instantiated for the first time. It will be cached on the next explicit call  to ``process_instance.create_time()`. I will try to address this in a separate commit.

giampaolo:
Removal of `boot_time()` cache as it was implemented in here landed into https://github.com/giampaolo/psutil/pull/2544.

PR #2535, opened 2025-04-03
fix doc string

kuzmich321:
## Summary

just fix doc string


giampaolo:
Good catch. There is also one occurrence in `docs/index.rst`. Could you change that as well?

PR #2549, opened 2025-04-08
"unix" should be added to conn_tmap on illumos/SunOS too

### LINKED ISSUE #2547: [illumos] `TestConnectedSocket.test_unix` fails
## Summary

* OS: OpenIndiana (illumos)
* Architecture: 64bit
* Psutil version: 7.0.0
* Python version: Python 3.9.21
* Type: core

## Description

The following test fails on OpenIndiana:
```
________________________ TestConnectedSocket.test_unix _________________________

self = <psutil.tests.test_connections.TestConnectedSocket testMethod=test_unix>

    @pytest.mark.skipif(not POSIX, reason="POSIX only")
    def test_unix(self):
        testfn = self.get_testfn()
        server, client = unix_socketpair(testfn)
        try:
>           cons = this_proc_net_connections(kind='unix')

psutil/tests/test_connections.py:232:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
psutil/tests/test_connections.py:50: in this_proc_net_connections
    cons = psutil.Process().net_connections(kind=kind)
psutil/__init__.py:1234: in net_connections
    _check_conn_kind(kind)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

kind = 'unix'

    def _check_conn_kind(kind):
        """Check net_connections()'s `kind` parameter."""
        kinds = tuple(_common.conn_tmap)
        if kind not in kinds:
            msg = f"invalid kind argument {kind!r}; valid ones are: {kinds}"
>           raise ValueError(msg)
E           ValueError: invalid kind argument 'unix'; valid ones are: ('all', 'tcp', 'tcp4', 'udp', 'udp4', 'inet', 'inet4', 'inet6', 'tcp6', 'udp6')

psutil/__init__.py:280: ValueError
```
When I revert (partially) changeset 5d4be42bd309dc3c7e93d5d319577ec4b8027445 using this:
```
--- psutil-7.0.0/psutil/_common.py.orig
+++ psutil-7.0.0/psutil/_common.py
@@ -249,7 +249,7 @@
         "udp6": ([AF_INET6], [SOCK_DGRAM]),
     })

-if AF_UNIX is not None and not SUNOS:
+if AF_UNIX is not None:
     conn_tmap.update({"unix": ([AF_UNIX], [SOCK_STREAM, SOCK_DGRAM])})


```
Then the test passes.

mtelka:
## Summary

* OS: OpenIndiana (illumos)
* Bug fix: yes
* Type: core
* Fixes: #2547

## Description

Commit 5d4be42bd309dc3c7e93d5d319577ec4b8027445 conditionally removed the addition of "unix" to `conn_tmap` for SUNOS.  This seems to be wrong and is causing four tests to fail (see #2547 for details).  This change reverts the offending part of 5d4be42bd309dc3c7e93d5d319577ec4b8027445 to avoid test failures documented in #2547.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -22,6 +22,7 @@ XXXX-XX-XX
 - 2542_: if system clock is updated `Process.children()`_ and
   `Process.parent()`_ may not be able to return the right information.
 - 2545_: [illumos]: Fix handling of MIB2_UDP_ENTRY in `net_connections()`_.
+- 2547_: [illumos]: Add "unix" to conn_tmap on illumos/SunOS too.
 
 7.0.0
 =====
--- a/psutil/_common.py
+++ b/psutil/_common.py
@@ -249,7 +249,7 @@ if AF_INET6 is not None:
         "udp6": ([AF_INET6], [SOCK_DGRAM]),
     })
 
-if AF_UNIX is not None and not SUNOS:
+if AF_UNIX is not None:
     conn_tmap.update({"unix": ([AF_UNIX], [SOCK_STREAM, SOCK_DGRAM])})
 
 


HISTORY.rst
```diff
@@ -22,6 +22,7 @@ XXXX-XX-XX
 - 2542_: if system clock is updated `Process.children()`_ and
   `Process.parent()`_ may not be able to return the right information.
 - 2545_: [illumos]: Fix handling of MIB2_UDP_ENTRY in `net_connections()`_.
+- 2547_: [illumos]: Add "unix" to conn_tmap on illumos/SunOS too.
```
giampaolo, line 25: you can remove this; not worth mentioning

giampaolo, line 25: What's the change exactly? Was only the unit test broken or also the function per se? If the former, please explain what was the problem (exception? crash?).

giampaolo (changes requested):

HISTORY.rst
mtelka: The change exactly is to add "unix" back to `conn_tmap` as it was before December 2024.

I found the issue by running `psutil`'s tests.  I do not use `psutil` for anything specific.  I just package it for OpenIndiana (and so I run its tests to make sure it works), primarily because it is needed by some other Python projects packaged.

giampaolo: It looks like UNIX sockets are not supported on SunOS, that's why I removed it:
https://github.com/giampaolo/psutil/blob/11fd4026d55b978b95ac6d15398b85c6c0c52bec/psutil/_pssunos.py#L262

PR #2550, opened 2025-04-09
Fix tests for riscv64

kxxt:
## Summary

* OS: Arch Linux RISC-V
* Bug fix: yes
* Type: tests
* Fixes:  test failure on riscv64

## Description

Some tests are failing on riscv64 because e.g. CPU frequency is not available in /proc/cpuinfo.

Those tests are already marked for skip on AArch64. This patch skips them for riscv64 as well.


kxxt:
> I pushed a more generic solution: [8d8961f](https://github.com/giampaolo/psutil/commit/8d8961f033c3198a6d893753a9c9ebe60e3f0798). Can you check if it works?

Thanks for the quick fix. I have tested it on riscv64 and it works.

PR #2560, opened 2025-04-15
linux: fix smaps parsing crash on empty VmFlags: lines

justeph:
## Summary

* OS: ubuntu
* Bug fix: yes
* Type: core

## Description
On some systems (observed at least on riscv64 under qemu emulation), /proc/<pid>/smaps contains lines like:

    VmFlags:

without any value. The current parser assumed at least two fields per line, causing an IndexError during memory_maps() processing.

This patch defensively skips lines with less than 2 fields, which matches existing logic for ignoring unhandled VmFlags lines.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -2021,7 +2021,7 @@ class Process:
                     else:
                         try:
                             data[fields[0]] = int(fields[1]) * 1024
-                        except ValueError:
+                        except (ValueError, IndexError):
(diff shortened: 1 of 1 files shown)


justeph:
Hi @giampaolo not sure the 2 failing test are related to my PR.. Also [bsd-tests / freebsd (pull_request)](https://github.com/giampaolo/psutil/actions/runs/14470422917/job/40582656355?pr=2560) is still running with `VM is booting`

PR #2577, opened 2025-05-28
Adopt multi-phase init (PEP 489)

AA-Turner:
cc @giampaolo 

xref #2576 

The mechanical changes for each module are:

1. Change `PyMODINIT_FUNC PyInit_{name}(void)` to `static int {name}_exec(PyObject *mod)`
2. Change `return NULL` to `-1` and `return mod` to `0`
3. Remove `PyModule_Create` and `PyUnstable_Module_SetGIL`
4. Add a PyModuleDef_Slot slots array containing the exec function & slots to declare subinterpreter and no-GIL support
5. Move the PyModuleDef struct after the slots array & add the new `.m_slots` member
6. Restore the `PyMODINIT_FUNC PyInit_{name}(void)` function, now just with the `PyModuleDef_Init()` call

A

### DIFF
--- a/psutil/_psutil_linux.c
+++ b/psutil/_psutil_linux.c
@@ -67,10 +67,10 @@ _psutil_linux_exec(PyObject *mod) {
 
 static struct PyModuleDef_Slot _psutil_linux_slots[] = {
     {Py_mod_exec, _psutil_linux_exec},
-#if PY_VERSION_HEX >= 0x030c00f0  // Python 3.12+
+#ifdef Py_mod_multiple_interpreters  // Python 3.12+
     {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
 #endif
-#if PY_VERSION_HEX >= 0x030d00f0  // Python 3.13+
+#ifdef Py_mod_gil  // Python 3.13+
     // signal that this module supports running without an active GIL
     {Py_mod_gil, Py_MOD_GIL_NOT_USED},
 #endif
--- a/psutil/_psutil_linux.c
+++ b/psutil/_psutil_linux.c
@@ -48,8 +48,23 @@ static PyMethodDef mod_methods[] = {
 };
 
 
+static int module_loaded = 0;
+
 static int
 _psutil_linux_exec(PyObject *mod) {
+    // https://docs.python.org/3/howto/isolating-extensions.html#opt-out-limiting-to-one-module-object-per-process
+    if (module_loaded) {
+        PyErr_SetString(PyExc_ImportError,
+                        "cannot load module more than once per process");
+        return -1;
+    }
+    module_loaded = 1;
+
+#ifdef Py_GIL_DISABLED
+    if (PyUnstable_Module_SetGIL(mod, Py_MOD_GIL_NOT_USED))
+        return NULL;
+#endif
+
     if (psutil_setup() != 0)
         return -1;
 
@@ -67,13 +82,6 @@ _psutil_linux_exec(PyObject *mod) {
 
 static struct PyModuleDef_Slot _psutil_linux_slots[] = {
     {Py_mod_exec, _psutil_linux_exec},
-#ifdef Py_mod_multiple_interpreters  // Python 3.12+
-    {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
-#endif
-#ifdef Py_mod_gil  // Python 3.13+
-    // signal that this module supports running without an active GIL
-    {Py_mod_gil, Py_MOD_GIL_NOT_USED},
-#endif
     {0, NULL},
 };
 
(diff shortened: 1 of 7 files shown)


psutil/_psutil_windows.c
```diff
@@ -133,127 +133,142 @@ psutil_windows_clear(PyObject *m) {
...
-#endif
+static int
+_psutil_windows_exec(PyObject *mod) {
+    // https://docs.python.org/3/howto/isolating-extensions.html#opt-out-limiting-to-one-module-object-per-process
+    if (module_loaded) {
+        PyErr_SetString(PyExc_ImportError,
+                        "cannot load module more than once per process");
+        return -1;
+    }
+    module_loaded = 1;
```
AA-Turner, line 146: The Windows extension is the only one to use global state, via the `TimeoutExpired` and `TimeoutAbandoned` variables. Here we block reinitialisation to avoid UB/crashes.

giampaolo:
There's errors like:

```
DPSUTIL_SIZEOF_PID_T=4 -DPSUTIL_VERSION=701 -DPSUTIL_FREEBSD=1 -I/usr/local/include/python3.11 -c psutil/_psutil_bsd.c -o /tmp/tmpt5wq7zks.build-temp/psutil/_psutil_bsd.o
2025-05-28T13:38:48.0092731Z       psutil/_psutil_bsd.c:226:30: error: use of undeclared identifier 'moduledef'; did you mean 'module_def'?
2025-05-28T13:38:48.0093197Z         226 |     return PyModuleDef_Init(&moduledef);
2025-05-28T13:38:48.0093482Z             |                              ^~~~~~~~~
2025-05-28T13:38:48.0093727Z             |                              module_def
2025-05-28T13:38:48.0094421Z       psutil/_psutil_bsd.c:216:27: note: 'module_def' declared here
2025-05-28T13:38:48.0094769Z         216 | static struct PyModuleDef module_def = {
2025-05-28T13:38:48.0095021Z             |                           ^
2025-05-28T13:38:48.0095249Z       1 error generated.
2025-05-28T13:38:48.0095492Z       error: command '/usr/bin/cc' failed with exit code 1
2025-05-28T13:38:48.0095761Z       [end of output]
```

psutil/_psutil_linux.c
```diff
@@ -48,41 +48,44 @@ static PyMethodDef mod_methods[] = {
...
-PyObject *
-PyInit__psutil_linux(void) {
-    PyObject *mod = PyModule_Create(&moduledef);
-    if (mod == NULL)
-        return NULL;
-
-#ifdef Py_GIL_DISABLED
-    if (PyUnstable_Module_SetGIL(mod, Py_MOD_GIL_NOT_USED))
-        return NULL;
-#endif
```
giampaolo, line 73: This was added by @colesbury in https://github.com/giampaolo/psutil/commit/73af40731. Is it OK to delete it?

colesbury: Yeah, it's fine to remove. The `{Py_mod_gil, Py_MOD_GIL_NOT_USED},` in the slots is the equivalent for multi-phase init.

giampaolo: Thank you. ;)

psutil/_psutil_posix.c
```diff
@@ -1022,11 +1001,36 @@ PyInit__psutil_posix(void) {
...
 #endif  // defined(PSUTIL_LINUX) || defined(PSUTIL_FREEBSD)
 
-    return mod;
+    return 0;
+}
+
+static struct PyModuleDef_Slot _psutil_posix_slots[] = {
+    {Py_mod_exec, _psutil_posix_exec},
+#ifdef Py_mod_multiple_interpreters  // Python 3.12+
+    {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
```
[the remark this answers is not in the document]
AA-Turner: @colesbury Indeed, I've thinking about this but haven't come up with any good solutions. The 'simplest' is for *psutil* to distribute a wheel built with the limited API set to 3.12+, which would double the number of wheels in the medium term, but still be O(1).

One potential solution would be to define two slots arrays, `slots_py36` and `slots_py312`. We could then choose between them based on the runtime version (though [`Py_Version`](https://docs.python.org/dev/c-api/apiabiversion.html#run-time-version) was only added in 3.11, so this might have to go via `sys.hexversion`). This would still need to hard-code the symbols that don't exist in Python 3.6, but avoids dynamic manipulation of the arrays. 

cc @encukou for any advice you may have here!

A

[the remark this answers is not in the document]
encukou: @AA-Turner, the suggestions should fix those.

giampaolo:
Closing this out in favor of https://github.com/giampaolo/psutil/pull/2855.

giampaolo:
Implemented in https://github.com/giampaolo/psutil/pull/2855/.

giampaolo:
Also see https://github.com/giampaolo/psutil/issues/2881.

PR #2583, opened 2025-05-31
gh-1946: Add PEP 484 type annotations

### LINKED ISSUE #1946: Type hints & PEP 561 packaging

Harry-Lees:
## Summary

- OS: MacOS Sequoia (15.3.2)
- Bug fix: no
- Type: core
- Fixes: #1946

## Description

This PR adds support for [PEP 484](https://peps.python.org/pep-0484/) type annotations. Currently, type annotations for this library are maintained in the 3rd party typeshed project.

There has been interest shown in implementing type annotations natively, IMO one major blocker for implementing this was the support for Python 2.6+, which prevented use of any PEP 484 syntax, instead requiring type comments. Since this project dropped support for Python 2.x versions, it is possible to maintain the type annotations inline in the project.

### Current Implementation

It has been several years since #1946 was first opened, so I'm aware that opinions may have changed about the usefulness of this feature. I've currently implemented a "first pass", implementing most of the public API as proposed in the original issue [here](https://github.com/giampaolo/psutil/issues/1946#issuecomment-839108820)

> Also, the whole public API is defined in this file:
https://github.com/giampaolo/psutil/blob/master/psutil/__init__.py
...so I wouldn’t bother to support / modify other files other than this one.

There is also some more complex behaviour for some methods which I haven't implemented. If discussion around this PR is generally positive towards being helpful for the project, I'm happy to go back and implement some of the more complex behaviour. Some examples of this can be seen in the pre-existing typeshed implementation [e.g.](https://github.com/python/typeshed/blob/9b7b2bafab7599195f400e1bf9f31620321aeb2e/stubs/psutil/psutil/__init__.pyi#L254-L259)

```python
 @overload
def cpu_percent(interval: float | None = None, percpu: Literal[False] = False) -> float: ...
@overload
def cpu_percent(interval: float | None, percpu: Literal[True]) -> list[float]: ...
@overload
def cpu_percent(*, percpu: Literal[True]) -> list[float]: ...
```

### Alternative Approaches

There was some discussion in the linked issue (#1946) surrounding alternative approaches. One of these is supplying type annotations in `.pyi` files as with typeshed. The TLDR of discussions in the issue is that

1. Type annotations in `.pyi` files avoid cluttering the main codebase with type annotations (this PR itself introduces a significant number of imports, and `type: ignore` comments to keep existing behaviour.
2. Type annotations in the main codebase remove the possibility of having out-of-sync annotations and code.

I'm personally quite neutral about this, leaning towards inclusion in the main codebase, hence this PR :D, but I would be interested in alternative approaches as-well.

### Notes

If this PR was to be considered, I think it would also be useful to add MyPy to the CI to ensure the annotations are kept up-to-date and valid.

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -29,11 +29,24 @@ import subprocess
 import sys
 import threading
 import time
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import FrozenSet
+from typing import Generator
+from typing import Iterable
+from typing import Iterator
+from typing import List
+from typing import OrderedDict
+from typing import Set
+from typing import Tuple
+from typing import Union
 
 try:
     import pwd
 except ImportError:
-    pwd = None
+    pwd = None  # type: ignore[assignment]
 
 from . import _common
 from ._common import AIX
@@ -97,7 +110,7 @@ if LINUX:
     from ._pslinux import IOPRIO_CLASS_RT  # noqa: F401
 
 elif WINDOWS:
-    from . import _pswindows as _psplatform
+    from . import _pswindows as _psplatform  # type: ignore[no-redef]
     from ._psutil_windows import ABOVE_NORMAL_PRIORITY_CLASS  # noqa: F401
     from ._psutil_windows import BELOW_NORMAL_PRIORITY_CLASS  # noqa: F401
     from ._psutil_windows import HIGH_PRIORITY_CLASS  # noqa: F401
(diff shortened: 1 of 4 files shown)


giampaolo:
Hi. I must admit I am not a fan of type annotations. I would prefer to have them in a separate .pyi file.

giampaolo:
I ended up doing this in https://github.com/giampaolo/psutil/pull/2756

Issue #2565, opened 2025-04-21
Add wheels for the free-threaded CPython build & test on CI


lysnikolaou:
## Summary

* OS: All OS's
* Type: core, wheels

## Description

Currently, `psutil` only offers wheels for the stable ABI. Ideally, we should also provide wheels for the free-threaded CPython build and also test it in CI. 

There's also some global state in various places that needs to be locked in case it's intended to be used from multiple threads.

rgommers:
Note that this was already proposed in gh-2491, with a lot of relevant discussion. The key comment is https://github.com/giampaolo/psutil/pull/2491#issuecomment-2591265995 (wait for `abi4`). That decision seems a little unfortunate (but it's early days); `psutil` is quite heavily used, so for CPython 3.14 in a few months it seems like `cp314t` wheels will help a lot of users. `abi4` won't arrive until Python 3.15. Our team will be happy to contribute the extra maintenance effort that may be required to build and support `cp314t` wheels.

colesbury:
> There's also some global state in various places that needs to be locked ...

What global state is there?

giampaolo:
For things like that (`psutil_xfiles`) feel free to open individual PRs (actually it's better).

giampaolo:
I reverted https://github.com/giampaolo/psutil/pull/2590 since it was causing a segfault.

stonebig:
any hope to get 3.14 and 3.14t wheels at next release ?

giampaolo:
py314t wheels added in https://github.com/giampaolo/psutil/pull/2659.
I'll publish a release soon.

giampaolo:
And it's done. Release 7.1.2 with free-thread wheels is out!
Thanks all.

giampaolo:
Blog post. @lysnikolaou  I took the liberty to mention you:
https://gmpy.dev/blog/2025/wheels-for-free-threaded-python-now-available-in-psutil

PR #2590, opened 2025-06-24
Call GetExtended[Tcp|Udp]Table twice under free-threaded build

lysnikolaou:
## Summary

- OS: Windows
- Bug fix: yes
- Type: core

## Description

To avoid using global state for the size of the allocated tcp/udp table, we fall back to the slower path of calling the `GetExtended[Tcp|Udp]Table` APIs twice, once to get the size of the table we need to allocate, and once to actually populate the table. Related to #2565.

### DIFF
--- a/psutil/arch/windows/socks.c
+++ b/psutil/arch/windows/socks.c
@@ -18,8 +18,10 @@
 #define BYTESWAP_USHORT(x) ((((USHORT)(x) << 8) | ((USHORT)(x) >> 8)) & 0xffff)
 #define STATUS_UNSUCCESSFUL 0xC0000001
 
+#ifndef Py_GIL_DISABLED
 ULONG g_TcpTableSize = 0;
 ULONG g_UdpTableSize = 0;
+#endif
 
 
 // Note about GetExtended[Tcp|Udp]Table syscalls: due to other processes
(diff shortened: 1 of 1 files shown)


lysnikolaou:
Friendly ping for a review here as well if you ever find some time.

lysnikolaou:
@giampaolo Is there anything I can do to help push this forward?

psutil/arch/windows/socks.c
```diff
@@ -38,13 +40,18 @@ static PVOID __GetExtendedTcpTable(ULONG family) {
...
     if (size == 0) {
         GetExtendedTcpTable(NULL, &size, FALSE, family, class, 0);
         // reserve 25% more space
         size = size + (size / 2 / 2);
         g_TcpTableSize = size;
     }
+#else
+    GetExtendedTcpTable(NULL, &size, FALSE, family, class, 0);
+    size = size + (size / 2 / 2);
+#endif
```
giampaolo, line 54: I don't like the fact that here you use 2 different code paths depending on Py_GIL_DISABLED. There should be only one logic.

giampaolo (commented):
Sorry for the delay. I added 1 comment.

lysnikolaou:
> I'm sorry but I had to revert this in [ef72dcd](https://github.com/giampaolo/psutil/commit/ef72dcd076f1cec94ba50ec30b5dbebb272845d0) since it caused a segfault on my local Windows machine, and reverting it to the old code fixed it. I'm not really sure what was wrong here, but something is.

Thanks for the message here @giampaolo. Unfortunately, I don't have access to a Windows machine to debug that. Can you maybe provide more information about the segfault?

PR #2591, opened 2025-06-24
Fix thread-safety issues in getloadavg on Windows

lysnikolaou:
## Summary

- OS: Windows
- Bug fix: yes
- Type: core

## Description

Lock around reading & writing of global state in `getloadavg` on Windows.

### DIFF
--- a/psutil/arch/windows/wmi.c
+++ b/psutil/arch/windows/wmi.c
@@ -31,6 +31,15 @@ double load_avg_1m = 0;
 double load_avg_5m = 0;
 double load_avg_15m = 0;
 
+#ifdef Py_GIL_DISABLED
+static PyMutex mutex;
+#define MUTEX_LOCK(m) PyMutex_Lock(m)
+#define MUTEX_UNLOCK(m) PyMutex_Unlock(m)
+#else
+#define MUTEX_LOCK(m)
+#define MUTEX_UNLOCK(m)
+#endif
+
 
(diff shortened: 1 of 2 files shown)


psutil/arch/windows/wmi.c
```diff
@@ -31,6 +31,15 @@ double load_avg_1m = 0;
...
 double load_avg_15m = 0;
 
+#ifdef Py_GIL_DISABLED
+static PyMutex mutex;
+#define MUTEX_LOCK(m) PyMutex_Lock(m)
+#define MUTEX_UNLOCK(m) PyMutex_Unlock(m)
+#else
+#define MUTEX_LOCK(m)
+#define MUTEX_UNLOCK(m)
+#endif
```
giampaolo, line 41: OK, this is probably a matter of personal taste (and I know it's probably incorrect), but in this case I find it more clear for this to be indented:

```c
#ifdef Py_GIL_DISABLED
    static PyMutex mutex;
    #define MUTEX_LOCK(m) PyMutex_Lock(m)
    #define MUTEX_UNLOCK(m) PyMutex_Unlock(m)
#else
    #define MUTEX_LOCK(m)
    #define MUTEX_UNLOCK(m)
#endif

giampaolo (changes requested):
Looks very good and makes sense to me. I just added a minor style-related comment.

psutil/arch/windows/wmi.c
lysnikolaou: Done!

PR #2602, opened 2025-07-15
ci(setup-python): migrate LizardByte/setup-python-action

ReenigneArcher:
## Summary

- OS: ci
- Bug fix: no
- Type: ci
- Fixes: n/a

## Description

LizardByte/setup-python-action is deprecated and has moved to LizardByte/actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
-      - uses: LizardByte/setup-python-action@master
+      - uses: LizardByte/actions/actions/setup_python@master
         with:
           python-version: "2.7"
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
> https://github.com/giampaolo/psutil/pull/2602/checks?check_run_id=45976140287

I'm not going to do this. I just opened the PR as a courtesy, but I did so from GitHub's UI (which does not make it possible to edit commits).

The commit is already signed and verified though. Perhaps you should consider using branch protection rules that allow you to check if a commit is signed?
image

Anyway, I will close this. Just want you to be aware that the previous action is no longer being maintained and I have moved it to a different repository.

giampaolo:
It's OK, feel free not to sign and reopen so we can merge it.

PR #2609, opened 2025-08-06
Build wheels for cp313t

lysnikolaou:
## Summary

- OS: All
- Bug fix: no
- Type: wheels

## Description

Build wheels for cp313t. Related to #2565.


lysnikolaou:
> We currently have OSX builds which hang for no apparent reason, and then there is [#2491 (comment)](https://github.com/giampaolo/psutil/pull/2491#issuecomment-2591265995). So don't hold your breath on this one. I'll likely get to it later this year.

For what it's worth, the hangs happend on 3.13, not 3.13t, but I get what you mean about having more tests in the matrix. Also, in general, the test suite seems to be doing well under free-threading (I haven't had failures after all the free-threading-related PRs were merged, minus the Windows ones because of `pywin32`).



> I think the right solution would be to just have a pipeline which creates the free-threaded wheels, but does not run the unit tests against them.

Let's go with that if you think is best! Done in this PR.

giampaolo:
OT for this discussion but....

What I don't understand is this: I tried running a test pipeline  on OSX for 7.0.0 and that also hangs. But it didn't hang back then. So perhaps something changed on the build/github side.

lysnikolaou:
Friendly ping. Can I do anything to help push this forward?

ngoldbaum:
I think the linter failures will go away if you rebase.

ngoldbaum:
Hopefully 3.15 will fix this once and for all, see the competing [PEP 803](https://peps.python.org/pep-0803/) and [PEP 809](https://peps.python.org/pep-0809/) for the current situation.

PR #2615, opened 2025-08-08
Lock around uses of getutent/setutent/endutent

### LINKED ISSUE #2614: `psutil.users()` is not thread-safe
Related to https://github.com/giampaolo/psutil/issues/2565#issuecomment-3168110743. See file:
https://github.com/giampaolo/psutil/blob/01111d78/psutil/arch/linux/users.c
In here we use `setutent()` and `endutent()` syscalls, which appear to be "global", and hence not thread safe.

I'm not sure how to properly handle this because I'm confused by the existence of the 2 thread models: the "normal" thread-model, for which we could use `PyThread_acquire_lock` / `PyThread_release_lock` or `Py_BEGIN_ALLOW_THREADS` / `Py_END_ALLOW_THREADS` (are they the same?). And now there is the "free-thread python mode" which looks like a different beast. Perhaps @colesbury has some advice?

lysnikolaou:
## Summary

- OS: Arch Linux
- Bug fix: no
- Type: core
- Fixes: #2614 

## Description

Lock around uses of `getutent`/`setutent`/`endutent` to prevent concurrent access.


colesbury:
This LGTM for what it's worth

giampaolo:
Thanks a lot @lysnikolaou. Yes, the `users()` code is (was) basically duplicated on different platforms. I did https://github.com/giampaolo/psutil/pull/2616 to unify it. Could you please adapt your patch to the new code? The new file is now https://github.com/giampaolo/psutil/blob/064fb39d/psutil/arch/posix/users.c.

lysnikolaou:
@giampaolo Done!

lysnikolaou:
> It looks like you're using an old version of the code. Try to git pull & merge from master.

It looks okay to me now, no?

giampaolo:
Mmm no, you're still behind master. `psutil/_psutil_aix.c` file no longer has `psutil_users()`.

giampaolo:
Oh sorry, I didn't realize it was boot_time() function. I think this poses another problem though. We have multiple locks at play, one for boot_time() and one for users(), but at its core `setutxent()` is a global thingie, so it requires a global lock. I think we should define a single `PSUTIL_MUTEX`, and put it under `psutil/arch/all/init.h`, meant to be used as such (a global lock). For now it's only used by `users()` and `boot_time()`, but it could be reused later for cases similar to `setutxent()`.

lysnikolaou:
Hey @giampaolo! I've updated the PR to use one global lock. Sorry for the delay.

PR #2630, opened 2025-09-19
Exclude test files from package distribution

mbeijen:
- Remove 'psutil.tests' from packages list in setup.py
- Remove explicit test file includes from MANIFEST.in
- Add exclusion pattern for psutil/tests/* in MANIFEST.in

This ensures that test files are not included in the distributed package, reducing package size and following Python packaging best practices.

## Summary

- OS: Debian
- Bug fix: yes
- Type: packaging

## Description
I found that when installing psutil its test suite also gets installed in `site-packages`. This is not common for python modules to do. With these changes test files are no longer installed in peoples site-packages, leading to smaller package sizes and smaller install sizes.


giampaolo:
The idea of including tests is to be able to test the psutil installation (e.g. on a server). But I do agree it's not standard practice. Uhm...

mbeijen:
Yeah, and I see that tests are now failing because on some platforms the wheels are built and then the tests inside the wheels are run which are no longer included ;-) 

What should I do, just retract this PR and open up an issue instead? I don't really feel like messing with the CI...

mayeut:
If the tests are excluded from the wheels, then it would likely help to move to an [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), moving the tests to the root of the repo.
This would remove the "change cwd" hack but wouldn't allow to call other test targets as long as they call the build target unconditionally (as that would override the wheel installed).

mbeijen:
I opened #2632 for the move to the src layout, closing this PR.

PR #2632, opened 2025-09-20
switch to src layout

mbeijen:
This switches the project to use a src/ layout, keeping tests in the top level. See for more information:

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

One of the benefits of this layout is that tests are now in the top level directory. They are no longer part of the generated wheel, keeping dist size small and installations clean.

See also https://github.com/giampaolo/psutil/pull/2630

## Summary

- OS: Linux
- Bug fix: no
- Type: dev engineering

## Description

This change is a result of #2630 and my initial proposal for a fix to remove test files from the wheels. Moving to src layout provides a little more 'modern' project layout and fixes the issue of having test files in the dist.


mbeijen:
I went ahead and moved to src layout as it was suggested by @mayeut which you asked for help... https://github.com/giampaolo/psutil/pull/2630#issuecomment-3313619615

I'm sorry, I am not properly aware of who is who in this project. I'll close this PR and make yet another one where we move the tests to root and outside of psutil to make sure they're not inside the wheel.

PR #2638, opened 2025-09-30
avoid raising the result of pytest.skip/pytest.fail

elisw93:
## Summary

- OS: Linux
- Bug fix: no
- Type: tests
- Fixes: N/A

## Description

First of all, if there is some other reason for this pattern which I have failed to realize, I apologize. I assume this code was necessary at some point, and it went unnoticed in the transition to Python 3 and/or pytest.

I noticed that a few of the tests in psutil will raise the result of `pytest.skip` or `pytest.fail` as an exception when tests need to skip/fail. However, these functions (both the native functions in pytest, and the fallbacks which call unittest functions) raise their own internal exception types and don't return anything; they are marked `-> NoReturn`.

Because of this, a statement like
```python
raise pytest.skip("raises AccessDenied")
```

is really equivalent to
```python
raise None
```

which would raise a `TypeError`, if it ever executed. (Of course it never executes, because `pytest.skip` never returns.)

~~In addition to removing these `raise` statements, I also moved the most trivial `if`/`pytest.skip` checks to `@pytest.mark.skipif` decorators.~~ **Edit:** I didn't consider that the decorators will be evaluated even if they're inside a block that was marked `skip` for other reasons, so this caused platform-specific code to run on the wrong platform and broke CI pretty badly. I rolled this back in https://github.com/giampaolo/psutil/pull/2638/commits/7f9418d6984114907c5f10d2e66067aa6eb78342.

As for the check for outdated `free`, I noticed that the new calculation of `MEMINFO_MEM_USED` was introduced in v4.0.1 according to the tag list [on the commit](https://gitlab.com/procps-ng/procps/-/commit/2184e90d2e7cdb582f9a5b706b47015e56707e4d), so I updated the version check there as well.

It would be nice to fix the static analysis configuration so that this type of exception mismatch would be flagged, but I'll start with fixing the code itself.

### DIFF
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -942,10 +942,11 @@ def create_py_exe(path):
     return path
 
 
-@pytest.mark.skipif(not shutil.which("gcc"), reason="gcc is not installed")
 def create_c_exe(path, c_code=None):
     """Create a compiled C executable in the given location."""
     assert not os.path.exists(path), path
+    if not shutil.which("gcc"):
+        pytest.skip("gcc is not installed")
     if c_code is None:
         c_code = textwrap.dedent("""
             #include <unistd.h>
--- a/psutil/tests/__init__.py
+++ b/psutil/tests/__init__.py
@@ -946,7 +946,7 @@ def create_c_exe(path, c_code=None):
     """Create a compiled C executable in the given location."""
     assert not os.path.exists(path), path
     if not shutil.which("gcc"):
-        raise pytest.skip("gcc is not installed")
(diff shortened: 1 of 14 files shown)


giampaolo:
Hi. For the record, despite not really necessary per se, I have used `raise` as a visual enforcement that the execution blocks at that point. E.g. something that immediately catches the eye. But I understand your argument that since it returns None it's incorrect to raise it. I still would like to keep the "visual enforcement" though, so I would kindly ask you if you can replace `raise pytest.fail(...)` occurrences with `return pytest.fail(...)`

PR #2652, opened 2025-10-21
Ensure exit code from fake_pytest.main() is propagated

hroncok:
Or else failed tests exit with exit code 0.

## Summary

- OS: all
- Bug fix: yes
- Type: tests

## Description

I noticed failed tests did not fail the CI in Fedora. When looking into why: It was because when pytest was not installed and fake_pytest was used, `python3 -m psutil.tests` always exited with 0.

### DIFF
--- a/psutil/tests/__main__.py
+++ b/psutil/tests/__main__.py
@@ -6,6 +6,8 @@
 $ python3 -m psutil.tests.
 """
 
+import sys
+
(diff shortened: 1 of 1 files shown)


hroncok:
Apparently, this also exposes the exit code of pytest. And the CI failures are seen just because of that.

PR #2659, opened 2025-10-24
Build 3.14t wheels

lysnikolaou:
## Summary

- OS: All
- Bug fix: no
- Type: wheels

## Description

I did some testing locally and it looks like 3.14t wheels can be safely built as well now that the gil-enabled build of 3.14 is supported.


giampaolo:
Boom! Merged.

PR #2665, opened 2025-10-26
Windows getloadavg improvements. Fixes #2664

### LINKED ISSUE #2664: [Windows] `getloadavg()` improvements
A few things could be improved about `getloadavg()` on Windows:
* The 5 second interval is very prone to moire artifacts, see https://ripke.com/loadavg/moire. I recommend using (60/13).
* Queue length do not include current running count, only the waiting-count, unlike Unix load average. You gotta add processor util% to match.
* Depending if you want to match Linux or other Unix-like systems, you may want to add disk queue length.

_Originally posted by @Artoria2e5 in https://github.com/giampaolo/psutil/issues/1485#issuecomment-3448289579_

Artoria2e5:
## Summary

- OS: Windows
- Bug fix: kinda?
- Type: core
- Fixes: #2664

## Description

* Add accounting for current-running processes via CPU load. The queue length would otherwise be not comparable to Unix systems.
* Initialize at current load value instead of 0.
* Use a new sampling interval of 4.615 seconds per Ripke.

The disk queue length is still not implemented as it is not required to match most of Unix and would require another counter handle.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -41,12 +41,19 @@ ColumnLimit: 79
 DerivePointerAlignment: false
 IndentCaseBlocks: true
 IndentCaseLabels: true
+IndentPPDirectives: BeforeHash
 IndentWidth: 4
 MaxEmptyLinesToKeep: 2
 PointerAlignment: Right
 SortIncludes: false
 SpaceBeforeParens: ControlStatementsExceptControlMacros
 UseTab: Never
+
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+  - WITH_MUTEX
 
 # Force fun return type and fun definition to stay on 2 different lines:
 #    static int
@@ -60,6 +67,7 @@ AlwaysBreakAfterReturnType: TopLevelDefinitions
 #        Bar(...)
 PenaltyBreakAssignment: 400
 PenaltyBreakBeforeFirstCallParameter: 0
+
 
 # Handle macros with no `;` at EOL, so that they don't include the next line
 # into them.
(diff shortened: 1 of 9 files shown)


Artoria2e5:
Just one newline missing. Might as well let it finish the checks... Or not. The new push is gonna go through checks too.

Artoria2e5:
On second thought disk queue length is actually useful because Windows does have this tendency of freezing the entire explorer when you insert a broken drive. Alright it's just one more handle, how hard can it be.

Artoria2e5:
There's no need for initialization mutex in C because (1) it's already held by a lock the Python part (2) it finishes one initialization in the main thread before handing it off to the timer. So it's always only gonna be one thread spawned. So why am I even doing HeapAlloc or bothering to protect any part of context?

Mutex on the load and error variables is not strictly necessary for "okay" operation but might as well keep it. Bumping into it as held is going to be rare.

Artoria2e5:
Aha, there is clang-format now. Just gotta do some merging...

giampaolo:
Hi,
this PR aims at doing too much. I suggest to remove the disk queue length stuff, and for now only focus on these 2 items from #2664:

> - The 5 second interval is very prone to moire artifacts, see https://ripke.com/loadavg/moire. I recommend using (60/13).
> - Queue length do not include current running count, only the waiting-count, unlike Unix load average. You gotta add processor util% to match

Or to put it another way, let's not change the function signature to accept new args (`getloadavg(selector="qu", instant=False)`). Only focus on returning a more accurate value for now, since you seem to suggest that the current one is unrealistic. The minimal the changes, the better.

And once that is done I'll probably ask @ammaraskar if he can take a look, because this part of the code is obscure to me.

giampaolo:
Closing this out per my previous comment. Feel free to open another PR which returns more accurate results without changing the function signature (that, if needed, should be discussed in a ticket first).

Thanks

PR #2688, opened 2025-11-16
[illumos] skip test_cpu_times_comparison() test on SunOS

mtelka:
## Summary

- OS: illumos
- Bug fix: yes
- Type: tests
- Fixes: 

## Description

This test is flaky and so it is better to skip it on illumos.


giampaolo:
Thank you

PR #2700, opened 2025-12-29
Static metadata

dimbleby:
Specify platform-specific dependencies using PEP508 markers.

So that all wheels contain the same metadata: allowing tools like uv and poetry, which build cross-platform lockfiles, to get the right answers.

### DIFF
--- a/.github/workflows/sunos.yml
+++ b/.github/workflows/sunos.yml
@@ -34,5 +34,5 @@ jobs:
             python3 setup.py build_ext -i --parallel 4
             python3 -c "import psutil"
             python3 scripts/internal/install_pip.py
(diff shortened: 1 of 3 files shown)


dimbleby:
That doesn't sound like it should be a problem!   But perhaps you are looking for `--upgrade-strategy eager` https://pip.pypa.io/en/stable/development/architecture/upgrade-options/

giampaolo:
I remember I tried that but didn't work. The test should be: 
* install an old version of, say, pytest: `pip install pytest==8.0.0`
* run `make install-pydeps-test`
* pytest should be upgraded to 9.0.2

dimbleby:
None of this sounds like a problem to me.  If you require eg pytest 9.0.2, then just say so in your dependencies.  

When is the CI ever installing older versions anyway?

PR #2706, opened 2026-01-17
Eliminate busy-loop in `Process.wait()` on Linux and BSD

### LINKED ISSUE #2705: [Linux, macOS, BSD] speedup `Process.wait()`
## Summary

- OS: Linux, macOS, BSD
- Type: core

## Description

See https://github.com/giampaolo/psutil/issues/2703#issuecomment-3725848010.

`Process.wait()` implementation uses a busy loop with small sleeps, see: https://github.com/giampaolo/psutil/blob/77cfb0c6/psutil/_psposix.py#L61-L154

On Linux, this can be made more efficient by using `pidfd_open()` and passing the FD to a multiplexor like `select()`, `poll()` or `epoll()`.

On BSD and macOS we can use `kqueue()`.

giampaolo:
## Summary

- OS: Linux, macOS, BSD
- Bug fix: no
- Type: performance
- Fixes: #2705

## Description

* Linux: `Process.wait()` now uses `pidfd_open()` + `poll()` instead of busy-loop polling. Requires Linux >= 5.3, Python >= 3.9.
* macOS/BSD: `Process.wait()` now uses `kqueue()` instead of busy-loop polling.
* Both implementations fall back to traditional busy loop polling if not available or on certain errors.

EDIT: this change has been proposed for inclusion into cPython: https://github.com/python/cpython/pull/144047.

EDIT: this change was turned into a blogpost: https://gmpy.dev/blog/2026/event-driven-process-waiting.

### DIFF
--- a/.github/workflows/bsd.yml
+++ b/.github/workflows/bsd.yml
@@ -9,15 +9,20 @@ on:
     # only run this job if the following files are modified
     paths: &bsd_paths
       - ".github/workflows/bsd.yml"
+      - "psutil/__init__.py"
+      - "psutil/_common.py"
+      - "psutil/_ntuples.py"
       - "psutil/_psbsd.py"
+      - "psutil/_psposix.py"
       - "psutil/_psutil_bsd.c"
       - "psutil/arch/bsd/**"
       - "psutil/arch/freebsd/**"
       - "psutil/arch/netbsd/**"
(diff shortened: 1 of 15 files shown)


giampaolo:
This change was turned into a blogpost: https://gmpy.dev/blog/2026/event-driven-process-waiting.

PR #2717, opened 2026-02-11
Re-enable tests after MacOs cpu_freq 1892 was fixed

kbaikov:
## Summary

- OS: several
- Bug fix: no
- Type: tests
- Fixes: -

## Description

  Remove pytest skipif decorator from some tests that needed the 1892 to be fixed since the 1892 is fixed

### DIFF
--- a/tests/test_contracts.py
+++ b/tests/test_contracts.py
@@ -111,7 +111,9 @@ class TestAvailSystemAPIs(PsutilTestCase):
     def test_win_service_get(self):
         assert hasattr(psutil, "win_service_get") == WINDOWS
 
-    @pytest.mark.skipif(MACOS and AARCH64, reason="skipped due to #1892")
+    @pytest.mark.skipif(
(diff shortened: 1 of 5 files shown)


tests/test_contracts.py
```diff
@@ -111,7 +110,6 @@ def test_win_service_iter(self):
     def test_win_service_get(self):
         assert hasattr(psutil, "win_service_get") == WINDOWS
 
-    @pytest.mark.skipif(MACOS and AARCH64, reason="skipped due to #1892")
     def test_cpu_freq(self):
```
giampaolo: This makes CI fail. I think you can add (not tested):

```
@pytest.mark.skipif(MACOS and AARCH64 and not HAS_CPU_FREQ, reason="not available on MACOS + AARCH64")
```

giampaolo (changes requested):

PR #2718, opened 2026-02-12
Re-enable tests after BSD open_files 595 was fixed

kbaikov:
## Summary

- OS: BSD
- Bug fix: no
- Type: tests
- Fixes: -

## Description

An attempt to re-enable tests after BSD open_files #595 was fixed.
Not sure if CI has BSD tests.


giampaolo:
It was never fixed. BSD simply does not provide a way to retrieve the file paths.

PR #2720, opened 2026-02-13
linux: improve cpu_count() fallback for offline CPUs

### LINKED ISSUE #2683: TestSystemCPUCountLogical fails on sparc64 due to wrong data sources
On sparc64, the test `TestSystemCPUCountLogical` fails with the following error message:

```
____________ TestSystemCPUCountLogical.test_against_sysdev_cpu_num _____________
psutil/tests/test_linux.py:664: in test_against_sysdev_cpu_num
    assert psutil.cpu_count() == count
E   assert 48 == 64
E    +  where 48 = <function cpu_count at 0xfff80001058cb240>()
E    +    where <function cpu_count at 0xfff80001058cb240> = psutil.cpu_count

psutil/tests/test_linux.py::TestSystemCPUCountLogical::test_against_sysdev_cpu_online PASSED
psutil/tests/test_linux.py::TestSystemCPUCountLogical::test_emulate_fallbacks FAILED
```

Full build log available at: https://buildd.debian.org/status/fetch.php?pkg=python-psutil&arch=sparc64&ver=7.1.0-1&stamp=1762556398&raw=0

The reason why the assert in this case fails is because the number of CPUs reported as 48 here is the number of **online** CPUs while the number 64 is the number of **configured** CPUs. Apparently, the code in `psutil` does not differentiate both and hence the assert fails.

From the manpage of `sysconf()`:

```
        - _SC_NPROCESSORS_CONF
              The number of processors configured.  See also get_nprocs_conf(3).

        - _SC_NPROCESSORS_ONLN
              The number of processors currently online (available).  See also get_nprocs_conf(3).
```

I assume that `psutil` is somehow querying `sysconf()` and asking for configured processors (`_SC_NPROCESSORS_CONF`) when it should be querying for online processors (`_SC_NPROCESSORS_ONLN`).

Access to a SPARC machine running both Solaris and Linux can be obtained through the GCC Compile Farm: https://gcc.gnu.org/wiki/CompileFarm

fabiancrt:
## Summary

- OS: Linux (offline CPUs / sparc64 reported in #2683)
- Bug fix: yes
- Type: core, tests
- Fixes: #2683

## Description

**psutil.cpu_count(logical=True)** uses **os.sysconf("SC_NPROCESSORS_ONLN")** when available and on some platforms (for example, sparc64 with offline CPUs), tests were comparing against configured CPUs rather than comparing against online CPUs

in this PR sysconf fast path is unchanged, also when **os.sysconf()** is unavailable (when it raises ValueError), the fallback path now uses **/sys/devices/system/cpu/online** (cpulist format ) before falling back to the existing **/proc/cpuinfo** and /**proc/stat** parsing. also tests were updated to check the sysfs online format and the fallback order

tested with:
`python -m pytest -q tests/test_linux.py -k CPUCountLogical -vv`

note:
 added per-file ignore for **RUF069** in **tests/test_linux.py** to avoid float-equality assertions in that file blocking commits

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -532,12 +532,45 @@ def per_cpu_times():
         return cpus
 
 
+def _parse_cpulist(cpulist):
+    """Parse a Linux CPU list string (e.g. "0-3,8,10-11")."""
+    cpulist = cpulist.strip()
+    if not cpulist:
+        return 0
+    num = 0
+    for chunk in cpulist.split(','):
+        chunk = chunk.strip()
+        if not chunk:
+            continue
+        if '-' in chunk:
+            start, _, end = chunk.partition('-')
+            start = int(start)
+            end = int(end)
+            if start > end:
+                msg = f"invalid CPU range {chunk!r}"
+                raise ValueError(msg)
+            num += end - start + 1
+        else:
+            int(chunk)
+            num += 1
+    return num
+
+
 def cpu_count_logical():
     """Return the number of logical CPUs in the system."""
     try:
         return os.sysconf("SC_NPROCESSORS_ONLN")
     except ValueError:
-        # as a second fallback we try to parse /proc/cpuinfo
+        # Fallback: /sys/devices/system/cpu/online
+        try:
+            with open_text("/sys/devices/system/cpu/online") as f:
+                num = _parse_cpulist(f.read())
+        except (OSError, ValueError):
+            num = 0
+        if num > 0:
+            return num
+
+        # Fallback: /proc/cpuinfo
         num = 0
         with open_binary(f"{get_procfs_path()}/cpuinfo") as f:
             for line in f:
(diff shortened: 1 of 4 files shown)


giampaolo:
Thanks, but I fixed #2683 instead: 42f1d734 and 72763161.

It was purely a test bug. The `_pslinux.py` change here can't help: the new `/sys/devices/system/cpu/online` lookup only runs inside `except ValueError:`. I checked on a sparc64 box and `os.sysconf("SC_NPROCESSORS_ONLN")` works and returns the correct 64. The tests were comparing against *possible* CPUs (`nproc --all` = 256, 256 `cpu[0-9]*` dirs, `Cpus_allowed_list` = 0-255) while `cpu_count()` reports online ones.

With that said, if you want to pick up the cpulist parser separately: make it return a list rather than a count, and use it in `_get_eligible_cpus()` (`_pslinux.py:2120`), which today only matches a single `N-M` range with a regex and silently gives up on `0-3,8`, returning an error message which is not 100% correct.

PR #2721, opened 2026-02-14
Ignore unicode error on BSD cpu_freq().

kbaikov:
## Summary

- OS: *BSD
- Bug fix: yes
- Type: core
- Fixes: 2635, 1801

## Description

Since the actual fix would be somewhere in C psutil_cpu_freq() to handle the bytes with `PyUnicode_FromString`
I think this is good enough for now.


kbaikov:
@giampaolo Hi, could you please re-trigger the github actions for this PR? I think windows fails because of some leftover artifact and linters not sure why.

kbaikov:
@giampaolo Hi, the pr is green now. Any comment/review?

PR #2810, opened 2026-04-08
[Linux] virtual_memory() fails when /proc/meminfo contains ShadowCallStack >= 10GB

### LINKED ISSUE #2809: [Linux] virtual_memory() fails when /proc/meminfo contains ShadowCallStack >= 10MB
## Summary

- OS: Linux
- Architecture: ARM
- Psutil version: 7.1.2
- Python version: 3.12.9
- Type: core

## Description

On an arm64 Linux machine, `psutil.virtual_memory()` fails when `/proc/meminfo` contains a line like:

```text
ShadowCallStack:10373888 kB
```

Because the parsing in `_pslinux.py` breaks when no space is present after the metric name:

https://github.com/giampaolo/psutil/blob/a59dae7/psutil/_pslinux.py#L258-L259

```py
            fields = line.split()
            mems[fields[0]] = int(fields[1]) * 1024
```

The upstream cause is actually due to a quirk in the Linux kernel, where for some reason `ShadowCallStack` is printed with padding up to 8 digits (none of the other meminfo fields are like this) but without a guaranteed space:

https://github.com/torvalds/linux/blob/8356a5a3b078ca89c526dd6d71e9a76fec571c37/fs/proc/meminfo.c#L113-L116

```cpp
	seq_printf(m, "ShadowCallStack:%8lu kB\n",
```

So whenever `ShadowCallStack` >= 10MB this issue occurs! The simplest suggested fix would be to split by colon instead. Happy to submit a PR if this sounds good.

hansonw:
## Summary

- OS: Linux
- Bug fix: yes
- Type: core
- Fixes: #2809

## Description

See issue description. Fixes an issue on arm64 Linux where `psutil.virtual_memory()` may fail when /proc/meminfo contains `ShadowCallStack` with a value greater than 10GB, e.g. `ShadowCallStack:10373888 kB` - the lack of space after `ShadowCallStack:` breaks the current parsing logic.


gabrielchangamire-arch:
FWIW I independently reproduced this on arm64 (mocked `/proc/meminfo` containing `ShadowCallStack:10373888 kB`) and confirmed this fix resolves the `ValueError` in `virtual_memory()`.

Note that `swap_memory()` has the same two parsing lines and crashes the same way — I've opened #2865 applying the same fix there, so the two PRs together should fully address #2809.

Issue #2382, opened 2024-03-15
[macOS] cpu_freq() fails on arm64


giampaolo:
We recently added a macOS build for arm64 and we get a failure:
https://github.com/giampaolo/psutil/actions/runs/8301314525/job/22721158479

```
2024-03-15T19:24:20.3697330Z ======================================================================
2024-03-15T19:24:20.3697970Z ERROR: psutil.tests.test_misc.TestMisc.test_serialization (name='cpu_freq')
2024-03-15T19:24:20.3698460Z ----------------------------------------------------------------------
2024-03-15T19:24:20.3698790Z Traceback (most recent call last):
2024-03-15T19:24:20.3708960Z   File "/Users/runner/work/psutil/psutil/psutil/tests/test_misc.py", line 292, in test_serialization
2024-03-15T19:24:20.3709580Z     ret = fun()
2024-03-15T19:24:20.3710550Z   File "/private/var/folders/1k/qq3pcbf12vb6vyblh81736p40000gn/T/cibw-run-m728_sh2/cp39-macosx_arm64/venv-test-arm64/lib/python3.9/site-packages/psutil/__init__.py", line 1913, in cpu_freq
2024-03-15T19:24:20.3711420Z     ret = _psplatform.cpu_freq()
2024-03-15T19:24:20.3712370Z   File "/private/var/folders/1k/qq3pcbf12vb6vyblh81736p40000gn/T/cibw-run-m728_sh2/cp39-macosx_arm64/venv-test-arm64/lib/python3.9/site-packages/psutil/_psosx.py", line 182, in cpu_freq
2024-03-15T19:24:20.3713240Z     curr, min_, max_ = cext.cpu_freq()
2024-03-15T19:24:20.3713680Z RuntimeError: 'pmgr' entry was not found in AppleARMIODevice service
```

Error originates from: https://github.com/giampaolo/psutil/blob/f51f62beef5e2ce0d64c9cb76fb7b4e9094b1864/psutil/arch/osx/cpu.c#L160-L166.

Change was introduced in https://github.com/giampaolo/psutil/pull/2222. 

@snOm3ad:  Any chance you can take a look?

snOm3ad:
I will check further! Will also look into #2354 at the same time.

dbwiddis:
I think this is a relic of the VM used by GitHub (and other CI providers) that don't include the `pmgr` info that on-metal hardware provides.

snOm3ad:
[...]

Issue #2285, opened 2023-08-01
Snowflake -- cpu_count() got an unexpected keyword argument 'logical'


wishshi:
## Summary

* Psutil version: 5.9.0
* Python version: { python3.8 - 3.10 }

## Description
Try to build a procedure with statement "psutil.cpu_count(logical=False)" and got error message:

> TypeError: cpu_count() got an unexpected keyword argument 'logical'

BTW, without augument, cpu_count() return -1 on Snowflake.

SimonHeimberg:
That is strange. In [5.9.0 `psutil.cpu_count()` ](https://github.com/giampaolo/psutil/blob/release-5.9.0/psutil/__init__.py#L1570) has the argument. And it checks for value < 0 and replaces it with `None`.

What does `os.cpu_count()` return? (no argument supported)

What is Snowflake? The cloud service on snowflake . com ? Any more information about the platform (`uname -a`, ...)?

giampaolo:
Snowflake ships its own stubbed psutil rather than the real one (a bare cpu_count() returning -1 is another giveaway, we never do that). Nothing we can fix on our side, you'd have to raise it with them.

Issue #11, opened 2014-05-23
Get process parent


giampaolo:
_From [l.vbo...@gmail.com](https://code.google.com/u/116335943976280081305/) on February 16, 2009 23:02:35_

```
As a new feature, it would be nice to get the process parent as a psutil class.

Example:

>>> parent = psutil.parent(os.getpid())
>>> parent.pid
11038
>>> parentPID = psutil.Process(11038).parent.pid
>>> parentPID
1

Thank you.
```

_Original issue: http://code.google.com/p/psutil/issues/detail?id=11_

PR #2824, opened 2026-04-20
[macOS] fix cpu_freq on Apple M4 and M5-family chips

### LINKED ISSUE #2642: [MacOS] psutil.cpu_freq() results are off by 1000x on Apple M4
## Summary

- OS: MacOS
- Architecture: arm64
- Psutil version: 7.1.0
- Python version: 3.13.7
- Type: core

## Description

Starting on Apple M4 (not on Apple M3) cpu_freq() returns values that are 1000x too small, i.e showing (min, current, max) frequency as (1, 4, 4). Correct values should be approx. (1050, 4500, 4500).

This seems to boil down to a change that Apple made in the power management service where the cpu frequency is extracted from.  Starting from M4, the internal representation used by Apple seems to have changed from Hz to kHz.  Since psutil has a hard-coded division by 1000000 (see [here](https://github.com/giampaolo/psutil/blob/master/psutil/arch/osx/cpu.c#L257)), this results in psutil return the cpu frequency values in GHz instead of MHz.  

No Apple documentation can be found related to this change, but Apple typically doesn't seem to provide details like this anyway.

### 1. detect cpu type

pseudo-code:

```
If (cpu in [M1, M2, M3]):
   divide by 1_000_000
else
  divide by 1_000
```

### 2. heuristic to always return cpu freq. in appropriate range 

Given that this code path is only applied for arm64 Apple CPUs, it is pretty safe to assume the max cpu frequency should be somewhere in the range of [100MHz, 100GHz].  

pseudo-code:

```
If (100_000_000 < extracted_cpu_freq):
   divide by 1_000_000
else
  divide by 1_000
```

## Suggestion

I believe option 2. is the simplest one to implement and likely to be sufficiently generic (assuming Apple will stick with the current implementation for a few generations).  

I can assist with testing, as I have both M3 & M4 hardware available.

bertpl:
## Summary

- OS: macOS
- Bug fix: yes
- Type: core
- Fixes: #2642

### Context

On Apple Silicon, macOS exposes no public API for CPU frequency. `psutil` therefore reads the DVFS (dynamic voltage and frequency scaling) tables that the power manager (`pmgr` entry of `AppleARMIODevice`) publishes in the IORegistry: properties named `voltage-states<N>-sram`, containing `(frequency, voltage)` pairs as raw little-endian integers.

These tables are undocumented by Apple, and they don't only describe CPU cores — the same key family also covers GPU, NPU and fabric clusters. Until now, `psutil` relied on two conventions that happened to hold from M1 through M3:

* the CPU tables live at **fixed indexes**: `voltage-states5-sram` (P cores) and `voltage-states1-sram` (E cores);
* frequencies are encoded in **Hz**, so a hardcoded division by 10⁶ yields MHz.

### Bug

On M4 and later chips, both conventions were broken by (undocumented) hardware/firmware changes:

* **M4 and above** — frequencies are now encoded in **kHz** instead of Hz, so the hardcoded ÷10⁶ produces values ~1000× too small (`cpu_freq()` reporting e.g. `4.5 MHz` instead of `4512 MHz`). This is #2642.
* **M5 and above** — the `voltage-states<N>-sram` index space was **renumbered**, so the hardcoded indexes 5 and 1 no longer point at CPU clusters at all.

The hard part of fixing this is *not* finding the updated indexes and division factors for M4/M5 — it is finding a mechanism that doesn't need re-fixing on every future M-series release. A lookup table keyed on chip generation would break again, silently or loudly, each time Apple reshuffles the registry.

### Adjacent ecosystem fixes

Other open-source tools read the exact same IORegistry properties and were hit by the same two changes. How they fixed it:

| Tool | M4 unit fix (Hz → kHz) | M5 index renumbering | Approach |
|---|---|---|---|
| [fastfetch](https://github.com/fastfetch-cli/fastfetch) ([#1394](https://github.com/fastfetch-cli/fastfetch/discussions/1394), Nov 2024) | Per-value magnitude heuristic: `> 10⁸ ⇒ Hz, else kHz` | ❌ not handled — still hardcodes `voltage-states5-sram` | Partial: dynamic unit, fixed index |
| [macmon](https://github.com/vladkens/macmon) ([#11](https://github.com/vladkens/macmon/issues/11) → v0.4.0, Dec 2024; v0.7.0, Jun 2025) | Keyed on chip-name substring (`M1/M2/M3 ⇒ Hz, else kHz`) | Fallback to Apple's `acc-clusters` property when the hardcoded keys are missing | Hardcoded primary path + structural fallback |
| [stats](https://github.com/exelban/stats) ([#2285](https://github.com/exelban/stats/issues/2285) → v2.11.24, Jan 2025; v2.11.67 for M5) | Per-generation handling | Handled in a separate M5-specific fix | Per-generation conditionals |
| [macpow](https://github.com/k06a/macpow) | Per-table magnitude heuristic | Full dynamic enumeration — no hardcoded indexes at all | Enumerate & classify (fmax threshold) |
| **this PR** | Per-value magnitude heuristic | Full dynamic enumeration — no hardcoded indexes | Enumerate & classify, simplified: no E/P/S labeling needed since `psutil` exposes only overall min/max |

Two observations from this survey:

1. The tools that kept hardcoded indexes (fastfetch) or per-generation conditionals (stats, macmon's primary path) each needed a *second* fix when M5 arrived — confirming that generation-keyed logic is a recurring maintenance burden.
2. macmon's `acc-clusters` fallback is not trustworthy as a general mechanism: on a physical M3 Max we found `acc-clusters` mislabels `voltage-states5-sram` as E-core and omits the real E-core table entirely (macmon never hits this in practice only because its hardcoded primary path succeeds on M3).

The most future-proof design in the ecosystem is macpow's: **enumerate all tables, classify by content, hardcode nothing**. That is the approach this PR adopts.

### Solution proposed in this PR

`psutil_cpu_freq()` now discovers the CPU tables dynamically instead of assuming indexes or units:

1. **Enumerate** all properties on the `pmgr` entry matching `voltage-states<N>-sram` (digit-only `<N>`). No hardcoded indexes — renumbering (M5) is handled automatically.
2. **Decode** each table as little-endian `uint32` `(frequency, voltage)` pairs.
3. **Detect the unit per value, by magnitude**: raw value > 10⁸ ⇒ Hz (÷10⁶); else ⇒ kHz (÷10³). A real CPU frequency can't be below 100 MHz-as-Hz or above 100 GHz-as-kHz, so the two encodings are unambiguous. This handles M1–M3 (Hz) and M4+ (kHz) with no generation table.
4. **Filter implausible values**: keep only 500–20000 MHz, dropping terminator zeros and junk entries.
5. **Classify CPU vs. non-CPU tables**: a table qualifies as a CPU cluster only if its maximum frequency is **≥ 2000 MHz**. This is what filters out the GPU/NPU/fabric DVFS tables, which share the `-sram` suffix.
6. Report `min` = lowest qualifying frequency, `max` = highest; `current` = `max` (Apple Silicon exposes no live-frequency API, matching previous behavior).

Since `psutil` only exposes an overall min/max (not per-cluster values), no E/P/S/GPU *labeling* is needed — only the CPU/non-CPU split, which makes the 2000 MHz threshold the single load-bearing heuristic.

Note the classification exists primarily to keep `min` correct: `max` alone would need no filter, since the fastest table is always a CPU cluster — but without it, the low rungs of GPU/NPU ladders (e.g. 648 MHz on M3 Max, above the 500 MHz plausibility floor yet below the real CPU minimum of 1020) would leak into the reported minimum.

**Why not the obvious alternatives?**

- *Keep the hardcoded indexes and just update them for M4/M5?* Breaks again at the next renumbering — M5 already proved this happens, and fastfetch (which kept hardcoded index 5) remains exposed to exactly this on M5-family chips.
- *Dispatch on chip generation (via `machdep.cpu.brand_string`)?* An earlier draft of this PR did exactly that. It works, but (a) it couples correctness to a lookup table that needs updating on every Apple release, and (b) it forces users on unknown future chips onto a fallback path. The magnitude of the raw values already tells us the unit — no generation awareness needed.
- *Use Apple's `acc-clusters` property (macmon's M5+ fallback)?* Empirically untrustworthy — see observation 2 above (mislabels and omits CPU clusters on M3 Max).

**Why 2000 MHz is a safe threshold.** Every non-CPU DVFS ladder we could find data for peaks well below 2 GHz — GPUs are usually the closest (ANE/fabric are comparable or lower; on M1 Max the ANE ladder peaks at 1500 MHz, slightly above its GPU's 1296). So the threshold only needs to separate the slowest CPU cluster ever shipped from the fastest GPU:

*CPU cluster frequency ranges per generation (MHz):*

| Chip family | E-core max | P-core max | S-core max |
|---|---|---|---|
| M1 (all variants) | 2064 | 3228 | — |
| M2 (all variants) | 2424 | 3696 | — |
| M3 (all variants) | 2748 | 4056 | — |
| M4 (all variants) | 2892 | 4512 | — |
| M5 (all variants) | 3048 | 4380 | 4608 |

*(M5 introduced a third CPU cluster type: "Super" (S) cores, clocked above the P-cores. For the threshold discussion only the slowest and fastest CPU clusters matter, so S-cores fold in transparently.)*

*Non-CPU DVFS tables observed on a physical M3 Max (all `-sram` suffixed, i.e. all would be misclassified by a suffix-only filter):*

| Table | fmax (MHz) | Component |
|---|---|---|
| `voltage-states1-sram` | 2568 | CPU (E cluster) |
| `voltage-states5-sram` | 4056 | CPU (P cluster) |
| `voltage-states13-sram` | 4056 | CPU (P cluster, duplicate on Max) |
| `voltage-states9-sram` | 1380 | GPU |
| `voltage-states14/15/16-sram` | 1380 | GPU/NPU/fabric |

The slowest CPU cluster ever shipped (M1 E-cores) peaks at **2064 MHz**; measured GPU fmax is 1278/1398/1380/1578 MHz on M1/M2/M3/M4 (Refs 7–8), with base-M5 measurements around 1578–1620 MHz (GPU clocks have historically been shared across base/Pro/Max variants within a generation) — so the 2000 MHz threshold sits between CPU and non-CPU on every generation to date. (The long-term evolution of this gap, and a possible relative-threshold variant, are discussed in a comment below.)

### Implementation notes

All changes live in `psutil_cpu_freq()` in `psutil/arch/osx/cpu.c` (single code path, no generation detection, no `sysctl` lookups, no `acc-clusters` dependency):

- Properties are enumerated via `IORegistryEntryCreateCFProperties()` on the `pmgr` entry; a key qualifies only if it matches `voltage-states<N>-sram` with digit-only `<N>` (defensive against future non-numeric key variants).
- Each qualifying `CFData` blob is walked in 8-byte strides: 4-byte little-endian frequency + 4-byte voltage; only the frequency is read.
- The named constants encoding the heuristics described above:
  - `UNIT_THRESHOLD = 1e8` — raw value above ⇒ Hz (÷10⁶), below ⇒ kHz (÷10³); see step 3 above for why the encodings can't collide.
  - `MIN_PLAUSIBLE_MHZ = 500` / `MAX_PLAUSIBLE_MHZ = 20000` — per-value filter dropping terminator zeros and junk entries.
  - `CPU_CLUSTER_MIN_FMAX_MHZ = 2000` — per-table CPU-vs-GPU/NPU classification.
- Min/max are first accumulated **per table**, and a table's values are folded into the overall result only if the table qualifies as a CPU cluster — so a GPU table's low frequencies can never leak into the reported minimum.
- Duplicate CPU tables (Max/Ultra chips replicate P-cluster tables, e.g. `voltage-states5-sram` and `voltage-states13-sram` on M3 Max) are harmless: they contribute the same min/max already represented, so no deduplication is needed.
- If the `pmgr` entry is missing (e.g. on virtualized ARM64 such as CI runners) or no table qualifies, the function returns `None` (with a `psutil_debug` message), which `cpu_freq()` surfaces as an empty list — following the contract introduced in #2382 / #2868.
- `current` is reported as `max`, unchanged from the previous implementation (Apple Silicon exposes no live-frequency API). The forward-risk trade-off of the 2 GHz threshold is documented in a comment block above the function, so a future maintainer hits the reasoning before the code.

### Tests

No new test is included. An earlier revision added a plausibility-bounds test, but it was removed during review: the C code already clamps every value to the same [500 MHz, 20 GHz] range, making the assertions tautological, and CI runs on virtualized ARM64 where `cpu_freq()` returns no data and the test would skip anyway. Verification is by direct measurement on physical hardware (below).

### Validated on physical hardware

| Machine | Result | Cross-check |
|---|---|---|
| Apple M3 Max (macOS 15) — Hz path | `scpufreq(current=4056, min=1020, max=4056)` | matches published M3 P-core peak of 4056 MHz ✓ |
| Apple M4 Pro — kHz path | `scpufreq(current=4512, min=1020, max=4512)` | matches published M4 P-core peak of 4512 MHz ✓ — the hardware class where #2642 was filed |

Also validated on a base M3 (Hz path). Values on all three machines match Apple's published specs and Eclectic Light's crowdsourced DVFS tables (Ref 5).

### References

1. fastfetch Discussion #1394 — primary external confirmation of the Hz → kHz unit change on M4 (maintainer diagnosis: *"Apple changed the unit reported by IO service. Your M4 Pro reports kHz while my M1 Pro reports Hz"*), including a raw ioreg dump that only decodes plausibly as kHz: https://github.com/fastfetch-cli/fastfetch/discussions/1394
2. macmon Issue #11 (M4 unit bug) and v0.7.0 release notes (M5 Max renumbered-keys crash fix): https://github.com/vladkens/macmon/issues/11, https://github.com/vladkens/macmon/releases
3. stats Issue #2285 — the same M4 bug independently hit and fixed in a third tool (v2.11.24): https://github.com/exelban/stats/issues/2285
4. k06a/macpow — reference implementation of the dynamic-enumeration pattern adopted here: https://github.com/k06a/macpow/blob/main/src/ioreport.rs
5. Eclectic Light — crowdsourced per-chip DVFS min/max tables for all Apple Silicon Macs (ground truth used for validation): https://eclecticlight.co/2026/04/13/cpu-core-frequencies-updated-for-all-current-apple-silicon-macs/
6. Wikipedia — Apple M1/M4/M5 (chip release dates, peak frequencies, variant line-up): https://en.wikipedia.org/wiki/Apple_M1, https://en.wikipedia.org/wiki/Apple_M4, https://en.wikipedia.org/wiki/Apple_M5
7. Eclectic Light — "Power modes and Apple silicon GPUs" (measured GPU fmax: M3 Pro 1380 MHz, M4 Pro 1578 MHz): https://eclecticlight.co/2025/01/06/power-modes-and-apple-silicon-gpus/ — and ASi-Mac-DVFS (per-chip DVFS ladder dumps incl. GPU/ANE, e.g. M1 Max GPU 1296 / ANE 1500 MHz): https://github.com/GaoKaipeng2023/ASi-Mac-DVFS
8. "Apple vs. Oranges: Evaluating the Apple Silicon M-Series SoCs for HPC Performance and Efficiency" (Table I: GPU clocks M1 1.27 / M2 1.39 / M3 1.38 / M4 1.47 GHz): https://arxiv.org/html/2502.05317v1

---

*AI assistance was used for reference research, prose drafting, and to cross-check the fix; all code and technical decisions reviewed by me.*

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -224,6 +224,14 @@ Others:
 - :gh:`2411` [macOS]: :meth:`Process.cpu_times` and :meth:`Process.cpu_percent`
   calculation on macOS x86_64 (arm64 is fine) was highly inaccurate (41.67x
   lower).
+- :gh:`2642`, [macOS]: fix :func:`cpu_freq` on Apple Silicon. On M4+ it
+  returned values ~1000x too small because the ``voltage-statesN-sram``
+  IORegistry tables switched from Hz to kHz; on M5-family chips it failed
+  because the hardcoded table indexes were renumbered. The implementation now
+  enumerates CPU ``voltage-states*-sram`` tables dynamically, detects the unit
+  per-value by magnitude, and filters CPU clusters from GPU/NPU tables via a
+  per-table fmax threshold. Works uniformly from M1 through M5 Max. (patch by
+  Bert Pluymers)
 - :gh:`2711`, [Windows]: :func:`net_if_addrs` was returning ``None`` for the
   ``broadcast`` field of network interfaces instead of the correct broadcast
   address.
--- a/docs/credits.rst
+++ b/docs/credits.rst
@@ -111,6 +111,7 @@ Code contributors by year
 
 * :user:`Alex Chen <l46983284-cpu>` - :gh:`2859`
 * :user:`Amaan Qureshi <amaanq>` - :gh:`2770`
+* :user:`Bert Pluymers <bertpl>` - :gh:`2642`
 * :user:`Ding Qiuran <dynapx>` - :gh:`2860`
 * :user:`Felix Yan <felixonmars>` - :gh:`2732`
 * :user:`Gabriel Changamire <gabrielchangamire-arch>` - :gh:`2809`
--- a/psutil/arch/osx/cpu.c
+++ b/psutil/arch/osx/cpu.c
@@ -170,14 +170,46 @@ psutil_find_pmgr_entry(io_registry_entry_t *out_entry) {
     return 0;
 }
 
+// Report overall min/max CPU frequency by enumerating every
+// `voltage-states<N>-sram` property on the pmgr entry, instead of
+// reading fixed indexes: Apple renumbered that key space on M5, so
+// any hardcoded index breaks there.
+//
+// Frequencies are Hz on M1-M3 and kHz on M4+. We pick the unit per
+// value by magnitude rather than keying on chip generation: real
+// clocks are ~0.6-4.6 GHz, so Hz values sit well above 1e8 and kHz
+// values well below it.
+//
+// The `-sram` suffix isn't CPU-specific: GPU/NPU/fabric publish
+// -sram DVFS tables too (fmax ~1.4 GHz on M3 Max). A table counts as
+// a CPU cluster only if its fmax is >= 2 GHz, which clears every GPU
+// peak seen so far but stays under the slowest CPU cluster shipped
+// (M1 E-core, 2064 MHz). GPU clocks rise ~5-6%/gen, so revisit this
+// floor if a future chip closes the gap.
 PyObject *
 psutil_cpu_freq(PyObject *self, PyObject *args) {
+    static const char KEY_PREFIX[] = "voltage-states";
+    static const char KEY_SUFFIX[] = "-sram";
+    static const size_t KEY_PREFIX_LEN = sizeof(KEY_PREFIX) - 1;
+    static const size_t KEY_SUFFIX_LEN = sizeof(KEY_SUFFIX) - 1;
+    // Raw frequency values above 10^8 are Hz; below are kHz.
+    static const uint32_t UNIT_THRESHOLD = 100000000U;
+    // Per-value plausibility bounds. Drops terminator zeros and any
+    // junk the enumeration might pick up.
+    static const uint32_t MIN_PLAUSIBLE_MHZ = 500;
+    static const uint32_t MAX_PLAUSIBLE_MHZ = 20000;
+    // Per-table classification: CPU clusters have fmax >= 2 GHz. See
+    // block comment above for justification.
+    static const uint32_t CPU_CLUSTER_MIN_FMAX_MHZ = 2000;
+
     io_registry_entry_t entry = IO_OBJECT_NULL;
-    CFTypeRef pCoreRef = NULL;
-    CFTypeRef eCoreRef = NULL;
-    size_t pCoreLength = 0;
-    uint32_t pMin = 0, eMin = 0, min = 0, max = 0, curr = 0;
-    PyObject *py_ret = NULL;
+    CFMutableDictionaryRef props = NULL;
+    const void **keys = NULL;
+    const void **values = NULL;
+    CFIndex count = 0;
+    uint32_t min_mhz = UINT32_MAX;
+    uint32_t max_mhz = 0;
+    int found_any = 0;
 
     // No 'pmgr' entry (e.g. virtualized ARM64); frequency is
     // undeterminable, so return None (see #2382).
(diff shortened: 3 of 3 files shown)


bertpl:
**On the long-term robustness of the 2 GHz threshold — an alternative worth considering**

The fixed 2000 MHz CPU-cluster threshold in this PR has one known long-term weakness: GPU frequencies are rising too, so a future GPU could eventually exceed 2 GHz and be misclassified as a CPU cluster. The growth is slow, though — ~5–6%/generation averaged over M1→M5 (1278 → ~1600 MHz), and flat-to-modest in two of the four generation steps — so at historical rates the crossing is roughly 4+ generations away.

An alternative that avoids this — not adopted in the current code, but a small change if preferred — is to make the threshold **relative to the highest frequency found across all tables** (always the fastest CPU cluster: P-cores, or S-cores on M5+):

> a table qualifies as CPU if its fmax ≥ 0.5 × overall fmax

This works on every generation investigated (narrowest margin: M1, where the ANE ladder at 1500 MHz sits 114 MHz below the cut):

| Chip | 0.5 × fastest CPU cluster | Slowest CPU cluster (E) | Fastest non-CPU |
|---|---|---|---|
| M1 | 1614 | 2064 ✅ above | 1500 (ANE; GPU 1278) ✅ below |
| M2 | 1848 | 2424 ✅ | 1398 (GPU) ✅ |
| M3 (Max) | 2028 | 2568 ✅ | 1380 (GPU) ✅ |
| M4 | 2256 | 2892 ✅ | 1578 (GPU) ✅ |
| M5 | 2304 | 3048 ✅ | ~1578–1620 (GPU, measured base M5) ✅ |

*(For M3, 2568 MHz is the DVFS-table fmax measured on our M3 Max; the published base-M3 E-core max is 2748 MHz. Both clear the cut with margin. Non-CPU values: M1–M4 measured — sources: [Eclectic Light's powermetrics measurements](https://eclecticlight.co/2025/01/06/power-modes-and-apple-silicon-gpus/), [ASi-Mac-DVFS ladder dumps](https://github.com/GaoKaipeng2023/ASi-Mac-DVFS), [arXiv HPC study, Table I](https://arxiv.org/html/2502.05317v1). M5 range reflects base-M5 measurements; Pro/Max DVFS tables are not yet publicly dumped, but GPU clocks have historically been shared across variants within a generation.)*

Because the slowest and fastest CPU clusters have historically grown at similar rates, their ratio is stable (~0.64–0.68 across M1–M5, including M5's E-to-S ratio of 0.66), keeping the slowest cluster safely above the 0.5 cut regardless of absolute clock speeds — while GPUs would need to reach *half the fastest CPU clock* to be misclassified, rather than a fixed 2 GHz.

We went with the fixed threshold for the initial version because it is simpler to reason about and demonstrably correct on all shipped hardware, but happy to switch to the dynamic threshold if that's preferred.

giampaolo (changes requested):
Thanks for the thorough writeup. The approach looks resonable. Three small things before merge:

1. The block comment is too long. Suggested trimmed version:

```c
// Report overall min/max CPU frequency by enumerating every
// `voltage-states<N>-sram` property on the pmgr entry, instead of
// reading fixed indexes: Apple renumbered that key space on M5, so
// any hardcoded index breaks there.
//
// Frequencies are Hz on M1-M3 and kHz on M4+. We pick the unit per
// value by magnitude rather than keying on chip generation: real
// clocks are ~0.6-4.6 GHz, so Hz values sit well above 1e8 and kHz
// values well below it.
//
// The `-sram` suffix isn't CPU-specific: GPU/NPU/fabric publish
// -sram DVFS tables too (fmax ~1.4 GHz on M3 Max). A table counts as
// a CPU cluster only if its fmax is >= 2 GHz, which clears every GPU
// peak seen so far but stays under the slowest CPU cluster shipped
// (M1 E-core, 2064 MHz). GPU clocks rise ~10%/gen, so revisit this
// floor if a future chip closes the gap.
```

2. test_cpu_freq_bounds asserts a range the C code already clamps to, so it can't really fail (and CI is M1, where #2642 never shows). I'd remove it.

3. Credits: use `* :user:Bert Pluymers <bertpl> - :gh:2642` and drop the `.. _Bert Pluymers:` line, to match the recent migration.

Nice work.

giampaolo:
Also beware of my recent https://github.com/giampaolo/psutil/pull/2868: you'll have to rebase.

bertpl:
> Thanks for the thorough writeup. The approach looks resonable. Three small things before merge:
> 
> 1. The block comment is too long. Suggested trimmed version:
> 
> ```c
> // Report overall min/max CPU frequency by enumerating every
> // `voltage-states<N>-sram` property on the pmgr entry, instead of
> // reading fixed indexes: Apple renumbered that key space on M5, so
> // any hardcoded index breaks there.
> //
> // Frequencies are Hz on M1-M3 and kHz on M4+. We pick the unit per
> // value by magnitude rather than keying on chip generation: real
> // clocks are ~0.6-4.6 GHz, so Hz values sit well above 1e8 and kHz
> // values well below it.
> //
> // The `-sram` suffix isn't CPU-specific: GPU/NPU/fabric publish
> // -sram DVFS tables too (fmax ~1.4 GHz on M3 Max). A table counts as
> // a CPU cluster only if its fmax is >= 2 GHz, which clears every GPU
> // peak seen so far but stays under the slowest CPU cluster shipped
> // (M1 E-core, 2064 MHz). GPU clocks rise ~10%/gen, so revisit this
> // floor if a future chip closes the gap.
> ```
> 
> 2. test_cpu_freq_bounds asserts a range the C code already clamps to, so it can't really fail (and CI is M1, where [[MacOS] psutil.cpu_freq() results are off by 1000x on Apple M4 #2642](https://github.com/giampaolo/psutil/issues/2642) never shows). I'd remove it.
> 3. Credits: use `* :user:Bert Pluymers <bertpl> - :gh:2642` and drop the `.. _Bert Pluymers:` line, to match the recent migration.
> 
> Nice work.

Thx for the review!  I addressed all 3 points of feedback + rebased.

I took the liberty of updating the ~10%/gen number to **~5-6% per generation**.  The 10% number appeared in a previous PR description and was based on a misleading source for M5 GPU freq (for which it seems notoriously hard to find reliable numbers).  This then leads to the smaller %/gen number, which is also more reassuring toward keeping the 2GHz fixed heuristic threshold.  **_Happy to change that back if you prefer the round number that errs on the 'safe' side in this context._**

I will likely circle back here when I notice at some point in the future Apple GPU freqs get closer to 2GHz.

Let me know if anything else is needed for this PR.

PR #2839, opened 2026-05-31
docs: add explanatory comments to CPU and Memory examples in README

AkshajSinghal:
New users may not immediately understand what `percpu=True` or  `logical=False` means, or what the `svmem` tuple contains.  Added brief inline comments to clarify the output and parameters.

## Summary

- OS: N/A (Documentation only)
- Bug fix: no
- Type: {doc}
- Fixes: N/A

## Description

Hi Giampaolo,
I've been using `psutil` heavily for my own project (TruShell) and found the README examples very helpful. However, I noticed that new users might not immediately understand what different parameters like `percpu=True` or `logical=False` actually do, or what the fields in the `svmem` named tuple represent.

I added a few inline comments to the CPU and Memory examples in the README to make them self-explanatory for beginners. It’s a small change, but it should help new users understand the output without leaving the page.

Changes:
- Added comments explaining `interval` and `percpu` in `cpu_percent`.
- Clarified `logical=False` vs `logical=True` in `cpu_count`.
- Explained the fields in the `virtual_memory()` return value.

Hope this helps!!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,10 +125,15 @@ For the full API with more examples, see the
 .. code-block:: python
 
     >>> import psutil
+    # Get CPU usage percentage for each core over a 1-second interval
     >>> psutil.cpu_percent(interval=1, percpu=True)
     [4.0, 6.9, 3.7, 9.2]
+    
+    # Get number of physical cores (excluding hyper-threading)
     >>> psutil.cpu_count(logical=False)
     2
+    
+    # Get current CPU frequency (in MHz)
     >>> psutil.cpu_freq()
     scpufreq(current=931.42, min=800.0, max=3500.0)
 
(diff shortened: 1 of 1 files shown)


giampaolo:
Hi there. The README is supposed to be short on purpose, and just show a bunch of the most important APIs. It's the docs that is supposed to clarify these sort of things. FYI, there's a brand new doc still not "promoted" to official because I have to finalize a few things: https://psutil.readthedocs.io/latest/.
That should somewhat relief the pain your describing, because it's a complete rewrite of previous doc, and very extended in many aspects (see https://github.com/giampaolo/psutil/blob/master/docs/changelog.rst#800-in-development). E.g. there's even FAQs specifically for CPU and memory, a glossary entry re. what is a "logical" CPU etc.

PR #2840, opened 2026-05-31
Fix RCE in changelog_bot workflow

giampaolo:
The changelog-bot workflow (triggered when a maintainer comments /changelog on a PR) checked out the PR head and then executed `.github/workflows/changelog_bot.py` straight from that untrusted checkout. Although only a trusted commenter can trigger the job, the code being run comes from the PR branch, so an attacker could open a benign PR, wait for the /changelog comment, and swap in a malicious changelog_bot.py via a TOCTOU race on the mutable PR head. The result was arbitrary code execution on the runner with contents: write / pull-requests: write and access to GITHUB_TOKEN and ANTHROPIC_API_KEY (validated PoC dumped runner memory and exfiltrated the token).

Fix:

- Run the trusted changelog_bot.py from master (extracted via git show into /tmp) instead of the PR's copy. Untrusted code is never executed, which removes the RCE primitive entirely. The script only needs the PR branch checked out to commit/push the generated changelog; it reads PR data through the API.
- Pin the checkout to the PR head SHA (`gh pr view --json headRefOid`) instead of `refs/pull/<n>/head`. The intermediate `github.event.pull_request.head.sha` is null on `issue_comment` events, so this also fixes the head-SHA resolution.

Reported via Depi, validated by Garance de la Brosse (Lupin & Holmes).

### DIFF
--- a/.github/workflows/changelog_bot.yml
+++ b/.github/workflows/changelog_bot.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Checkout PR branch
         uses: actions/checkout@v5
         with:
-          ref: refs/pull/${{ github.event.issue.number }}/head
+          ref: ${{ github.event.pull_request.head.sha }}
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install dependencies


giampaolo:
I made a bit of a mess with this PR (it was incomplete, forgot to push). Definitive fix is https://github.com/giampaolo/psutil/commit/127840daad0b474065ed42583f8db4f80ecfa033.

PR #2855, opened 2026-07-02
Use multi-phase initialisation (PEP 489)

### LINKED ISSUE #2576: Use multi-phase initialisation (PEP 489)
## Summary

- OS: N/A (all)
- Type: core

## Description

Hello,

Please consider adopting multi-phase init for your Python extension modules. In short, multi-phase makes extension modules behave more like pure-Python modules and means that the import system can determine what a capabilities a module has before executing it.

Multi-phase initialisation was introduced in [PEP 489](https://peps.python.org/pep-0489/) for Python 3.5. It replaces the previous 'single-phase' mechanism by splitting the module creation process into creation and execution phases. Quoting from [the documentation](https://docs.python.org/3/c-api/module.html#multi-phase-initialization):

> Extension modules created this way behave more like Python modules: the initialization is split between the creation phase, when the module object is created, and the execution phase, when it is populated. The distinction is similar to the `__new__()` and `__init__()` methods of classes.

It is the preferred mechanism for extension module initialisation, and there is a proposal to [deprecate](https://discuss.python.org/t/89262/45) single phase init.

Thanks,
Adam

relates to #2565, as this would avoid using Unstable C API.

jinh-labs:
## Summary

- OS: all (Linux, macOS, FreeBSD, OpenBSD, NetBSD, AIX, SunOS, Windows)
- Bug fix: no
- Type: core
- Fixes: #2576

## Description

Converts all C extension modules from single-phase to **PEP 489 multi-phase initialisation**, as requested in #2576. Supersedes the abandoned #2577.

Each `PyInit_*` now returns `PyModuleDef_Init(&moduledef)` with a `Py_mod_exec` slot; the old body moves into an `_exec()` function that returns `0`/`-1`. This is a mechanical conversion that **preserves the existing runtime semantics**: as in #2577, each module keeps a `static int module_loaded` flag, and the exec slot raises `ImportError` if it runs more than once, so a module is still only ever initialised once per process (which also keeps Windows' `InitializeCriticalSection` running only once).

### Related cleanups

- The free-threading GIL declaration moved from the **unstable** `PyUnstable_Module_SetGIL()` call to the stable `Py_mod_gil` slot, with the same `#ifdef` guard (the Unstable-C-API avoidance mentioned in #2576 / #2565).
- The `m_name` of the sunos/aix/windows modules is corrected to match the import name (`psutil_sunos` → `_psutil_sunos`, etc.).
- `.m_size = 0` with no GC hooks, since no module needs per-module state yet.

### Not covered (deliberately, per review)

Kept small and mechanical. Two follow-up PRs are planned:
1. move the exception classes (`ZombieProcessError`; on Windows `TimeoutExpired` / `TimeoutAbandoned`) into per-module state, accessed via the module object;
2. implement sub-interpreter support.

## References

- [PEP 489 - Multi-phase extension module initialization](https://peps.python.org/pep-0489/)
- [CPython: Isolating Extension Modules](https://docs.python.org/3/howto/isolating-extensions.html)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -175,6 +175,9 @@ Others:
 
 - :gh:`2872`: Dropped support for Python 3.6 and 3.7. Minimum version is now
   3.8.
+- :gh:`2576`: the C extension modules now use PEP 489 multi-phase
+  initialization instead of single-phase, which is the preferred mechanism for
+  extension modules. Runtime behavior is unchanged.
 - :gh:`2687`, [SunOS]: :func:`users` fails with ``ValueError``.
 - :gh:`2747`: the field order of the named tuple returned by :func:`cpu_times`
   has been normalized on all platforms, and the first 3 fields are now always
--- a/docs/credits.rst
+++ b/docs/credits.rst
@@ -118,6 +118,7 @@ Code contributors by year
 * :user:`Felix Yan <felixonmars>` - :gh:`2732`
 * :user:`Gabriel Changamire <gabrielchangamire-arch>` - :gh:`2809`
 * :user:`Hanson Wang <hansonw>` - :gh:`2810`
+* :user:`Jinhyuk Hong <jinh-labs>` - :gh:`2855`
 * :user:`Karl Hill <karlhillx>` - :gh:`2857`
 * :user:`Kataoka Katsuki <kataokatsuki>` - :gh:`2854`
 * :user:`Santhosh Raju <fraggerfox>` - :gh:`2805`
--- a/psutil/_psutil_aix.c
+++ b/psutil/_psutil_aix.c
@@ -907,97 +907,78 @@ static PyMethodDef PsutilMethods[] = {
 };
 
 
-struct module_state {
-    PyObject *error;
-};
-
-#define GETSTATE(m) ((struct module_state *)PyModule_GetState(m))
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 static int
-psutil_aix_traverse(PyObject *m, visitproc visit, void *arg) {
-    Py_VISIT(GETSTATE(m)->error);
+psutil_aix_exec(PyObject *mod) {
+    if (psutil_setup() != 0)
+        return -1;
+    if (psutil_posix_add_constants(mod) != 0)
+        return -1;
+    if (psutil_posix_add_methods(mod) != 0)
+        return -1;
+
+    if (PyModule_AddIntConstant(mod, "version", PSUTIL_VERSION))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "SIDL", SIDL))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "SZOMB", SZOMB))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "SACTIVE", SACTIVE))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "SSWAP", SSWAP))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "SSTOP", SSTOP))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "TCPS_CLOSED", TCPS_CLOSED))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "TCPS_CLOSING", TCPS_CLOSING))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "TCPS_CLOSE_WAIT", TCPS_CLOSE_WAIT))
+        return -1;
+    if (PyModule_AddIntConstant(mod, "TCPS_LISTEN", TCPS_LISTEN))
+        return -1;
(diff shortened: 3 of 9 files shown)


jinh-labs:
Gentle ping 👀

giampaolo:
Hi. Just as a quick note: this is greatly appreciated but it will take me some time to review (too busy lately - house renovation and other stuff :)).

giampaolo:
First of all, thanks. This is a high quality PR. But it aims at doing too much, and it's hard to review. I spotted some potential problems:

1. on Windows `InitializeCriticalSection` re-runs on every interpreter (it shouldn't).

2. `m_traverse`, `m_clear`, and `m_free` can crash on Python 3.7 and 3.8 (which we support). You currently do `GETSTATE(mod)->...` without checking if `GETSTATE(mod)` is `NULL`. That is fine on newer Python versions, but not on 3.7/3.8. With multi-phase module init the module state is allocated only during the exec phase. So the traverse/clear hooks should start with:

```c
if (GETSTATE(mod) == NULL)
    return 0;
```

...while `m_free` is covered, since it delegates to `m_clear`.

With that said, I would proceed in smaller steps here. I would stay focused on what #2577 was trying to do, and nothing more. That means converting the modules to multi-phase init, while keeping the old "one module instance per process" behavior. In practice: keep a `static int module_loaded` flag and raise `ImportError` if the exec slot runs more than once, just like #2577 did. It is fine to include the `Py_mod_gil` slot as a replacement for `PyUnstable_Module_SetGIL()`, with the same `#ifdef` guard you already added. The `m_name` fixes can go in this PR too. For modules that do not need per-module state yet, use:

```c
.m_size = 0
```

...and do not add GC hooks. This PR should not need new tests. It is mostly a mechanical conversion to multi-phase init while preserving the existing runtime semantics.

After this is done I can imagine 2 other PRs: 1) move the exception classes into per-module state (but accessed via the module object instead of the sys.modules-by-name lookup) and 2) implement sub-interpreter support.

jinh-labs:
Thanks for the feedback. I'm on vacation this week and will address this when I return!

jinh-labs:
Applie feedback: reduced this to the mechanical multi-phase conversion. Will work on the per-module state and sub-interpreter support PRs in the future.

giampaolo:
Thanks, this is close. A few things.

1. Please remove:

```c
    if (module_loaded) {
        PyErr_SetString(
            PyExc_ImportError, "cannot load module more than once per process"
        );
        return -1;
    }
```

Before this PR `import psutil` in a subinterpreter just worked, so this is a regression. Drop the flag in `_psutil_linux.c`, `_psutil_bsd.c`, `_psutil_osx.c`, `_psutil_sunos.c`, `_psutil_aix.c`. Nothing there needs to run only once. Windows is the exception: keep the flag, but wrap only `psutil_setup_windows()`, `psutil_set_se_debug()` and the 2 `PyErr_NewException()` calls with it (that's the `InitializeCriticalSection` issue). `TimeoutExpired` / `TimeoutAbandoned` then need a `Py_INCREF()` before `PyModule_AddObject()`, which steals a reference. The rest we'll handle when we add subinterpreter support.

2. Drop the leading underscores: `_psutil_aix_exec` -> `psutil_aix_exec`, `_psutil_linux_slots` -> `psutil_linux_slots`, etc.

3. You added the changelog in the typing section. Move it into "Enhancements".

Thanks

giampaolo:
Nice cleanup, the conversion itself looks right. I built the branch and tested it against 3.13.

.1 In `_psutil_windows.c`, the line `module_loaded = 1;` should be set at the end of the if block:

```c
if (!module_loaded) {
    if (psutil_setup_windows() != 0)
        return -1;
    ...
    module_loaded = 1;
}
```

Reason: `psutil_setup_windows()` and the two `PyErr_NewException()` calls can fail. If they do, a second import skips the block and falls through to `Py_INCREF(TimeoutExpired)` with `TimeoutExpired == NULL`.

.2 `ZombieProcessError` in `arch/posix/init.c` needs the same treatment:

```c
if (ZombieProcessError == NULL) {
    ZombieProcessError = PyErr_NewException(
        "_psutil_posix.ZombieProcessError", NULL, NULL
    );
    if (ZombieProcessError == NULL)
        return -1;
}
Py_INCREF(ZombieProcessError);
if (PyModule_AddObject(mod, "ZombieProcessError", ZombieProcessError)) {
    Py_DECREF(ZombieProcessError);
    return -1;
}
```

E.g. this fails with this PR on Python 3.13:

```python
import sys
import psutil

E1 = psutil._pslinux.cext.ZombieProcessError
del sys.modules['psutil._psutil_linux']
import psutil._psutil_linux
E2 = sys.modules['psutil._psutil_linux'].ZombieProcessError

assert E1 is E2 
```

giampaolo:
Perfect. Merging.

giampaolo:
Also see https://github.com/giampaolo/psutil/issues/2881.

PR #2854, opened 2026-06-30
macOS: fix SystemError in Process.cmdline() and environ()

kataokatsuki:
## Summary

- OS: macOS
- Bug fix: yes
- Type: core
- Fixes: N/A (related to #2708)

## Description

Related issue: #2708. There is no separate public ticket for the resulting
`SystemError` — it was observed locally while running Glances 4.5.5 with psutil
7.2.2 on macOS (Nix / Python 3.13).

The 7.2.2 fix for #2708 made `psutil_sysctl_procargs()` set `AccessDenied`
when `sysctl(KERN_PROCARGS2)` fails with `errno == 0`, but that branch returns
`0`. `0` means success for this helper, so `Process.cmdline()` and
`Process.environ()` continue and return a Python object while an exception is
already set, which CPython reports as:

```text
SystemError: <built-in function proc_cmdline> returned a result with an exception set
```

Return `-1` from that branch, matching the existing `EINVAL` and `EIO`
`AccessDenied` branches.

This also avoids parsing the malloc'd procargs buffer after the failed
`sysctl()` call.

Verified with:

- `make build PYTHON=.venv/bin/python`
- `make test PYTHON=.venv/bin/python ARGS='tests/test_process.py::TestProcess::test_cmdline tests/test_process.py::TestProcess::test_long_cmdline tests/test_process.py::TestProcess::test_environ tests/test_process.py::TestProcess::test_weird_environ tests/test_process.py::TestProcess::test_as_dict tests/test_unicode.py::TestFSAPIs::test_proc_cmdline tests/test_unicode.py::TestNonFSAPIS::test_proc_environ'`

Issue #2841, opened 2026-06-06
[macOS,M5 hardware] cpu_freq returned a result with an exception set


georgevreilly:
## Summary

- OS: macOS Tahoe 26.5.1
- Architecture: 64bit M5
- Psutil version: 7.2.2
- Python version: 3.14.5
- Type: core

## Description

This simple script

```python
#!/usr/bin/env python3

import psutil

print(psutil.cpu_freq(True))
```

fails

```
❯ uv run x.py
RuntimeError: invalid CPU frequency data

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/georgevreilly/src/pangur/x.py", line 5, in <module>
    print(psutil.cpu_freq(True))
          ~~~~~~~~~~~~~~~^^^^^^
  File "/Users/georgevreilly/src/pangur/.venv/lib/python3.14/site-packages/psutil/__init__.py", line 1945, in cpu_freq
    ret = _psplatform.cpu_freq()
  File "/Users/georgevreilly/src/pangur/.venv/lib/python3.14/site-packages/psutil/_psosx.py", line 166, in cpu_freq
    curr, min_, max_ = cext.cpu_freq()
                       ~~~~~~~~~~~~~^^
SystemError: <built-in function cpu_freq> returned a result with an exception set
```

When I run the repro on an M1 Mac on the same version of macOS, I get

```
[scpufreq(current=3228, min=600, max=3228)]
```

so this problem is likely specific to the new M5 hardware.

kataokatsuki:
Two separate things are going on here:

1. The `SystemError` is a psutil bug in the error path: when the
`voltage-states*-sram` properties fail validation, `psutil_cpu_freq()` sets
`RuntimeError: invalid CPU frequency data` (the direct cause in your
traceback) but still returns a result. I opened #2862 to fix that, so
you'd get the intended `RuntimeError` instead.

2. Why validation fails on M5 appears to be already identified in #2824:
Apple renumbered the `voltage-statesN-sram` keys on M5-family chips, so the
two keys hardcoded by the current code no longer match. That pending PR
rewrites the key discovery dynamically, which is what would make
`cpu_freq()` actually return frequencies on your machine.

giampaolo:
Fixed in 8.0.0. The voltage-states keys are discovered dynamically now instead of hardcoding two of them, and cpu_freq() returns None when the data isn't there, so no more SystemError.

BeArchiTek:
Hello,

[...]

PR #2862, opened 2026-07-10
macOS: fix SystemError in cpu_freq()

kataokatsuki:
## Summary

- OS: macOS
- Bug fix: yes
- Type: core
- Fixes: N/A (related to #2841)

## Description

Related issue: #2841. This only fixes the `SystemError` error path; actually
returning frequencies on the affected chips is covered by the pending
rewrite in #2824, so this PR does not close the issue.

In `psutil_cpu_freq()` (`psutil/arch/osx/cpu.c`), when either
`voltage-states*-sram` property is missing, has the wrong Core Foundation
type, or is too short (which is what happens on M5-family chips, see
#2824), the function sets `RuntimeError: invalid CPU frequency data` but
then falls through to the `Py_BuildValue()` at the end of `cleanup`,
returning a result with an exception set:

```text
SystemError: <built-in function cpu_freq> returned a result with an exception set
```

Same bug class as #2854. I checked the other `error:` / `cleanup:` labels
under `psutil/arch/osx/` and this is the only remaining spot.

Fix: initialize `py_ret` to `NULL`, assign it only on success, and return
it after the releases in `cleanup`.

No M5 hardware here, so I forced the error branch by temporarily changing
one property name in `cpu.c` to a non-existent key on an M1-family Mac:
unpatched, this reproduces the exact `SystemError` from the report;
patched, the same forced failure raises the intended `RuntimeError`. The
normal path returns the same values as unmodified master.

Verified with:

- `make build PYTHON=.venv/bin/python`
- `make test PYTHON=.venv/bin/python ARGS='tests/test_system.py -k cpu_freq'`

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -266,6 +266,9 @@ Others:
 - :gh:`2822`, [BSD]: :meth:`Process.cmdline` on NetBSD could raise
   ``OSError: [Errno 14] Bad address`` if the process about to exit. It now
   raises :exc:`NoSuchProcess` instead.
+- :gh:`2841`, [macOS]: :func:`cpu_freq` could raise :exc:`SystemError` when CPU
+  frequency data is missing or invalid in the IORegistry (e.g. on Apple M5
+  chips). It now raises :exc:`RuntimeError` instead.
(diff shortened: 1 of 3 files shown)


kataokatsuki:
Thanks for the review and merge!

PR #2863, opened 2026-07-11
Windows: speedup net_io_counters() by ~5x

### LINKED ISSUE #2695: [Windows] net_io_counters() performance
## Summary

- OS: Windows
- Type: performance

## Description

I was troubleshooting a CPU usage issue in Dask https://github.com/dask/distributed/issues/9161, and noticed that its usage of `psutil.net_io_counters()` was inducing nontrivial CPU load in the "Service Host: DNS Client" task.  (Each Dask worker is calling this at least once a second, which is turning out to add up to something non-trivial on a many-core machine)

I traced this to the use of `GetAdaptersAddresses` in `psutil_get_nic_addresses`. Experimenting with this interface, I determined there are a few opportunities to speed this up, with differing orders of complexity, some anticipated by https://github.com/giampaolo/psutil/issues/1426

I made a CLI utility to make the same sequence of syscalls as `psutil.net_io_counters`, to benchmark. The base version calls `GetAdapterAddresses` once to get a buffer size, then again to fetch data. Then it calls `GetIfEntry2` for each interface (of which there are maybe a dozen on my machine).

These are the optimization "phases"

Base version: 6.6 msec.
Skip flags: 2.8 msec.
Skip flags & remember size: 1.6 msec  
Cache all interfaces: 0.5 ms  (Possibly too aggressive)

First, there are flags that can be provided to `GetAdaptersAddresses` to skip the collection of various information. I tested with `GAA_FLAG_SKIP_DNS_INFO  | GAA_FLAG_SKIP_UNICAST | GAA_FLAG_SKIP_ANYCAST | GAA_FLAG_SKIP_MULTICAST  | GAA_FLAG_SKIP_DNS_SERVER`  which skips all the skippable options except the friendly name. This may render it incompatible with other uses of psutil_get_nic_addresses, so it might be necessary to take an input parameter to indicate whether the full information should be collected. But this is a big win.

If we remember the buffer size from the first call to `GetAdaptersAddresses` we can re-use it, and it will almost always be correct, differing only when a new interface appears. This means we only call `GetAdaptersAddresses` once per call, instead of twice, which helps a fair bit. (And if it does change, we'll still detect this and update the cached size, we won't miss any interfaces).

If we remember all the interfaces and don't call `GetAdaptersAddresses` at all, we get an even bigger speedup. This was suggested as a future optimization in https://github.com/giampaolo/psutil/issues/1426 but it means not detecting new interfaces when they come up, which changes the semantics since new interfaces won't be noticed.

Arman-Luthra:
## Summary

- OS: Windows
- Bug fix: no
- Type: performance
- Fixes: #2695

## Description

Implements the direction discussed in #2695: make `net_io_counters()` call
`GetAdaptersAddresses()` only once, and pass the `GAA_FLAG_SKIP_*` flags so it
stops collecting data we never read. No caching is introduced, so semantics
are unchanged: new interfaces are still detected on every call.

Changes to `psutil_get_nic_addresses()` (psutil/arch/windows/net.c):

- Instead of one call to query the buffer size + a second call to fetch the
  data, start with the 15KB buffer recommended by MSDN and retry on
  `ERROR_BUFFER_OVERFLOW` (which updates `bufferLength` to the required size).
  In practice this means a single syscall per invocation.
- The function now takes a `flags` argument:
  - `net_io_counters()` only uses `IfIndex` and `FriendlyName`, so it passes
    `GAA_FLAG_SKIP_UNICAST | GAA_FLAG_SKIP_ANYCAST | GAA_FLAG_SKIP_MULTICAST |
    GAA_FLAG_SKIP_DNS_SERVER | GAA_FLAG_SKIP_DNS_INFO` (everything skippable
    except friendly names).
  - `net_if_stats()` only uses `FriendlyName` and `Description`, so it passes
    the same skip flags.
  - `net_if_addrs()` needs the unicast addresses, so it keeps `flags=0`; it
    still benefits from the single-call buffer strategy.

### Benchmark

Windows 11, 8 NICs, Python 3.12 x64. Median over 1000 calls of the raw
`_psutil_windows` functions, 3 rounds (old and new extension builds loaded in
the same process):

| API                | before  | after   | speedup |
|--------------------|---------|---------|---------|
| `net_io_counters()` | 4.96 ms | 0.88 ms | ~5.5x   |
| `net_if_stats()`    | 6.79 ms | 2.67 ms | ~2.5x   |
| `net_if_addrs()`    | 5.20 ms | 2.41 ms | ~2.2x   |

Round-by-round medians for `net_io_counters()` were stable: 4.90 / 4.97 /
4.96 ms before vs 0.95 / 0.88 / 0.89 ms after. This is in line with the
numbers reported in #2695 (6.6 ms -> 1.7 ms on the reporter's machine).

### Behavior equivalence

To verify counters are unaffected by the skip flags, I loaded the old and new
builds of `_psutil_windows` side by side in one process and interleaved calls
(`old <= new <= old`, valid because the counters are monotonic):

- `net_io_counters()`: 50 rounds x 8 NICs x all 8 fields -> 0 violations,
  identical NIC name sets.
- `net_if_stats()` and `net_if_addrs()`: output compared for exact equality
  -> identical (8 NICs / 20 address entries).
- The `ERROR_BUFFER_OVERFLOW` retry path was exercised by temporarily building
  with a 64-byte initial buffer: all three functions still returned complete
  results.

### Tests

`pytest tests/test_windows.py tests/test_system.py -k net` run 3x locally:
10 passed, 2 skipped each time (one unrelated flake on the first cold run:
`test_net_if_addrs` timed out spawning its PowerShell subprocess; it passes
consistently on re-runs, including 3 consecutive runs of `TestNetAPIs`).

### Notes / limitations

- `GAA_FLAG_SKIP_DNS_INFO` is not defined in older SDK headers, so it's
  defined conditionally (`#ifndef`). It's documented for Windows 10 2004+; on
  older systems the extra bit is not expected to have any effect, but I could
  drop it from the mask if you prefer (the other 4 flags carry most of the
  win).
- I left `net_if_addrs()` at `flags=0` rather than skipping
  anycast/multicast/DNS there too, to keep this conservative; it can be
  tightened in a follow-up if desired.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -186,6 +186,11 @@ Others:
   initialization instead of single-phase, which is the preferred mechanism for
   extension modules. Runtime behavior is unchanged.
 - :gh:`2687`, [SunOS]: :func:`users` fails with ``ValueError``.
+- :gh:`2695`, [Windows]: :func:`net_io_counters` is ~5x faster.
+  ``GetAdaptersAddresses()`` is now invoked once instead of twice, and it skips
+  collecting unicast / anycast / multicast / DNS details, which were retrieved
+  but never used. :func:`net_if_stats` and :func:`net_if_addrs` also got
+  faster. (patch by :user:`Arman Luthra <Arman-Luthra>`)
 - :gh:`2747`: the field order of the named tuple returned by :func:`cpu_times`
   has been normalized on all platforms, and the first 3 fields are now always
   :field:`user`, :field:`system`, :field:`idle`. See compatibility notes below.
--- a/docs/credits.rst
+++ b/docs/credits.rst
@@ -111,6 +111,7 @@ Code contributors by year
 
 * :user:`Alex Chen <l46983284-cpu>` - :gh:`2859`
 * :user:`Amaan Qureshi <amaanq>` - :gh:`2770`
+* :user:`Arman Luthra <Arman-Luthra>` - :gh:`2695`
 * :user:`Bert Pluymers <bertpl>` - :gh:`2642`
 * :user:`Data-hYg <Data-hYg>` - :gh:`2789`
 * :user:`Ding Qiuran <dynapx>` - :gh:`2860`
--- a/psutil/arch/windows/net.c
+++ b/psutil/arch/windows/net.c
@@ -20,42 +20,35 @@
 #define GAA_FLAGS                                    \
     (GAA_FLAG_SKIP_ANYCAST | GAA_FLAG_SKIP_MULTICAST \
      | GAA_FLAG_SKIP_DNS_SERVER)
+// net_io_counters() and net_if_stats() don't read the unicast addresses
+// either, so they can also skip those (which are the most expensive to
+// collect).
+#define GAA_FLAGS_SKIP_UNICAST (GAA_FLAGS | GAA_FLAG_SKIP_UNICAST)
 
 
 static PIP_ADAPTER_ADDRESSES
-psutil_get_nic_addresses(void) {
-    ULONG bufferLength = 0;
+psutil_get_nic_addresses(ULONG flags) {
+    // A 15KB buffer is recommended by MSDN as it's usually big enough
+    // to make GetAdaptersAddresses() succeed on the first call, saving
+    // an extra (expensive) syscall to determine the required size.
+    ULONG bufferLength = 15 * 1024;
     ULONG ret;
     PIP_ADAPTER_ADDRESSES buffer;
     int attempt;
 
-    // A NIC showing up between the call determining the size and the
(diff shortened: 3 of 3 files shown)


giampaolo:
Excellent change. Please rebase.

Arman-Luthra:
Thanks! Rebased on top of 5010f4e7 and checks are green. What's left here after your commit: the single-call buffer strategy (15KB initial buffer instead of size query + fill), and skipping unicast collection for net_io_counters() and net_if_stats(), which never read it. net_if_addrs() keeps unicast. Benchmark numbers in the description are vs the pre-5010f4e7 baseline.

giampaolo:
Perfect, thanks. Very nice change.

Issue #2743, opened 2026-03-04
[OS] title: Proposal: Enabling Android/Termux Support in psutil


Manamama:
After installing (sic, can be done) and testing `psutil` on a modern Android 11 environment (Termux/aarch64), I've identified the minimal changes required to resolve the "platform not supported" errors and allow the library to function reliably (both with and without root).

### 1. Platform Recognition (`psutil/_common.py`)

In modern Termux environments, `sys.platform` returns `"android"`. The current logic strictly looks for `"linux"`, which causes `setup.py` and imports to fail immediately.
**Fix:**

```python
# psutil/_common.py
LINUX = sys.platform.startswith(("linux", "android"))
```

### 2. Graceful Degradation for Restricted Environments (`psutil/_pslinux.py`)

On Android, many `/proc` and `/sys` nodes are restricted to the shell user. Currently, `psutil` raises `PermissionError` (crashing the call). To make the library usable on non-rooted devices, these should be caught to allow "partial" system info retrieval.

**Key areas for `try...except PermissionError` wraps:**

- **`disk_partitions`**: Access to `/proc/filesystems` is often blocked.
- **`net_io_counters`**: Access to `/proc/net/dev` is blocked on many modern Android versions.
- **`sensors_battery`**: Access to `/sys/class/power_supply` is restricted.

**Example Fix:**

```python
def net_io_counters():
    try:
        with open_text(f"{get_procfs_path()}/net/dev") as f:
            lines = f.readlines()
    except PermissionError:
        return {}  # Return empty dict instead of crashing
```

### 3. Empirical Test Results (Termux / API 30)

- **Non-Rooted:** Core APIs like `cpu_count`, `cpu_percent`, `virtual_memory`, and `boot_time` work perfectly out of the box once the platform check is bypassed.
- **Rooted (`sudo`):** Running with elevated permissions resolves most `PermissionError` issues.
  - **Tests Passed:** 129
  - **Tests Failed:** 22 (mostly due to missing CLI tools like `who` or `vmstat` not being in the standard path/format).

### 4. Special Considerations for ARM/Android

- **CPU Frequency:** On Big.LITTLE architectures (common on Android), `cpu_freq()` may return entries per *cluster* rather than per *logical core* (e.g., 2 frequency entries for 8 cores). Test assertions should be relaxed to `len(freqs) <= cpu_count()`.
[...]

Issue #2191, opened 2023-01-09
[Termux] PermissionError: [Errno 13] Permission denied: '/proc/filesystems'


babaric-dev:
## Summary

* OS: Termux PRoot Ubuntu 22.04 (proot-distro)
* Architecture: arm64
* Psutil version:5.9.0
* Python version: 3.10.6
* Type: ??

## Description
I ran this in the Python interactive console of PRoot Ubuntu:
```
>>> import psutil
>>> psutil.disk_partitions()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/psutil/__init__.py", line 2021, in disk_partitions
    ret = _psplatform.disk_partitions(all)
  File "/usr/lib/python3/dist-packages/psutil/_pslinux.py", line 1275, in disk_partitions
    with open_text("%s/filesystems" % procfs_path) as f:
  File "/usr/lib/python3/dist-packages/psutil/_common.py", line 725, in open_text
    return open(fname, "rt", **kwargs)
PermissionError: [Errno 13] Permission denied: '/proc/filesystems'
>>>
```

It shows `PermissionError`.

True that we can't really access `/proc/filesystems` on Android. But at least handle the error with `try` `except` and return `[]`?
Better solution: use another way of getting partitions.

giampaolo:
Notes to self:
1) we read /proc/filesystems only if the function is invoked as `disk_partitions(all=False)`; if `all=True`, reading `/proc/filesystems` is superfluous, so it should not raise exception
2) find out how `/proc/filesystems` is written and provide a fallback function with hard-coded values in case of EPERM


@babaric-dev what happens if you call `disk_partitions(all=True)`? Does the function succeed? I want to make sure there are no other permission errors

babaric-dev:
With or without `all=True`, it shows the same error.

giampaolo:
I just pushed 2da99502. Can you try installing psutil from latest `master` branch and retry running `disk_partitions(all=True)`? Run:

```
python3 -m pip install --upgrade git+https://github.com/giampaolo/psutil.git@master
python3 -c "import psutil; print(psutil.disk_partitions(all=True))"
```

babaric-dev:
Works! But after updating to `git` version, I see it can't read `/proc/stat`. Another error. I will open an issue for this later

babaric-dev:
Why can't we close this?

giampaolo:
Because of:

[...]

PR #2891, opened 2026-07-24
[Android] Add support for Android and add support for `disk_partitions()` on Android

### LINKED ISSUE #2611: Android is not supported
## Summary

- OS: Android 10
- Architecture: aarch64
- Psutil version: 7.0.0
- Python version: 3.12.10
- Type: installation

Image

robertkirkman:
## Summary

- OS: Android
- Bug fix: yes
- Type: core
- Fixes: https://github.com/giampaolo/psutil/issues/2611, https://github.com/giampaolo/psutil/issues/2191, https://github.com/giampaolo/psutil/issues/2743

## Description

Fixes these two commands on Android

```bash
pip install psutil
python -c "import psutil; print(psutil.disk_partitions())"
```


robertkirkman:
This does not handle the `/proc/net/dev` or `/sys/class/power_supply` permission denials mentioned in #2743 . I have not personally needed or seen the use cases that require changes to those, so I think they are lower-priority, or should be handled by catching the exception in program code. If I should remove #2743 from the list of closed issues because of that, let me know.

PR #2896, opened 2026-07-25
fix: no false PID reuse when process identity is unknown

### LINKED ISSUE #2895: False `NoSuchProcess("PID has been reused")` when process identity can't be determined
Process identity is determined via `(pid, create_time)`, and is checked before terminate(), send_signal(), etc. Problem: sometimes create time is legitimately unknown. On Windows fast `create_time()` method, which raises `AccessDenied` for most elevated processes, and `Process.__init__` then stores `(pid, None)`. Same thing for zombies: `__init__` swallows `ZombieProcess` and keeps `(pid, None)`, and on NetBSD / OpenBSD a zombie that can be queried reports create time 0 (#2287, #2593).

When one side of the comparison has a real create time and the other has None / 0, the two idents differ, and psutil concludes the PID was reused by a new process. It wasn't: we simply couldn't fetch the identity the second time. Net effect, e.g. on Windows: `terminate()` on a process that meanwhile became inaccessible raises `NoSuchProcess("PID has been reused")` even though the process is still there.

I caught on NetBSD CI, where terminate() claimed pid 484 was reused while teardown proved the pid still existed 15 seconds later (it had become a zombie):

```
tests/test_process.py:1624: in test_wait_non_children
    child.terminate()
psutil/__init__.py:571: in _raise_if_pid_reused
    raise NoSuchProcess(self.pid, self._name, msg=msg)
E   psutil.NoSuchProcess: process no longer exists and its PID has been reused (pid=484)
...
E   psutil.TimeoutExpired: timeout after 15 seconds (pid=484)
E   AssertionError: 484   # pid_exists(484) still True
```

The fix: a null create time on either side is not proof of reuse; only flag reuse when both idents are known and differ.

Note this is really the completion of #2287: back then a zombie's null ctime silently broke `is_running()`, which returned `False` for a process that was still there. The PID-reuse detection added later in 3d21a43a turned that same condition into a raised `NoSuchProcess("PID has been reused")` on the `terminate()` / `send_signal()` / `kill()` path.

Sanjays2402:
## Summary

- OS: all (Windows / NetBSD / OpenBSD in practice)
- Bug fix: yes
- Type: core
- Fixes: #2895

## Description

`is_running()` compares the cached `(pid, create_time)` identity against a freshly constructed `Process`. When create time cannot be determined the identity is `(pid, None)` — the Windows fast `create_time()` raises `AccessDenied` for most elevated processes, `__init__` swallows `ZombieProcess` and keeps `None`, and on NetBSD/OpenBSD a queryable zombie reports create time 0.

With a real create time on one side and `None`/0 on the other the identities differ, so psutil concluded the PID was reused when it simply could not fetch the identity the second time — making `terminate()`/`send_signal()`/`kill()` raise `NoSuchProcess("...PID has been reused")` for a process that was still there. Reuse is now only flagged when both identities are known and differ.

New `test_unknown_ident_is_not_pid_reuse` fails without the fix (`assert p.is_running()` -> `terminated + PID reused`) and passes with it; `tests/test_process.py` is green (80 passed, 18 skipped).

### DIFF
--- a/psutil/__init__.py
+++ b/psutil/__init__.py
@@ -768,7 +768,14 @@ class Process:
             # have been reused by another process. Process identity /
             # uniqueness over time is guaranteed by (PID + creation
             # time) and that is verified in __eq__.
-            self._pid_reused = self != Process(self.pid)
+            other = Process(self.pid)
+            # A null creation time on either side means we simply could
+            # not determine the identity (AccessDenied on Windows, a
+            # zombie swallowed in __init__, or a zombie reporting ctime
+            # 0 on Net/OpenBSD). That is not proof of PID reuse, so only
+            # flag reuse when both identities are known and differ.
+            if self._ident[1] and other._ident[1]:
+                self._pid_reused = self != other
             if self._pid_reused:
                 _pids_reused.add(self.pid)
                 raise NoSuchProcess(self.pid)
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -1425,6 +1425,25 @@ class TestProcess(PsutilTestCase):
         with pytest.raises(psutil.NoSuchProcess, match=msg):
             p.children()
 
+    def test_unknown_ident_is_not_pid_reuse(self):
+        # A null create time on either side of the comparison means we
+        # simply could not determine the process identity (AccessDenied
+        # on Windows, or a zombie reporting no / 0 create time). That is
+        # not proof of PID reuse and must not raise NoSuchProcess.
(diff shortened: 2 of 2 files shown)


giampaolo:
?!?

I opened #2895 ticket 24 minutes ago, the PR https://github.com/giampaolo/psutil/pull/2897 1 minute ago and you opened this PR now.

giampaolo:
This is no good use of AI. Closing.

PR #2924, opened 2026-07-29
Fix net_if_stats() speed overflow on Windows for interfaces >4.29 Gbps

### LINKED ISSUE #2655: [Windows] network connect speed 5Gbps show 4294967295bps
## Summary

- OS: Windows
- Architecture: 64bit
- Psutil version:  5.9.8
- Python version: 3.12
- Type: core

## Description

{{{
  If the value exceeds the storage range of an int, a long should be used instead
}}}

omprakash2929:
## Summary

- OS: Windows
- Bug fix: yes
- Type: core
- Fixes: #2655

## Description

`net_if_stats()` was returning `4294967295` (32-bit overflow) as the
speed for network interfaces with link speed greater than 4.29 Gbps
(e.g. 5 Gbps).

### Root cause
The function used the legacy `GetIfTable()` API and `MIB_IFROW`
struct, where `dwSpeed` is a 32-bit `DWORD`. This overflows for any
interface speed above ~4.29 Gbps.

### Fix
Migrated `psutil_net_if_stats()` to use `GetIfEntry2()` with the
`MIB_IF_ROW2` struct instead - same approach already used by
`net_io_counters()` in this file. `TransmitLinkSpeed` in `MIB_IF_ROW2`
is a 64-bit `ULONG64`, so it correctly represents higher speeds.

### Testing
- Built locally on Windows with `pip install -e ".[test]"`
- Verified output manually via `psutil.net_if_stats()` - speeds now
  show correct values instead of overflowing
- Ran existing test suite: `test_net_if_stats` passes

Fixes #2655

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ docs/_build/
 dist/
 wheelhouse/
 .tests/
+test.psutilpy
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -328,6 +328,11 @@ Others:
 - :gh:`2628`, [Linux]: :func:`cpu_freq` no longer takes offline CPU cores into
   account. They were reported with all-zero frequencies, which dragged down the
   average ``current``, ``min`` and ``max`` values.
+- :gh:`2655`, [Windows]: :func:`net_if_stats` returned ``4294967295`` (32-bit
+  overflow) as the speed for network interfaces faster than ~4.29 Gbps (e.g. 5
+  Gbps NICs). Fixed by switching from the legacy ``GetIfTable()`` /
+  ``MIB_IFROW`` API to ``GetIfEntry2()`` / ``MIB_IF_ROW2``, which uses a 64-bit
+  ``TransmitLinkSpeed`` field.
 - :gh:`2711`, [Windows]: :func:`net_if_addrs` was returning ``None`` for the
   ``broadcast`` field of network interfaces instead of the correct broadcast
   address.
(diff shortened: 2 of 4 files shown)


giampaolo:
Thanks, the direction is right. A few things before this can go in:

1. Please rebase. 
2. The isup check lost the admin status. It should be:

```c
if (ifRow.OperStatus == IfOperStatusUp &&
    ifRow.AdminStatus == NET_IF_ADMIN_STATUS_UP)
```

3. Put back the three `Py_CLEAR()` calls before `return py_retdict;`. Those were added in #2691 to fix a leak. Without them `py_is_up` keeps a reference after the last iteration.
4. When `GetIfEntry2()` fails, add a `psutil_debug()` before skipping, so it's not silent.
5. Missing newline at end of file, lint will complain.

giampaolo:
Code changed. Please rebase.

omprakash2929:
@giampaolo I've addressed all the review comments:

1. Rebased on the latest master
2. Restored the admin status check in `isup`:
   `ifRow.OperStatus == IfOperStatusUp && ifRow.AdminStatus == NET_IF_ADMIN_STATUS_UP`
3. Restored the `Py_CLEAR()` calls at the start of the loop (py_nic_name, py_ifc_info, py_is_up)
4. Added `psutil_debug()` logging when `GetIfEntry2()` fails
5. Fixed the missing newline at the end of file

All checks are passing, and I've tested locally on Windows — `net_if_stats()` now returns correct speed values for interfaces above 4.29 Gbps, and the existing test suite passes.

PR #2960, opened 2026-08-06
fix linux hwmon high/crit of -273150 mC (0 K) (#2959)

### LINKED ISSUE #2959: [Linux] `sensors_temperatures()` treats placeholder thresholds as real alarm limits
## Summary

- OS: Linux
- Architecture: x86_64
- Psutil version: 7.2.2
- Python version: 3.14.6
- Type: core

## Description

On my own computer I noticed that applications like s-tui parse and use the alarm thresholds from psutil even if they definitely show negative placeholder values, resulting in alarms being triggered despite temperatures being okay.

`psutil.sensors_temperatures()` reads `high` and `critical` alarm thresholds from sysfs files like `/sys/class/hwmon/*/temp*_max` and `temp*_crit`. On many platforms these files hold a placeholder value when no threshold is configured. Psutil divides by 1000 and returns the sentinel as if it were a real alarm threshold.

In my specific case, the CROS EC / F75303 thermal drivers on my Framework 16 (and likely other ChromeOS devices) show `-273150` (absolute zero in milli-degrees Celsius) via these files. This doesn't just seem to happen on my computer but also seems to have tripped up an assertion in https://github.com/giampaolo/psutil/issues/2551 (though in that case the *current* value was unset, not an *alert threshold*).

icedream:
## Summary

- OS: Linux
- Bug fix: yes
- Type: core, tests
- Fixes: #2959

## Description

`sensors_temperatures()` returns -273.15 as the literal sensor value when a hwmon trip point has not been configured in the kernel. This is the sentinel value the core hwmon driver writes when a max/crit threshold is unset (-273150 mC = 0 K = absolute zero), not a real measured temperature.

The fix detects raw values of -273150 mC in both the hwmon and thermal zone paths and replaces them with None, matching the behavior for other unavailable thresholds. Apps like s-tui currently show -273.15 in red as an alarm limit, even though no alarm is present.

### DIFF
--- a/psutil/_pslinux.py
+++ b/psutil/_pslinux.py
@@ -1263,20 +1263,34 @@ def sensors_temperatures():
             # https://github.com/giampaolo/psutil/issues/1323
             continue
 
-        high = bcat(base + '_max', fallback=None)
-        critical = bcat(base + '_crit', fallback=None)
-        label = cat(base + '_label', fallback='').strip()
+        raw_high = bcat(base + '_max', fallback=None)
+        raw_crit = bcat(base + '_crit', fallback=None)
 
-        if high is not None:
+        def _is_zero_kelvin(val):
             try:
-                high = float(high) / 1000.0
-            except ValueError:
+                return int(val) == -273150
+            except (ValueError, TypeError):
+                return False
+
+        label = cat(base + '_label', fallback='').strip()
+        high = None
+        if raw_high is not None:
+            if _is_zero_kelvin(raw_high):
                 high = None
-        if critical is not None:
-            try:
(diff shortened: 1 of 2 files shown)


icedream:
I've backported my patch to 7.2.2 to test it out on system level, patching the `python-psutil` Arch Linux package on my system with it.

Screenshot showing it working within s-tui, top uses psutil 7.2.2 unpatched, bottom uses psutil 7.2.2 + my changes backported:

image

Backport patch file: 
[0001-fix-sensor-sentinel-thresholds.patch](https://github.com/user-attachments/files/30768795/0001-fix-sensor-sentinel-thresholds.patch)

PR #2970, opened 2026-08-09
Fix WinError 87 not handled in convert_oserror() (#2519)

### LINKED ISSUE #2519: [Windows] SystemError raised by process.send_signal(signal.CTRL_BREAK_EVENT)
## Summary

* OS: Microsoft Windows 11 Pro, 10.0.22631 Build 22631
* Architecture: 64bit
* Psutil version: 5.9.8
* Python version: 3.12.7
* Type: core

## Description

A call to `process.send_signal(signal.CTRL_BREAK_EVENT)` raises a `SystemError` exception whose original cause is a `WindowsError` exception saying `[WinError 87] The parameter is incorrect`.

In my case, I am sending CTRL_BREAK_EVENT to multiple processes. I suspect that the exception is raised because one of the processes manages to exit on its own just before the signal is sent, and the sending of the signal fails.

Traceback:

```
  File "<script_path_censored>.py", line 2019, in <module>
    sys.exit(inst.run())
             ^^^^^^^^^^
  File "<script_path_censored>.py", line 2014, in run
    return testbench.run("tb_ocd_testbench")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<script_path_censored>.py", line 223, in run
    self._run_impl(top)
  File "<script_path_censored>.py", line 689, in _run_impl
    rtlsim_proc.interrupt(sig, force_kill_after=self.HUNG_PROCESS_WAIT_TIME)
  File "<script_path_censored>.py", line 1332, in interrupt
    self._interrupt_whole_subtree_using_psutil(
  File "<script_path_censored>.py", line 1195, in _interrupt_whole_subtree_using_psutil
    p.send_signal(signal_type)
  File "C:\venv\Lib\site-packages\psutil\__init__.py", line 1283, in send_signal
    self._proc.send_signal(sig)
  File "C:\venv\Lib\site-packages\psutil\_pswindows.py", line 675, in wrapper
    raise convert_oserror(err, pid=self.pid, name=self._name) from err
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\venv\Lib\site-packages\psutil\_pswindows.py", line 664, in convert_oserror
    raise exc
  File "C:\venv\Lib\site-packages\psutil\_pswindows.py", line 673, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\venv\Lib\site-packages\psutil\_pswindows.py", line 858, in send_signal
    os.kill(self.pid, sig)
OSError: [WinError 87] The parameter is incorrect
```

It appears that a situation exists in which os.kill() call can end with WinError 87, however this case does not seem to be accounted for in [convert_oserror()](https://github.com/giampaolo/psutil/blob/de778ff674c4a4e2a5d5cf5cf12d52eb1d619ebb/psutil/_pswindows.py#L657).

omprakash2929:
## Summary
Fixes #2519

`Process.send_signal(CTRL_C_EVENT/CTRL_BREAK_EVENT)` on Windows can 
raise a raw `OSError` (WinError 87, "the parameter is incorrect") 
instead of `NoSuchProcess`, when the target process terminates on its 
own right before the signal is sent (race condition).

## Root cause
`convert_oserror()` in `_pswindows.py` only converts `ProcessLookupError` 
and permission errors into psutil-friendly exceptions. `WinError 87` 
raised by `os.kill()` in this race condition wasn't handled, so it 
leaked out as a raw `OSError`.

## Fix
Added a check for `winerror == 87` in `convert_oserror()`, converting 
it into `NoSuchProcess`, consistent with how other "process 
disappeared" cases are already handled.

## Testing
- `tests/test_windows.py::TestProcess::test_ctrl_signals` = PASSED
- `tests/test_windows.py::TestProcess::test_send_signal` =  PASSED
- Attempted to reproduce the exact race condition locally with a 
  stress test script; the underlying scenario is timing-dependent 
  (as noted by the original reporter), so the fix targets the error 
  code specifically, matching the documented `WinError 87` behavior.

### DIFF
--- a/psutil/_pswindows.py
+++ b/psutil/_pswindows.py
@@ -588,6 +588,13 @@ def convert_oserror(exc, pid=None, name=None):
         return AccessDenied(pid=pid, name=name)
     if isinstance(exc, ProcessLookupError):
         return NoSuchProcess(pid=pid, name=name)
+    if getattr(exc, "winerror", None) == 87:
+        # os.kill(pid, CTRL_C_EVENT / CTRL_BREAK_EVENT) can raise this
+        # (WinError 87, "the parameter is incorrect") when the process
+        # terminates on its own right before the signal is sent.
+        # See: https://github.com/giampaolo/psutil/issues/2519
+        msg = f"process {pid} disappeared before the signal could be sent"
+        return NoSuchProcess(pid=pid, name=name, msg=msg)
     raise exc
 
 


omprakash2929:
@giampaolo This PR addresses the WinError 87 issue reported in #2519. 

Since the race condition is timing-dependent and hard to reproduce 
deterministically in a local test, I focused on making the fix logically 
correct based on the documented error (`WinError 87` - invalid parameter 
raised by `os.kill()` when the process disappears right before the signal 
is sent), and verified it doesn't break the existing signal-related tests:

- `tests/test_windows.py::TestProcess::test_ctrl_signals` =  PASSED
- `tests/test_windows.py::TestProcess::test_send_signal` = PASSED

Happy to adjust the approach if you'd prefer a different way to handle this.

PR #2973, opened 2026-08-10
linux: follow up on cpulist parsing

fabiancrt:
## Summary

* OS: Linux
* Bug fix: yes
* Type: core, tests
* Fixes: N/A (follow-up to #2720)

## Description

This follows up on the suggestion in #2720 to pick up the cpulist parser separately.

`_get_eligible_cpus()` was only parsing a single `N-M` range, so a value like `0-3,8` would fall back to all CPUs instead of returning the actual eligible CPU list.

This adds `_parse_cpulist()` returning the expanded CPU IDs and uses it in `_get_eligible_cpus()`.

For example:

`0-3,8` -> `[0, 1, 2, 3, 8]`

Also added parser coverage and a regression test for the non-contiguous `Cpus_allowed_list` case.

Tested with:

* `tests/test_linux.py`: 131 passed, 4 skipped
* Ruff
* Black
* `git diff --check`


psutil/_pslinux.py
```diff
@@ -2117,12 +2140,15 @@ def cpu_affinity_get(self):
...
+            _re=re.compile(
+                br"^Cpus_allowed_list:[ \t]*([^\r\n]*)", re.MULTILINE
+            ),
         ):
             # See: https://github.com/giampaolo/psutil/issues/956
             data = self._read_status_file()
-            if match := _re.findall(data):
-                return list(range(int(match[0][0]), int(match[0][1]) + 1))
+            if match := _re.search(data):
+                return _parse_cpulist(decode(match.group(1)))
```
giampaolo: Do this:

```
    try:
        return _parse_cpulist(decode(match.group(1)))
    except ValueError as err:
        debug(f"can't parse Cpus_allowed_list ({err}); falling back")
```

tests/test_linux.py
```diff
@@ -708,6 +709,34 @@ def test_no_space_after_colon(self):
...
+
+    def test_invalid_integer(self):
+        for value in ("x", "0-x", "0-1-2"):
+            with self.subTest(value=value):
+                with pytest.raises(ValueError):
+                    _parse_cpulist(value)
+
+    def test_reversed_range(self):
+        with pytest.raises(ValueError, match="invalid CPU range"):
+            _parse_cpulist("3-1")
```
giampaolo: I think this (class) is too much for such a tiny functionality. I'd just keep the other test.

giampaolo (changes requested):

PR #2975, opened 2026-08-12
[Windows] disk_partitions: use wide-char APIs to fix UnicodeDecodeError (#1959)

### LINKED ISSUE #1959: [windows] UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc5
## Summary

* OS: { win10 }
* Architecture: { 64bit }
* Psutil version: { pip3 show psutil }
* Python version: { python3 -V }

  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python39\lib\site-packages\psutil\__init__.py", line 2013, in disk_partitions
    ret = _psplatform.disk_partitions(all)
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python39\lib\site-packages\psutil\_pswindows.py", line 274, in disk_partitions
    rawlist = cext.disk_partitions(all)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc5 in position 4: invalid continuation byte

i can't fix it, pls help me

alexchen-sys:
Fixes #1959.

On Windows, `psutil.disk_partitions()` crashed with `UnicodeDecodeError` on systems whose ANSI code page is not UTF-8 (e.g. cp1251): the ANSI variants of `GetLogicalDriveStrings`, `FindFirstVolumeMountPoint` and `GetVolumeInformation` return bytes in the system ANSI code page, and the `(ssss)` tuple format decodes them as strict UTF-8.

Following the maintainer's direction in the issue (use `GetVolumeInformationW`), this switches the whole function family to wide-char APIs: `GetLogicalDriveStringsW`, `GetDriveTypeW`, `GetVolumeInformationW`, `FindFirstVolumeMountPointW`/`FindNextVolumeMountPointW`, building strings via `PyUnicode_FromWideChar`. Widening `GetVolumeInformation` alone would not have been enough: the actual crash vector in the issue (byte 0xc5 in position 4 of a mount-point path under cp1251) comes from `FindFirstVolumeMountPoint`. `psutil_QueryDosDevice` is converted to `QueryDosDeviceW` as well — it had the same class of mismatch in reverse: the input path was UTF-8-decoded while `QueryDosDevice` returns ANSI bytes, so device targets with non-ASCII characters could never match.

Adds a regression test asserting the unicode contract of all four tuple fields (the original crash requires a live non-UTF8 ANSI code page and cannot be reproduced from Python on en-US CI).


tests/test_windows.py
```diff
@@ -372,6 +372,21 @@ def test_disk_partitions_mountpoint(self):
...
+        # and their bytes were decoded as UTF-8. The W (wide char)
+        # APIs fix the whole class of errors; here we assert the
+        # contract that all fields come out as unicode strings and
+        # the call does not crash.
+        for part in psutil.disk_partitions(all=True):
+            assert isinstance(part.device, str)
+            assert isinstance(part.mountpoint, str)
+            assert isinstance(part.fstype, str)
+            assert isinstance(part.opts, str)
+
```
giampaolo: We have exactly the same test in test_contracts.py. Just remove it.

giampaolo (changes requested):
Very nice. Just one small comment.

alexchen-sys:
Good catch, removed it — test_contracts.py already covers this. Thanks for the quick review.

Issue #2693, opened 2025-11-26
Linux on Apple ARM - psutil.net_if_stats() - OSError: [Errno 25] Inappropriate ioctl for device (originated from ioctl(SIOCETHTOOL))


damies13:
## Summary

- OS: Linux (debian docker container based on [python:3.13.9-slim](https://hub.docker.com/layers/library/python/3.13.9-slim/))
- Architecture: Apple ARM M4 (Docker running on Apple ARM M4 chip)
- Psutil version: 7.1.3
- Python version: python:3.13.9
- Type: core? (`psutil.net_if_stats()`)

## Description

My application uses psutil to get network card stats, I got a report from a user that when running in a docker container on a Mac that uses Apple ARM CPU they get the error `OSError: [Errno 25] Inappropriate ioctl for device (originated from ioctl(SIOCETHTOOL))` when my code calls `psutil.net_if_stats()`

User has confirmed running the same docker container on other hardware (e.g. windows machine with intel cpu) does not produce this error.

I've provided the stack trace that I was given below:

```
<hostname>  | Exception in thread Thread-452 (updatenetpct):
<hostname>  | Traceback (most recent call last):
<hostname>  |   File "/usr/local/lib/python3.13/threading.py", line 1043, in _bootstrap_inner
<hostname>  |     self.run()
<hostname>  |     ~~~~~~~~^^
<hostname>  |   File "/usr/local/lib/python3.13/threading.py", line 994, in run
<hostname>  |     self._target(*self._args, **self._kwargs)
<hostname>  |     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<hostname>  |   File "/usr/local/lib/python3.13/site-packages/rfswarm_agent/rfswarm_agent.py", line 667, in updatenetpct
<hostname>  |     nicstats = psutil.net_if_stats()
<hostname>  |   File "/usr/local/lib/python3.13/site-packages/psutil/__init__.py", line 2285, in net_if_stats
<hostname>  |     return _psplatform.net_if_stats()
<hostname>  |            ~~~~~~~~~~~~~~~~~~~~~~~~^^
<hostname>  |   File "/usr/local/lib/python3.13/site-packages/psutil/_pslinux.py", line 1044, in net_if_stats
<hostname>  |     duplex, speed = cext.net_if_duplex_speed(name)
<hostname>  |                     ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
[...]

PR #2989, opened 2026-08-28
fix: handle ENOTTY from ioctl(SIOCETHTOOL) in net_if_duplex_speed

TrueFurina:
Fixes #2985 (refs #2693)

**Problem**: on some setups ioctl(SIOCETHTOOL) returns ENOTTY while retrieving interface duplex/speed. Since ENOTTY is not handled, it propagates as OSError and makes net_if_stats() fail entirely.

**Fix**: treat ENOTTY like the existing EOPNOTSUPP / EINVAL / EBUSY cases (duplex = DUPLEX_UNKNOWN, speed = 0), so net_if_stats() keeps working on interfaces that do not support SIOCETHTOOL.

### DIFF
--- a/psutil/arch/linux/net.c
+++ b/psutil/arch/linux/net.c
@@ -97,7 +97,7 @@ psutil_net_if_duplex_speed(PyObject *self, PyObject *args) {
         }
     }
     else {
-        if ((errno == EOPNOTSUPP) || (errno == EINVAL) || (errno == EBUSY)) {
+        if ((errno == EOPNOTSUPP) || (errno == EINVAL) || (errno == EBUSY) || (errno == ENOTTY)) {
             // EOPNOTSUPP may occur in case of wi-fi cards.
             // For EINVAL see:
             // https://github.com/giampaolo/psutil/issues/797


fabiancrt:
This appears to duplicate #2985, which predates this PR and already contains the same ENOTTY change. Since this PR explicitly references #2985, could you clarify why a separate PR is needed? @giampaolo

psutil/arch/linux/net.c
```diff
@@ -97,7 +97,7 @@ psutil_net_if_duplex_speed(PyObject *self, PyObject *args) {
         }
     }
     else {
-        if ((errno == EOPNOTSUPP) || (errno == EINVAL) || (errno == EBUSY)) {
+        if ((errno == EOPNOTSUPP) || (errno == EINVAL) || (errno == EBUSY) || (errno == ENOTTY)) {
```
fabiancrt, line 100: This code is identical to the change already submitted in #2985. This PR was created afterward, from a fix-2985 branch, and explicitly references #2985. This appears to duplicate the existing PR.

fabiancrt (changes requested):

TrueFurina:
Thanks for the review. To clarify the relationship with #2985: #2985 is itself an open PR that already adds the same `ENOTTY` handling to `psutil/arch/linux/net.c` (treating `ENOTTY` like `EOPNOTSUPP` / `EINVAL` / `EBUSY`). So #2989 does overlap with #2985 rather than being independent — the only difference is brace/comment style.

Since #2985 already implements the fix, I'm happy to close #2989 in its favor unless you'd prefer #2989's formatting. Please let me know which PR you'd like to keep and I'll close the other.

